PR #1, opened 2024-11-19
Port over Slack server

jspahrsummers:

PR #2, opened 2024-11-19
Create package for each server

jspahrsummers:
These are now available on npm as individual packages:
image
Or you can `npm install @modelcontextprotocol/servers` to get all of them at once.

# Depends on

* #1

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,11 +37,11 @@ jobs:
         with:
           node-version: 18
           cache: npm
-          registry-url: 'https://registry.npmjs.org'
+          registry-url: "https://registry.npmjs.org"
 
       - run: npm ci
 
       # TODO: Add --provenance once the repo is public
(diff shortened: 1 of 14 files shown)


dsp-ant (approved):

zeweihan:
This is a real pain point for enterprise document workflows. In legal and financial sectors, file paths often contain special characters from various languages and legal terminology. 

We ran into similar path handling issues when building the document pipeline for [AI Workdeck](https://github.com/zeweihan/aiworkdeck) — our open-source AI workspace for legal teams. Chinese legal document filenames frequently contain characters like 「」、（）、《》 that trip up naive path parsing.

For the Windows path handling specifically, using `pathlib` with `PureWindowsPath` for normalization before passing to the filesystem operations should handle most edge cases. The UNC path prefix (`\\?\`) is also important for paths exceeding 260 characters, which is surprisingly common in legal document management systems with deep folder hierarchies.

PR #28, opened 2024-11-21
Fix SQLite server logging to file

jspahrsummers:
When launched from a GUI app, the working directory won't be writable, so this causes a fatal error.

Switch to just using stderr logging.


navayla777-dotcom:
Merge request merge

PR #176, opened 2024-12-02
servers: make tool call result spec compatible

### LINKED ISSUE #151: Bug in example servers (toolResult confusion)
**Describe the bug**

it seems like a couple of servers are returning `content` within `toolResult`.

Visible in this search (puppeteer/google maps): https://github.com/search?q=repo%3Amodelcontextprotocol%2Fservers%20toolResult&type=code

From the MCP client typescript completions, I see that `content` should be NEXT TO `toolResult`, not inside it.

**Expected behavior**
It makes sense to me that content is the "chat response" and toolResult is arbitrary data that the tool returns, so content should never be inside the result. **Please tell me** if this is a wrong assumption (since we have no examples of full featured clients).

**Additional context**
I am playing with the client and trying to understand inputs/outputs, vscode with zod make it _really_ hard to see what's happening.

Since I played with created my own MCP, I can see why the bug can easily happen, this is what I get as autocomplete:

```ts
server.setRequestHandler(CallToolRequestSchema, async (request) => {

return {
// autocomplete list I get in vscode (all of these are `unknown`):
capabilities
completion
content
contents
messages
prompts
protocolVersion
resources
resourceTemplates
serverInfo
then
tools
_meta
description
isError
}
```

1. It seems to be like a bunch of these don't belong here
2. No `toolResult` as an option at all
3. Nothing has a type

From the client side, I just see these from callTool `{ content, isError, toolResult, _meta }`

Thanks! MCP is cool!

dsp-ant:
Fixes #151


jspahrsummers (dismissed):
Thanks! We should also remove this from the SDK—added to my to-do list, but anyone can get to it first 😅

jspahrsummers (approved):

PR #181, opened 2024-12-02
Browserbase Integration

alexdphan:
A simple browserbase implementation built based upon puppeteer


jspahrsummers:
Thank you for taking the time to build this! We've now made [some changes](https://github.com/orgs/modelcontextprotocol/discussions/61) to how we'll manage the servers repo going forward.

**TLDR:** We're restructuring the MCP [servers](https://github.com/modelcontextprotocol/servers) repository to focus on reference implementations only. New server implementations should be developed in separate repositories and can be added as links to our README, rather than submitted as PRs to this repository.

In line with that new approach, I'll close out this PR, but we would love to include a link to this server once it's in its own repo. Please do submit a new PR when it's ready!

PR #221, opened 2024-12-04
docs: add llamacloud server

marcusschiesser:
Update README to include mcp llamacloud server from https://github.com/run-llama/mcp-server-llamacloud


jspahrsummers (approved):
Great, thank you!

Issue #267, opened 2024-12-08
Could not attach to MCP server filesystem


mearleycf:
**Describe the bug**
I'm trying to implement various MCP servers, including the filesystem server, the memory server, and the playwright server. However, I keep getting failures. 
![CleanShot 2024-12-08 at 03 24 23](https://github.com/user-attachments/assets/870d81d9-fa88-4071-a2fb-9bfb09cb7230)
I've uploaded a screenshot of the errors I get in the app. In the dev tools console, I see:

```
Could not attach to MCP server filesystem: eE: MCP error -2: Request timed out
  at https://claude.ai/_next/static/chunks/foo.js:1:4404
  at r (https://claude.ai/_next/static/chunks/bar.js:15:2950)
(anonymous) @baz.js:18
```
This is my config.json content:
```json
{
  "mcpServers": {
    "filesystem": {
      "command": "npx",
      "args": [
        "@modelcontextprotocol/server-filesystem",
        "/Users/myusername/code/"
      ]
    },
    "mcp-obsidian": {
      "command": "uvx",
      "args": ["mcp-obsidian"],
      "env": {
        "OBSIDIAN_API_KEY": "reallylongkey"
      }
    },
    "@modelcontextprotocol-server-memory": {
      "runtime": "node",
      "command": "npx",
      "args": ["-y", "@modelcontextprotocol/server-memory"]
    },
    "playwright": {
      "command": "npx",
      "args": ["-y", "@automatalabs/mcp-server-playwright"]
    }
  }
}

Note: MCP Obsidian DOES work--the other three do not. 


**To Reproduce**
Steps to reproduce the behavior:
1. Set up config.json
2. launch Claude desktop app
3. Wait for error message to appear

**Expected behavior**
I expect to see something like this:
```
MCP Obsidian Server running on stdio
Allowed directories: [ '/users/myusername/code/pathtovault/vaultname' ]
```

**Logs**
This is what appears in the file `/Users/myusername/Library/Logs/Claude/mcp.log`:
```
2024-12-08T08:23:21.820Z [info] Attempting to connect to MCP server filesystem...
2024-12-08T08:23:21.823Z [info] Attempting to connect to MCP server mcp-obsidian...
2024-12-08T08:23:21.823Z [info] Attempting to connect to MCP server @modelcontextprotocol-server-memory...
2024-12-08T08:23:21.824Z [info] Attempting to connect to MCP server playwright...
[...]

PR #299, opened 2024-12-10
Update case sensitivity handling for filesystem server

### LINKED ISSUE #194: server-filesystem case insensitivity on MacOS
**Describe the bug**
Claude can't access files in local folder on MacOS via server-filesystem as it appears to ignore case.

**To Reproduce**
1) Add to `claude_desktop_config.json`:
```
{
    "mcpServers": {
        "filesystem": {
            "command": "npx",
            "args": [
                "-y",
                "@modelcontextprotocol/server-filesystem",
                "/Users/ido/Dropbox/Work"
            ]
        }
    }
}
```
2) Open new chat in Claude Desktop app and prompt:
```
list_allowed_directories
```
3) Receive a response
```
Let me check the allowed directories again to be completely clear about what access I have.

{}
Allowed directories:
/users/ido/dropbox/work

Based on this, the only allowed directory is:
/users/ido/dropbox/work

This seems to be the exact path (case sensitive) that I should be working with. Let me try to list its contents to make sure we can access it:

{
  `path`: `/users/ido/dropbox/work`
}
Error: Parent directory does not exist: /users/ido/dropbox

It seems there might be an issue with how the directories are set up or mounted on the server side. Even though this directory is listed as allowed, I'm unable to access it. Would you be able to verify if there might be any configuration issues with the file system access?
```

**Expected behavior**
Should read configuration correctly and use case sensitive folder and file names

**Additional context**
- MacOS Sequoia 15.1 (on Apple M2 Pro)
- Claude Desktop version 0.7.5
- other commands like `read_file` return similar issues like file not found
- other MCPs like `fetch` are working fine

maheshmurag:
Should fix #194 

This issue mentioned problems with how the filesystem server was handling case sensitivity - it shouldn't always do `.toLowerCase()` since this breaks access to paths on case-sensitive filesystems. This PR removes the forced lowercase conversion and lets the native filesystem handle case sensitivity.

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -24,7 +24,7 @@ if (args.length === 0) {
 
 // Normalize all paths consistently
 function normalizePath(p: string): string {
-  return path.normalize(p).toLowerCase();
+  return path.normalize(p);
 }
 
 function expandHome(filepath: string): string {


LarsBuur:
This change helps, and I never see the error about unknown directory in the "mcp-server-filesystem.log" mentioned here https://github.com/modelcontextprotocol/servers/issues/267#issuecomment-2526975128

Now, I constantly get:
```
Secure MCP Filesystem Server running on stdio
Allowed directories: [ '/Users/someone/Documents' ]
```

What is strange is that it is still very random when the filesystem mcp is working. Sometimes (and very random), I get a notification in Claude Desktop saying: "(i) Could not attach to MCP server filesystem"

Claude desktop Version 0.7.5 (0.7.5)

jspahrsummers:
Thanks @maheshmurag!

@LarsBuur Do you mind commenting on the issue instead? It seems unrelated to this fix specifically.

jspahrsummers (approved):

Issue #297, opened 2024-12-10
Improvment: Use resources and resource urls with the filesystem MCP server


snaggle-ai:
**Is your feature request related to a problem? Please describe.**
The filesystem MCP server method for reading non-text files is to call the read_file tool and return a base64 encoded string of the contents of the file. This is of limited usefulness and on files of moderate size it maxes out the response size and crashes claude.

This pretty much prevents using the filesystem MCP to interact with image files on local storage as any modern image is too large to load as a base64 string (and pass all the text through the LLM consuming tons of tokens).

The modelcontextprotocol spec describes how binary resources can be used for images and documents: https://modelcontextprotocol.io/docs/concepts/resources#binary-resources

**Describe the solution you'd like**
As a reference implementation the filesystem MCP should return resource urls rather than raw contents for non-text files, or for files over a certain size.

**Describe alternatives you've considered**
Allow Claude to directly load images into its context based on a path to a file without needing to use MCP

xhiroga:
I too am interested in this issue.
Having said that, I think its uniqueness would be an issue if it returns a URI, since I thought MCP was probably designed to run both on a local machine and on a web app.

Is there a solution to this problem, such as `http://resource-name.mcp`, where the mcp server is supposed to act as an HTTP server and pay out a local-machine-only domain?
(Sorry if I am wrong in my assumptions, etc. I would like to hear other people's opinions...)

OR13:
For small files, ideally we would have `read_file_as_data_uri`.

For larger files, the file needs to be a resource that is available to the mcp server that needs to process it, which could be implemented many different ways... but ideally, a local MCP server would upload the file and create a short URI and then a remote MCP server could access that URI assuming it had correct permissions.

I suppose you could also serve local files via a dev tunnel to a remote MCP server... security issues with all of these options.

PR #307, opened 2024-12-11
filesystem - load files into model context

snaggle-ai:
## Description

Adds resource templates and prompts to the filesystem reference server to load files and folders into model context.
Applied a limit of 30 files for loading into context.

- loads files (e.g. images, pdfs) into context (rather than as base64 in a message).
- demonstrates a practical use of prompts for loading files into context.
image
image
image

Note: This differs from drag/drop of files into Claude is that this allows the chat to keep track of the uris of files in the context and pass them to tools

## Server Details

- Server: filesystem
- Changes to: resources, prompts

## Motivation and Context
- Allows loading images, pdf etc. into model context. There are lots of reasons why files would be loaded into context.
- Tools do not load into model context but into a message.
- Model context is preferable for images, pdfs and most non-text data
- Model context is the only way to load larger files - these cannot be loaded via tools because they hit message response size limits.
- demonstrates a practical use of prompts for loading files into context.

## How Has This Been Tested?
Tested prompts with Claude

## Breaking Changes
None

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
Addresses this issue: https://github.com/modelcontextprotocol/servers/issues/297
Relevant to this discussion: https://github.com/modelcontextprotocol/specification/discussions/90
Very valuable conversation on reddit: https://www.reddit.com/r/mcp/comments/1hamhob/comment/m1cbq30/

### DIFF
--- a/src/filesystem/README.md
+++ b/src/filesystem/README.md
@@ -9,6 +9,7 @@ Node.js server implementing Model Context Protocol (MCP) for filesystem operatio
 - Move files/directories
 - Search files
 - Get file metadata
+- Load files into model context
 
 **Note**: The server will only allow operations within directories specified via `args`.
 
@@ -16,7 +17,18 @@ Node.js server implementing Model Context Protocol (MCP) for filesystem operatio
 
 ### Resources
 
-- `file://system`: File system operations interface
+The server supports files and directories as resources:
+
+- **Files** (`file:///<path>`)
+  - Text files are provided as UTF-8 text
+  - Binary files (images, PDFs, etc.) are provided as base64-encoded blobs
+  - Automatic MIME type detection for common file types
+
+- **Directories** (`folder:///<path>`)
+  - Listed as resources with `inode/directory` MIME type
+  - Contents provided as newline-separated file listing
+  - Pagination support for large directories
+
 
 ### Tools
 
@@ -102,6 +114,21 @@ Node.js server implementing Model Context Protocol (MCP) for filesystem operatio
   - No input required
   - Returns:
     - Directories that this server can read/write from
+
+### Prompts
+
+- **load_file**
+  - Load and a single file into model context
+  - Input: `path` (string, absolute path to file)
+  - Automatic file type detection and handling, supports text, image, and binary files
+  - Only supports files in allowed directories
+
+- **load_folder**
+  - Recursively load all files in a folder into model context
+  - Input: `path` (string, absolute path to folder)
+  - Automatic file type detection and handling, supports text, image, and binary files
+  - Limit of 30 files
+  - Only supports files in allowed directories
 
 ## Usage with Claude Desktop
 Add this to your `claude_desktop_config.json`:
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -4,7 +4,11 @@ import { Server } from "@modelcontextprotocol/sdk/server/index.js";
 import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
 import {
   CallToolRequestSchema,
+  GetPromptRequestSchema,
+  ListPromptsRequestSchema,
+  ListResourceTemplatesRequestSchema,
   ListToolsRequestSchema,
+  ReadResourceRequestSchema,
   ToolSchema,
 } from "@modelcontextprotocol/sdk/types.js";
 import fs from "fs/promises";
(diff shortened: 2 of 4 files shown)


src/filesystem/README.md
```diff
@@ -9,14 +9,26 @@ Node.js server implementing Model Context Protocol (MCP) for filesystem operatio
...
 - Get file metadata
+- Load files into model context
 
 **Note**: The server will only allow operations within directories specified via `args`.
 
 ## API
 
 ### Resources
 
-- `file://system`: File system operations interface
```
snaggle-ai, line 19: this was/is not true, there are no resources exposed by the filesystem mcp before this PR

src/filesystem/index.ts
```diff
@@ -577,6 +716,349 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
+
+server.setRequestHandler(ReadResourceRequestSchema, async (request) => {
+  try {
+    // Handle both file:// and folder:// URIs
+    const match = request.params.uri.match(/^(file|folder):\/\/(.+)/);
+    if (!match) {
+      throw new Error('Invalid URI format');
+    }
+
+    const [, scheme, pathStr] = match;
```
jspahrsummers, line 745: Consider using the `URL` class for parsing here–makes it easier to check the scheme and extract different parts of the URL.

src/filesystem/index.ts
```diff
@@ -577,6 +716,349 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
+    // Convert URI slashes to platform-specific path separators
+    const platformPath = pathStr.split('/').join(path.sep);
```
jspahrsummers, line 747: I _think_ this shouldn't be needed—forward slashes work fine on Windows.

src/filesystem/index.ts
```diff
@@ -577,6 +716,349 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
+    return FileType.BINARY;
+  }
+
+  // If file is smaller than 100kb and type is unknown, treat as text
+  if (stats.size < 100 * 1024) {
+    return FileType.TEXT;
+  }
+
+  // Otherwise fall back to binary
+  return FileType.BINARY;
```
jspahrsummers, line 1059: TIOLI: Best strategy here is probably to try reading it as a string, and if that fails, fall back to binary. This is something we could do later, though.

src/filesystem/index.ts
```diff
@@ -565,6 +573,137 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
         };
       }
 
+      case "load_folder": {
```
jspahrsummers, line 576: Did you mean to leave this tool in? I don't see it in the README.

jspahrsummers (changes requested):
Thanks, this is awesome! Just a few questions/suggestions, then happy to merge this in.

src/filesystem/index.ts
snaggle-ai: I _think_ it will always be possible to "read" any file "as a string", just the string might be absolute garbage and weird characters.

I was thinking about some options around this
 - have separate read_binary, read_image, read_text functions - the caller will generally know what they want and could keep this approach as a `read_auto` or someting

- alternatively make an optional input argument for binary/image/text and fall back to this behaviour if it's not specified

src/filesystem/index.ts
snaggle-ai: Good catch thank you, I will remove it - I was using this personally but I think it's still a bit too janky to commit.

Many issues/questions:
Should it recurse?
Loading more than 5 files into context this way isn't supported by Claude so it frequently breaks

src/filesystem/index.ts
snaggle-ai: great point!

src/filesystem/index.ts
jerome3o-anthropic: I can definitely see how this could blow out the context (depending on the folder) off the cuff I imagine a good ux would be:
* don't recurse, but indicate that there are nested folder which an AI can further load
* have some safeguards that prevent loading in too much data (maybe a text block saying there are "xyz more files" or something?)

jerome3o-anthropic (commented):
hey - I'm just getting spun up on this, these all look like great changes!

I haven't dug into the code for a proper review, but because I'm time poor atm, I'm keen to know if @snaggle-ai you're still keen to iterate on this/see it merged? I think they all look like awesome additions.

l1n:
@snaggle-ai would be really excited to see this continue!

TuanaCelik:
Same! Would love to see this PR merge ❤️

domdomegg:
Heya, this PR looks stale. I'm closing it to help us as maintainers more easily track active PRs, but feel free to reopen it if you're still working on it. I also see there's interest from others on this - I think we'd be open to a similar PR that is up to date and working to load in files as resources, if any of you have capacity :)

PR #368, opened 2024-12-17
Add Docker support for sample mcp servers

slimslenderslacks:
## Server Details

- Server: for all 18 servers, we are adding support for running them using docker
- Changes to: no changes to server functionality or how they are run using either npx or uv.  Minor code changes to filesystem, and puppeteer to support containerization

## Motivation and Context
This is an additional option for running mcp servers that addresses a few pain points:

* Environment Conflicts: Installing MCP servers often requires specific versions of Node.js, Python, and other dependencies, which may conflict with existing installations on a user’s machine.
* Complex Setup: Users need to manually clone the server repository, install dependencies, and configure their environment, which can be time-consuming and error-prone.
* Cross-Platform Challenges: Running the servers consistently across different architectures (e.g., x86 vs. ARM) or operating systems can introduce additional complexity.
* Dependency Isolation: Ensuring that server-specific runtime dependencies like Git or SQLite do not interfere with the host environment can be tricky.

## How Has This Been Tested?

We are testing with a claude_desktop_config.json containing all 18 servers. We are also adding additional entries to each of the server README.md projects to describe how to make the optional docker-based client configuration. The list of tests we have to describe here is quite long since we have to re-test all 18 servers.

## Breaking Changes
No, this will not impact any existing installs.  It is an orthogonal, and new, way to install and run the servers.

## Types of changes
- [x] Documentation update

## Checklist

### DIFF
--- a/src/everart/README.md
+++ b/src/everart/README.md
@@ -18,11 +18,10 @@ Add to Claude Desktop config:
     "everart": {
       "command": "docker",
       "args": ["run", "-i", "--rm", "-e", "EVERART_API_KEY", "mcp/everart"],
-    "env": {
-      "EVERART_API_KEY": "your_key_here"
+      "env": {
+        "EVERART_API_KEY": "your_key_here"
+      }
     }
-    },
-    
   }
 }
 ```
--- a/src/everart/README.md
+++ b/src/everart/README.md
@@ -17,7 +17,7 @@ Add to Claude Desktop config:
   "mcpServers": {
     "everart": {
       "command": "docker",
-      "args": ["run", "-i", "--rm", "-e", "EVERART_API_KEY=$EVERART_API_KEY", "ai/mcp-everart"],
+      "args": ["run", "-i", "--rm", "-e", "EVERART_API_KEY", "ai/mcp-everart"],
     "env": {
       "EVERART_API_KEY": "your_key_here"
     }
--- a/src/everart/README.md
+++ b/src/everart/README.md
@@ -17,7 +17,7 @@ Add to Claude Desktop config:
   "mcpServers": {
     "everart": {
       "command": "docker",
-      "args": ["run", "-i", "--rm", "-e", "EVERART_API_KEY", "ai/mcp-everart"],
+      "args": ["run", "-i", "--rm", "-e", "EVERART_API_KEY", "mcp/everart"],
     "env": {
       "EVERART_API_KEY": "your_key_here"
     }
@@ -94,5 +94,5 @@ You can also click the URL above to view the image again.
 ## Building w/ Docker
 
 ```sh
-docker build -t ai/mcp-everart -f src/everart/Dockerfile . 
+docker build -t mcp/everart -f src/everart/Dockerfile . 
 ```
--- a/src/everart/README.md
+++ b/src/everart/README.md
@@ -10,6 +10,25 @@ export EVERART_API_KEY=your_key_here
 
 ## Config
 Add to Claude Desktop config:
+
+### Docker
+```json
+{
+  "mcpServers": {
+    "everart": {
+      "command": "docker",
+      "args": ["run", "-i", "--rm", "-e", "EVERART_API_KEY=$EVERART_API_KEY", "ai/mcp-everart"],
+    "env": {
+      "EVERART_API_KEY": "your_key_here"
+    }
+    },
+    
+  }
+}
+```
+
+### NPX
+
 ```json
 {
   "mcpServers": {
@@ -70,4 +89,10 @@ Generation details:
 - Image URL: https://storage.googleapis.com/...
 
 You can also click the URL above to view the image again.
+```
+
+## Building w/ Docker
+
+```sh
+docker build -t ai/mcp-everart -f src/everart/Dockerfile . 
 ```
--- a/src/fetch/Dockerfile
+++ b/src/fetch/Dockerfile
@@ -1,0 +1,29 @@
+# Use a Python image with uv pre-installed
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
+
+# Install the project into `/app`
+WORKDIR /app
+
+# Enable bytecode compilation
+ENV UV_COMPILE_BYTECODE=1
+
+# Copy from the cache instead of linking since it's a mounted volume
+ENV UV_LINK_MODE=copy
+
+# Install the project's dependencies using the lockfile and settings
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --no-install-project --no-dev
+
+# Then, add the rest of the project source code and install it
+# Installing separately from its dependencies allows optimal layer caching
+ADD . /app
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --frozen --no-dev
+
+# Place executables in the environment at the front of the path
+ENV PATH="/app/.venv/bin:$PATH"
+
+# when running the container, add --db-path and a bind mount to the host's db file
+ENTRYPOINT ["uvx" , "mcp-server-fetch"]
--- a/src/fetch/Dockerfile
+++ b/src/fetch/Dockerfile
@@ -1,5 +1,5 @@
 # Use a Python image with uv pre-installed
-FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim AS uv
 
 # Install the project into `/app`
 WORKDIR /app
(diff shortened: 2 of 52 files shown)


src/fetch/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ENV UV_COMPILE_BYTECODE=1
+
+# Copy from the cache instead of linking since it's a mounted volume
+ENV UV_LINK_MODE=copy
+
+# Install the project's dependencies using the lockfile and settings
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --no-install-project --no-dev
```
dsp-ant, line 17: Any particular reason we use --no-install-project here? The alternative would be to do that and run just call `mcp-server-fetch` since our PATH already points to the .venv. This would avoid having to go through uvx. I am fine with this, but i am curious if there is a reason for one over the other.

src/gdrive/README.md
```diff
@@ -49,6 +49,21 @@ To authenticate and save credentials:
...
 To integrate this server with the desktop app, add the following to your app's server configuration:
 
+#### Docker
+
+```json
+{
+  "mcpServers": {
+    "gdrive": {
+      "command": "docker",
+      "args": ["run", "-i", "--rm", "--mount", "type=bind,source=/Users/colinmcneil/Desktop/gcp-oauth.keys.json,target=/gcp-oauth.keys.json", "-v", "mcp-gdrive:/gdrive-server", "-e", "GDRIVE_OAUTH_PATH=/gcp-oauth.keys.json", "-e", "GDRIVE_CREDENTIALS_PATH=/gdrive-server/credentials.json", "mcp/gdrive"]
```
dsp-ant, line 59: Is the username here a leftover?

src/gdrive/README.md
```diff
@@ -49,6 +49,21 @@ To authenticate and save credentials:
 
 To integrate this server with the desktop app, add the following to your app's server configuration:
 
+#### Docker
+
```
dsp-ant, line 53: We likely want ot highlight the difference between the npx and the docker version (mostly just passing credentials (which i think in the explizit form is nicer).

src/git/README.md
```diff
@@ -120,6 +120,19 @@ Add this to your `claude_desktop_config.json`:
...
 </details>
 
+<details>
+<summary>Using docker</summary>
+
+```json
+"mcpServers": {
+  "git": {
+    "command": "docker",
+    "args": ["run", "--rm", "-i", "type=bind,src=/Users/slim,dst=/projects", "mcp/git"]
```
dsp-ant: Is the username here deliberate? I think it's easy to miss that someoen needs to adjust the username. We should make a little note.

src/puppeteer/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM node:22-bullseye-slim
```
dsp-ant, line 1: any particular reason we use bullseye here and bookworm for uv? Also just in general, would it be more ergonomic to use alpine? Not sure if that would work. I assume moving to alpine would cause more issues due to musl and static linking.

src/puppeteer/README.md
```diff
@@ -65,6 +65,21 @@ The server provides access to two types of resources:
 ## Configuration to use Puppeteer Server
 Here's the Claude Desktop configuration to use the Puppeter server:
 
+### Docker
+
+```json
```
dsp-ant, line 70: Lets add a note about the difference between implementations: Something like 
```suggestion
### Docker

**NOTE** The docker implementation will use headless chromium, where as the NPX version will open a browser window.

```json
```

dsp-ant (changes requested):
Thank you for working on this. Quite excited.

src/everart/README.md
```diff
@@ -10,6 +10,25 @@ export EVERART_API_KEY=your_key_here
...
+### Docker
+```json
+{
+  "mcpServers": {
+    "everart": {
+      "command": "docker",
+      "args": ["run", "-i", "--rm", "-e", "EVERART_API_KEY=$EVERART_API_KEY", "ai/mcp-everart"],
+    "env": {
+      "EVERART_API_KEY": "your_key_here"
+    }
```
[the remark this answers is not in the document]
slimslenderslacks: Right, fixed.

src/fetch/Dockerfile
slimslenderslacks: @dsp-ant what you're saying here makes sense. We could potentially create a uv-less python environment and just copy the virtual environment from this stage into a small uv-less final stage.

src/puppeteer/Dockerfile
slimslenderslacks: @ColinMcNeil did you go with bullseye because of chromium?
I know we went with debian over alpine because of the chromium dependency. In the long run, I'd like to discuss a separate container for container but it almost feels like we'd need to explore that as an option first in puppeteer so I think we should leave that out for now.
Colin will know whether bookworm could be an option.

src/gdrive/README.md
slimslenderslacks: Ya, we've added a section on the explicit authentication step (which is also a docker cli command). This is a prerequisite step before adding the gdrive server to mcpServers

src/puppeteer/Dockerfile
ColinMcNeil: Tested with bookworm and it worked fine for the puppeteer MCP

src/gdrive/README.md
```diff
@@ -31,8 +31,9 @@ The server provides access to Google Drive files:
 3. [Configure an OAuth consent screen](https://console.cloud.google.com/apis/credentials/consent) ("internal" is fine for testing)
 4. Add OAuth scope `https://www.googleapis.com/auth/drive.readonly`
 5. [Create an OAuth Client ID](https://console.cloud.google.com/apis/credentials/oauthclient) for application type "Desktop App"
-6. Download the JSON file of your client's OAuth keys
-7. Rename the key file to `gcp-oauth.keys.json` and place into the root of this repo (i.e. `servers/gcp-oauth.keys.json`)
+6. Add http://localhost:3000/oauth2callback as a redirect URI
```
ColinMcNeil, line 34: @dsp-ant Hope you don't mind, but I found this extra step was important to get the mcp server to work regardless of Docker/npx

src/gdrive/README.md
ColinMcNeil: The section is right below this thread:
https://github.com/modelcontextprotocol/servers/pull/368/files#diff-4388c32ec9267085688e3653add6606627eecdac28c0748c171040499695493eR55

src/gdrive/replace_open.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
+
+# Basic script to replace opn(authorizeUrl, { wait: false }).then(cp => cp.unref()); with process.stdout.write(`Open this URL in your browser: ${authorizeUrl}`);
+
+sed -i 's/opn(authorizeUrl, { wait: false }).then(cp => cp.unref());/process.stderr.write(`Open this URL in your browser: ${authorizeUrl}\n`);/' node_modules/@google-cloud/local-auth/build/src/index.js
```
ColinMcNeil, line 5: Build-time workaround for trying to open user's browser from inside the container. Instead of triggering `xdg-open`, this prints a URL to `stderr`:

Screenshot 2024-12-19 at 6 54 35 PM

src/fetch/Dockerfile
slimslenderslacks: @dsp-ant so my summary here is that I think we should keep the uv stage running with two layers (one slow moving one that does not install the project) and a second faster moving one with the project.  This will be a lot faster when we pick up changes.

Your suggestion to run without uv is great.  In the latest Dockerfile, the final stage is uv-less.

src/git/README.md
slimslenderslacks: fixed

src/gdrive/README.md
dsp-ant: Great thank you.

src/gdrive/replace_open.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
+
+# Basic script to replace opn(authorizeUrl, { wait: false }).then(cp => cp.unref()); with process.stdout.write(`Open this URL in your browser: ${authorizeUrl}`);
```
dsp-ant, line 3: Maybe in a follow up, let's add the reasoning you described. future me will thank if this is inside the bash script instead of a github comment :)

src/git/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Use a Python image with uv pre-installed
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim AS uv
```
dsp-ant, line 2: In the longrun we should figure out what it takes to run this on alpine. There is a `ghcr.io/astral-sh/uv:python3.12-alpine`, but we might need to be careful that uv correctly uses system python in that case. Maybe we want to follow up on this PR, but it's not blocking.

dsp-ant (dismissed):
Thank you so much for getting this over the finish line. I am super stoked to have this and can't wait for people to work with MCP servers that can take really interesting dependencies because of docker containers.

A few suggestions:
1. We should probably add a note about the usage of Docker to the top-level readme, including a link to the mcp namespace on docker hub. 
2. I would love to see a section in docs/ on how to build MCP servers with docker.
3. I'd love if we would say on the README that this is maintained by Docker itself.

A few questions:
1. How should we deal with CI? Do we need to add anything to our .github/workflows to trigger builds?
2. Where would users report errors about docker setup. We can help with maintenance but I'd love if there you all could help out. Happy to redirect them to a docker repo or if you want a docker tag in this repo to follow.

src/git/Dockerfile
slimslenderslacks: Okay, maybe let's do this in the form of alpine support for python users who want to create docker mcp servers.

dsp-ant (approved):

PR #393, opened 2024-12-21
link to Fireproof MCP server

jchris:
## Description

Fireproof runs locally inside the MCP server runtime and synchronizes via the cloud, so you have local-first provable ledgers for your agent state.


jspahrsummers (approved):
Cool, thank you!

PR #552, opened 2025-01-22
windows credential file path resolution fixes #301

### LINKED ISSUE #301: [server-gdrive] Incorrect path resolution on Windows causing credentials file not found
## Bug Description
When attempting to authenticate with the Google Drive MCP server on Windows, the package fails to locate the credentials file due to incorrect path resolution. The package appears to be prepending "C:\" to already absolute Windows paths.

## Environment
- OS: Windows
- Node.js Version: v23.3.0
- Package: @modelcontextprotocol/server-gdrive
- Installation Method: Both local (npx) and global (npm install -g) installations tested

## Steps to Reproduce
1. Place valid Google OAuth credentials in [gcp-oauth.keys.json](cci:7://file:///C:/Users/privetin/AppData/Roaming/nvm/v23.3.0/node_modules/gcp-oauth.keys.json:0:0-0:0)
2. Run the authentication command:
   ```bash
   npx -y @modelcontextprotocol/server-gdrive auth
   ```
## Current Behavior
The package fails with an error indicating it cannot find the credentials file. The error message shows that it's incorrectly prepending "C:" to an already absolute Windows path:
```
Error: Cannot find module 'C:\C:\Users\...\gcp-oauth.keys.json'
```
## Expected Behavior
The package should correctly resolve Windows paths without double-prepending the drive letter, allowing it to find the credentials file at its actual location.

## Error Messages
When running locally:
```
Error: Cannot find module 'C:\c:\MCP\servers\gcp-oauth.keys.json'
Require stack:
- c:\MCP\servers\node_modules\@google-cloud\local-auth\build\src\index.js
```
When running globally installed package:
```
Error: Cannot find module 'C:\C:\Users\...\node_modules\gcp-oauth.keys.json'
Require stack:
- C:\Users\...\node_modules\@modelcontextprotocol\server-gdrive\node_modules\@google-cloud\local-auth\build\src\index.js
```
## Additional Context
This appears to be a Windows-specific path resolution issue. The error occurs in the @google-cloud/local-auth dependency when it attempts to load the credentials file. The package is incorrectly handling Windows absolute paths by prepending "C:" to already absolute paths.

## Possible Fix
The package should check if the path is already absolute (starts with a drive letter on Windows) before prepending any drive letter or root directory.

anishp:
Solution provided by @BindingOx 

Fixes path resolution issue for credential file in windows causing auth to fail.

### DIFF
--- a/src/gdrive/index.ts
+++ b/src/gdrive/index.ts
@@ -12,6 +12,7 @@ import {
 import fs from "fs";
 import { google } from "googleapis";
 import path from "path";
+import { fileURLToPath } from 'url';
 
 const drive = google.drive("v3");
 
@@ -176,7 +177,7 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
 });
 
 const credentialsPath = process.env.GDRIVE_CREDENTIALS_PATH || path.join(
-  path.dirname(new URL(import.meta.url).pathname),
+  path.dirname(fileURLToPath(import.meta.url)),
   "../../../.gdrive-server-credentials.json",
 );
 
@@ -184,7 +185,7 @@ async function authenticateAndSaveCredentials() {
   console.log("Launching auth flow…");
   const auth = await authenticate({
     keyfilePath: process.env.GDRIVE_OAUTH_PATH || path.join(
-      path.dirname(new URL(import.meta.url).pathname),
+      path.dirname(fileURLToPath(import.meta.url)),
       "../../../gcp-oauth.keys.json",
     ),
     scopes: ["https://www.googleapis.com/auth/drive.readonly"],


jerome3o-anthropic (approved):
Fantastic - thanks. Additional info from Claude for future reference:

This change looks like a good solution to handle file paths correctly across different operating systems. The modification replaces `new URL(import.meta.url).pathname` with `fileURLToPath(import.meta.url)`, which is the recommended way to convert `file:` URLs to filesystem paths in ES modules.

Here's why this change is good:

1. `fileURLToPath` is specifically designed to handle the conversion from URL to filesystem paths in a cross-platform way. It will:
   - Handle Windows-style paths correctly (including drive letters)
   - Properly decode URL-encoded characters
   - Remove the `file:` protocol prefix
   - Handle relative paths correctly

2. The previous code using `.pathname` directly could have issues on Windows because:
   - URL pathnames always use forward slashes
   - The leading slash in pathnames can interfere with Windows drive letters
   - URL encoding/decoding wasn't handled consistently

This change should work reliably across:
- Windows (properly handling `C:\path\to\file`)
- macOS (`/path/to/file`)
- Linux (`/path/to/file`)

You may also want to add `fileURLToPath` to the imports at the top of the file if it's not already there:

```typescript
import { fileURLToPath } from 'url';
```

PR #607, opened 2025-02-07
fix: detached frame error & feature: Puppeteer launch arguments support

AB498:
# Detached Frame Error Fix and Puppeteer Launch Arguments Support

## Description
This PR fixes the **"detached Frame" error** (`MCP error -32603: Attempted to use detached Frame 'D9B4203EF8575FD5A6ABAFC77F0F8E18'`) encountered in the MCP tool. Additionally, it adds support for specifying **Puppeteer launch arguments**, improving flexibility when initializing browser instances.

## Server Details
- **Server:** `modelcontextprotocol/servers`
- **Changes to:** MCP execution flow and Puppeteer browser launch configuration

## Motivation and Context
- The detached frame error caused failures in MCP tool execution, preventing some interactions.
- Allowing Puppeteer launch arguments gives users control over headless mode, debugging, and other configurations.

## How Has This Been Tested?
- Tested with an MCP client under various execution flows.
- Verified Puppeteer arguments work as expected with different browser setups.

## Breaking Changes
- No breaking changes expected. The fix ensures stability without altering existing APIs.

## Types of Changes
- [ ] Documentation update

## Checklist

## Additional Context
- Puppeteer launch arguments can now be configured via environment variables.
- The fix ensures that MCP no longer attempts to interact with detached frames.

### DIFF
--- a/src/puppeteer/index.ts
+++ b/src/puppeteer/index.ts
@@ -186,7 +186,13 @@ function deepMerge(target: any, source: any): any {
     const targetVal = target[key];
     const sourceVal = source[key];
     if (Array.isArray(targetVal) && Array.isArray(sourceVal)) {
-      output[key] = [...targetVal, ...sourceVal];
+      // Deduplicate args/ignoreDefaultArgs, prefer source values
+      output[key] = [...new Set([
+        ...(key === 'args' || key === 'ignoreDefaultArgs' ?
+          targetVal.filter((arg: string) => !sourceVal.some((launchArg: string) => arg.startsWith('--') && launchArg.startsWith(arg.split('=')[0]))) :
+          targetVal),
+        ...sourceVal
+      ])];
     } else if (sourceVal instanceof Object && key in target) {
       output[key] = deepMerge(targetVal, sourceVal);
     } else {
--- a/src/puppeteer/index.ts
+++ b/src/puppeteer/index.ts
@@ -22,7 +22,9 @@ const TOOLS: Tool[] = [
     inputSchema: {
       type: "object",
       properties: {
-        url: { type: "string" },
+        url: { type: "string", description: "URL to navigate to" },
+        launchOptions: { type: "object", description: "PuppeteerJS LaunchOptions. Default null. If changed and not null, browser restarts. Example: { headless: true, args: ['--no-sandbox'] }" },
+        allowDangerous: { type: "boolean", description: "Allow dangerous LaunchOptions that reduce security. When false, dangerous args like --no-sandbox will throw errors. Default false." },
       },
       required: ["url"],
     },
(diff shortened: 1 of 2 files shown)


kavukcutolga:
I have this issue, is the fix going to be merged ?

src/puppeteer/index.ts
```diff
@@ -22,7 +22,8 @@ const TOOLS: Tool[] = [
     inputSchema: {
       type: "object",
       properties: {
-        url: { type: "string" },
+        url: { type: "string", description: "URL to navigate to" },
+        args: { type: "object", description: "Puppeteer launch options. Default null. If changed and not null, browser restarts. Example: { headless: true, args: ['--no-sandbox'] }" },
```
olaservo, line 26: Should a specific subset of args be validated/allowed here to defend against the LLM hallucinating arbitrary args, or passing args that introduce security concerns?

olaservo (changes requested):
Thanks for fixing the detached frame error and for adding args support!  If we do include the args input here as well, could you also add a note to the Readme about this feature, something like:

```
### Launch Arguments
The server supports configuring Puppeteer launch options:

- **headless** (boolean): Run browser in headless mode
  Example:
  ```json
  {
    "url": "https://example.com",
    "args": {
      "headless": true
    }
  }
```

src/puppeteer/index.ts
AB498: Had to force push since the upstream was way ahead. Anyway, I updated the `README.md` and `index.ts` files as you suggested.

src/puppeteer/index.ts
```diff
@@ -126,6 +177,25 @@ async function ensureBrowser() {
...
+// Deep merge utility function
+function deepMerge(target: any, source: any): any {
+  const output = Object.assign({}, target);
+  if (typeof target !== 'object' || typeof source !== 'object') return source;
+
+  for (const key of Object.keys(source)) {
+    const targetVal = target[key];
+    const sourceVal = source[key];
+    if (Array.isArray(targetVal) && Array.isArray(sourceVal)) {
+      output[key] = [...targetVal, ...sourceVal];
```
olaservo, line 189: Is there a chance that combining these arrays without de-duping might duplicate some args?  Like one potential edge case could be if the LLM sends args in tool params that are already defined in the environment variable?  It might not break anything but it could get confusing.

olaservo:
I think besides the potential edge case I mentioned this is looking good and I will keep an eye out to merge.

src/puppeteer/index.ts
AB498: Added deduplication.

olaservo (approved):

PR #638, opened 2025-02-19
add forevervm

paulgb:
Adds MCP server for [ForeverVM](https://forevervm.com).

ForeverVM is a sessionless code sandbox with a REPL-style interface. Our MCP server allows clients to start a new sandbox and run code in it.


tadasant (approved):
🚀

PR #740, opened 2025-03-05
Introduce executeJSCode Function for Filesystem MCP Server without the need to have external files in context window

freiit:
New function `executeJSCode` that lets run JavaScript against files/directories in the MCP server, not only with the analysis tool. Files are not automatically loaded into the context window to enable the handling of large files.

Key details

* JS execution with lodash, mathjs, PapaParse, XLSX libraries
* Robust error handling for sync/async code
* Proper Node.js environment with fs, path access
* No context pollution: large files won't bloat context, must be explicitly read

The function expands MCP capabilities for complex file analysis and data processing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -695,6 +695,18 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@babel/runtime": {
+      "version": "7.26.9",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.26.9.tgz",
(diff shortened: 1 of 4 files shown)


alfonsograziano:
Ho @freiit, I see that you're experimenting with Node and MCP as well :D 

In case you think it might be helpful, I built this Node.js MCP server which leverages Docker containers and custom images: https://github.com/alfonsograziano/node-code-sandbox-mcp

olaservo:
Hi, thanks for the PR and apologies for not reviewing this sooner.  Looking at these changes, I think we will hold on merging this into the filesystem server at this time, since we're focusing more on stability/security at the moment and making sure the core MCP features work reliably.

PR #759, opened 2025-03-07
Update mount arguments for docker commands

edburns:
…
modified:   src/filesystem/README.md

- According to Claude, it should be source, not src, and target, not dst.

## Server Details

- Server: filesystem
- Changes to: README.md

## Motivation and Context

Without this change, I observe, "Could not attach to MCP server filesystem".

## Types of changes
- [X ] Documentation update

## Checklist
- [X ] I have updated the server's README accordingly

### DIFF
--- a/src/filesystem/README.md
+++ b/src/filesystem/README.md
@@ -120,9 +120,9 @@ Note: all directories must be mounted to `/projects` by default.
         "run",
         "-i",
         "--rm",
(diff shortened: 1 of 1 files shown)


domdomegg:
Heya, these are equivalent in docker:

CleanShot 2025-08-15 at 17 35 18@2x

From https://docs.docker.com/engine/storage/bind-mounts/

I'm going to close this for now, but if you can provide a repro where this isn't equivalent please do raise an issue :)

PR #745, opened 2025-03-05
Fix behaviour of search_files tool

Finndersen:
The `search_files` tool currently has inconsistent and unexpected behaviour. 
Issue: https://github.com/modelcontextprotocol/servers/issues/735

## Description

For some reason, glob pattern matching was only supported for `excludePatterns` argument and not `patterns`, which is unexpected and causes LLMs to get confused and waste a lot of tokens and request cycles figuring out how to use the tool. 

This change improves and standardises the behaviour by:

- Support  glob-style pattern matching in `pattern` argument. The full path to each item is checked (not just the item name like before), so a pattern like `**/*.ext` is required to recursively match files in sub-directories
- No longer automatically convert `excludePatterns` patterns to `**/<pattern>/**` format when no wildcard/* provided
- Update tool description so models can use it properly

## Server Details

- Server:  filesystem
- Changes to: search_files tool

## Motivation and Context
See above

## How Has This Been Tested?
Tested  directly and with Gemini model, with many combinations of `pattern` and `excludePatterns`

## Breaking Changes
Possibly, there is a change in behaviour (not change in tool interface). 

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ x] Breaking change (fix or feature that would cause existing functionality to change)
- [x ] Documentation update

## Checklist

- [ x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [ x] My changes follows MCP security best practices
- [x ] I have updated the server's README accordingly
- [x ] I have tested this with an LLM client
- [ x] My code follows the repository's style guidelines
- [x ] New and existing tests pass locally
- [x ] I have added appropriate error handling
- [ x] I have documented all environment variables and configuration options

PR #623, opened 2025-02-13
feat(directory_tree): add excludePatterns support & documentation

lamemind:
- Update documentation with directory_tree declaration
- Add excludePatterns parameter to DirectoryTreeArgsSchema
- Implement pattern exclusion in buildTree function using minimatch
- Pass excludePatterns through recursive calls
- Support both simple and glob patterns for exclusion
- Maintain consistent behavior with search_files implementation

## Motivation and Context
launching directory_tree inside a code project is a pain. adding exclusion pattern is my solution.

## How Has This Been Tested?
Tested locally

## Breaking Changes
No breaking changes. `excludePatterns` is empty by default.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [X] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [ ] My changes follows MCP security best practices 
- [X] I have updated the server's README accordingly
- [X] I have tested this with an LLM client
- [X] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [X] I have documented all environment variables and configuration options

PR #543, opened 2025-01-18
Windows filesystem MCP enhancements

mfcasazza:
## Description
Added support for Windows file paths with spaces in them.  Also added support for linked directories. 

## Server Details
- Server: filesystem
- Changes to: server implementation and path resolution logic

## Motivation and Context
I ran into this bug: https://github.com/modelcontextprotocol/servers/issues/447 when trying to use the claude desktop app.

## How Has This Been Tested?
Tested extensively with Claude desktop client on Windows using multiple scenarios:
1. Basic directory access (Desktop folder)
2. Access to program files (I used my World of Warcraft directory - C:\Program Files (x86)\World of Warcraft)
3. Confirmed proper security boundaries (access denied for non-allowed paths)
4. Verified directory/file type differentiation is maintained through symbolic links
5. Added tests to address the scenarios in the open bug

## Breaking Changes
No breaking changes. Existing configurations should continue to work as before - I tested locally with my desktop and downloads folders

## Types of changes
- [x] New feature (non-breaking change which adds functionality)
- [] Documentation update

## Checklist
- [x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [x] My changes follows MCP security best practices
- [x] I have updated the server's README accordingly
- [x] I have tested this with an LLM client
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have documented all environment variables and configuration options

## Additional context
I don't have a mac - the code shouldn't affect mac paths - however was not able to test that locally

PR #897, opened 2025-03-17
Rename "search_files" to "search_files_by_name"

### LINKED ISSUE #735: "search_files" tool of Filesystem example MCP server does not work properly
**Describe the bug**
It is expected that the `pattern` argument of the `search_files` tool would support glob/regex style wildcard patterns, e.g. "*.py", but it does not.

Looking at the [implementation](https://github.com/modelcontextprotocol/servers/blob/main/src/filesystem/index.ts#L217) it is easy to see why, for some reason it just does a basic substring inclusion test instead of a proper glob `minimatch()` like it does for the `excludePatterns` argument.

**To Reproduce**
1. Start Filesystem MCP and connect client
2. Try use `search_files` tool providing `pattern` argument with glob/regex style wildcard pattern  e.g. "*.py"
3. get "No matches found" even when there are files that match

**Expected behavior**
`search_files` tool  `pattern` argument should support glob/regex style wildcard patterns

I'm planning on creating a fix for this soon!

### LINKED ISSUE #896: filesystem search_files function returns no matches found
**Describe the bug**
The `search_files` function from the filesystem server often returns "no matches found", even though a local search (e.g. `grep`) does return results.

**To Reproduce**
1. Install Claude desktop for Mac Version 0.8.1
2. Install and configure the `filesystem` server. Cleanup npm cache just to be sure the latest version is installed.
3. Ask Claude desktop to search a known term that actually exists. Search in a large project that has lots of files.
4. Notice the search results return empty results within a few seconds.

```json
{
  `path`: `/Users/my_userid/my_project_path/`,
  `pattern`: `mysearch`
}
No matches found
```

**Expected behavior**
The filesystem search should return results.

**Logs**

**Additional context**

MCP configuration:
```json
{
    "mcpServers": {
      "filesystem": {
        "command": "npx",
        "args": [
          "-y",
          "/@modelcontextprotocol/server-filesystem",
          "/Users/my_userid/my_project_path"
        ]
      }
  }
```

Observations:

1. The search results return empty results within a few seconds.
2. I wrote a script to start debugging the issue:
    1. Script wraps the `searchFiles` in `modelcontextprotocol-servers/src/filesystem/index.ts`.
    2. Search content from the command line using, separate from Claude for desktop.
    3. Notice the search takes bout 45 seconds to return the results for a project that has about 300K lines of code.
    4. So it looks like Claude for desktop may apply a timeout if the search takes too long?
    5. I see opportunities to improve the search performance, with indexing, parallel search, chunk processing, etc

sebastien-rosset:
## Description

1. Fix #896: "Search file by name" enhancements
   1. Rename the existing `search_files` function to `search_files_by_name`
   2. Improve performance for deeply nested directories by processing directories in parallel batches.
   3. Improve description to make it clear it's a file/directory name matcher, not a file content search.
   4. Improve error handling
2. Add new `search_file_contents` to search file content.
   1. Plain text or regex pattern matching
   2. Case-sensitive/insensitive search
   3. Configurable result limits
   4. Context lines around matches
   5. File glob filtering (include/exclude)
3. Better parameter documentation throughout
4. Fix #735, #1067: sometimes the LLM was confused what is the exact syntax for file name matching. Is it substring match, regex match, glob pattern or something else? It wasn't very clear from the name and description. 

## Server Details

- Server: filesystem
- Changes to: tools

## Motivation and Context

Both Claude and human users may misinterpret the `search_files` function as a content search tool (like grep) rather than a file/directory name matcher. The suggested description makes the distinction clear from the beginning and explicitly notes that it doesn't search file contents.

## How Has This Been Tested?

Tested with Claude Desktop with the following prompts, before and after the change.
1. The "Before" column shows which function was selected by the LLM before this PR.
8. The "After" column shows which function was selected by the LLM with this PR, for the exact same prompt.

| Prompt | Before | After |
| --- | --- | --- |
| Find all references to "XYZ" under <br/>/Users/my_userid/my_project/ | search_files | search_file_content |
| Find files whose names contain "color" under <br/>/Users/my_userid/my_project/ | search_files | search_files_by_name |
| Search files that contain "XYZ" under <br/>/Users/my_userid/my_project using  filesystem MCP. | search_files | search_file_content |
| Search "XYZ" under <br/>/Users/my_userid/my_project using  filesystem MCP. | search_files | search_file_content |
| Search "xyz" in file a/b/c/foo.cpp | search_files | search_file_content |

For prompt 1, the LLM can easily infer it should be searching for file contents, not matching file names. Previously, the LLM mistakenly invoked the `search_files` function, which typically would return no matching results.

For prompts 3 and 4, there is intentionally some prompt ambiguity to test how the LLM decide whether to use `search_files_by_name` or `search_file_content` based on the function names and descriptions.

## Breaking Changes

> Will users need to update their MCP client configurations?

No, but I'm not sure I understand breaking changes. I'm unclear whether it's ok to rename a function, or if it's considered to be a breaking change.

## Types of changes

- [x] Documentation update

## Checklist

### Tests for the `search_file_contents` function

- [x] Basic text search returns correct results for common terms
- [x] Case-sensitive search correctly differentiates between uppercase and lowercase patterns
- [x] Regex pattern search correctly identifies patterns like function declarations `async function \w+\(`
- [x] Context lines display works correctly with different values (0, 2, 4 lines)
- [x] Max results parameter properly limits the number of results returned
- [x] File extension filtering correctly includes only the specified file types (.ts, .js, etc.)
- [x] File extension filtering handles files without extensions correctly
- [x] Exclude patterns with wildcards (`**/*.test.ts`) correctly filter out matching files
- [x] Exclude patterns with simple strings (`*config*`) correctly filter content containing those strings
- [x] Combined search with regex + case sensitivity + file types works correctly
- [x] Search handles large files efficiently without excessive memory usage
- [x] Directory recursion works correctly for nested file structures
- [x] Error handling for inaccessible files or directories works as expected
- [x] Race conditions are avoided when searching directories in parallel
- [x] Files with non-UTF8 encoding are handled gracefully without crashing

### DIFF
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -455,14 +455,3 @@ export async function searchFilesByName(
 
   return results;
 }
-
-// Legacy function for backward compatibility
-export async function searchFilesWithValidation(
-  rootPath: string,
-  pattern: string,
-  allowedDirectories: string[],
-  options: SearchOptions = {}
-): Promise<string[]> {
-  const { excludePatterns = [] } = options;
-  return searchFilesByName(rootPath, pattern, excludePatterns);
-}
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -1,4 +1,5 @@
 import fs from "fs/promises";
+import { Dirent } from "fs";
 import path from "path";
 import os from 'os';
 import { randomBytes } from 'crypto';
@@ -348,6 +349,114 @@ export async function headFile(filePath: string, numLines: number): Promise<stri
   }
 }
 
+// Search by name function with glob pattern support
+export async function searchFilesByName(
+  rootPath: string,
+  pattern: string,
+  excludePatterns: string[] = []
+): Promise<string[]> {
+  const results: string[] = [];
+  const queue: string[] = [rootPath];
+  const processedPaths = new Set<string>();
+  const caseSensitive = /[A-Z]/.test(pattern); // Check if pattern has uppercase characters
+
+  // Determine if the pattern is a glob pattern or a simple substring
+  const isGlobPattern = pattern.includes('*') || pattern.includes('?') || pattern.includes('[') || pattern.includes('{');
+
+  // Prepare the matcher function based on pattern type
+  let matcher: (name: string, fullPath: string) => boolean;
+
+  if (isGlobPattern) {
+    // For glob patterns, use minimatch
+    matcher = (name: string, fullPath: string) => {
+      // Handle different pattern types
+      if (pattern.includes('/')) {
+        // If pattern has path separators, match against relative path from root
+        const relativePath = path.relative(rootPath, fullPath);
+        return minimatch(relativePath, pattern, { nocase: !caseSensitive, dot: true });
+      } else {
+        // If pattern has no path separators, match just against the basename
+        return minimatch(name, pattern, { nocase: !caseSensitive, dot: true });
+      }
+    };
+  } else {
+    // For simple substrings, use includes() for better performance
+    const searchPattern = caseSensitive ? pattern : pattern.toLowerCase();
+    matcher = (name: string) => {
+      const nameToMatch = caseSensitive ? name : name.toLowerCase();
+      return nameToMatch.includes(searchPattern);
+    };
+  }
+
+  const compiledExcludes = excludePatterns.map(pattern => {
+    const globPattern = pattern.includes('*') ? pattern : `**/${pattern}/**`;
+    return (path: string) => minimatch(path, globPattern, { dot: true });
+  });
+
+  const shouldExclude = (relativePath: string): boolean => {
+    return compiledExcludes.some(matchFn => matchFn(relativePath));
+  };
+
+  // Process directories in a breadth-first manner
+  while (queue.length > 0) {
+    const currentBatch = [...queue]; // Copy current queue for parallel processing
+    queue.length = 0; // Clear queue for next batch
+
+    // Process current batch in parallel
+    const entriesBatches = await Promise.all(
+      currentBatch.map(async (currentPath): Promise<Dirent[]> => {
+        if (processedPaths.has(currentPath)) return []; // Skip if already processed
+        processedPaths.add(currentPath);
+
+        try {
+          await validatePath(currentPath);
+          return await fs.readdir(currentPath, { withFileTypes: true });
+        } catch (error) {
+          return []; // Return empty array on error
+        }
+      })
+    );
+
+    // Flatten and process entries
+    for (let i = 0; i < currentBatch.length; i++) {
+      const currentPath = currentBatch[i];
+      const entries = entriesBatches[i];
+
+      if (!entries) continue;
+
+      for (const entry of entries) {
+        const fullPath = path.join(currentPath, entry.name);
+        try {
+          // Validate path before processing
+          await validatePath(fullPath);
+
+          // Check exclude patterns (once per entry)
+          const relativePath = path.relative(rootPath, fullPath);
+          if (shouldExclude(relativePath)) {
+            continue;
+          }
+
+          // Apply the appropriate matcher function
+          if (matcher(entry.name, fullPath)) {
+            results.push(fullPath);
+          }
+
+          // Add directories to queue for next batch
+          if (entry.isDirectory()) {
+            queue.push(fullPath);
+          }
+        } catch (error) {
+          // Skip invalid paths
+          continue;
+        }
+      }
+    }
+  }
+
+  return results;
+}
+
+// Legacy function for backward compatibility
 export async function searchFilesWithValidation(
   rootPath: string,
   pattern: string,
@@ -355,38 +464,5 @@ export async function searchFilesWithValidation(
   options: SearchOptions = {}
 ): Promise<string[]> {
   const { excludePatterns = [] } = options;
-  const results: string[] = [];
-
-  async function search(currentPath: string) {
-    const entries = await fs.readdir(currentPath, { withFileTypes: true });
-
-    for (const entry of entries) {
-      const fullPath = path.join(currentPath, entry.name);
-
-      try {
-        await validatePath(fullPath);
-
-        const relativePath = path.relative(rootPath, fullPath);
-        const shouldExclude = excludePatterns.some(excludePattern =>
-          minimatch(relativePath, excludePattern, { dot: true })
-        );
-
-        if (shouldExclude) continue;
-
-        // Use glob matching for the search pattern
-        if (minimatch(relativePath, pattern, { dot: true })) {
-          results.push(fullPath);
-        }
-
-        if (entry.isDirectory()) {
-          await search(fullPath);
-        }
-      } catch {
-        continue;
-      }
-    }
-  }
-
-  await search(rootPath);
-  return results;
+  return searchFilesByName(rootPath, pattern, excludePatterns);
 }
--- a/src/filesystem/README.md
+++ b/src/filesystem/README.md
@@ -7,7 +7,8 @@ Node.js server implementing Model Context Protocol (MCP) for filesystem operatio
 - Read/write files
 - Create/list/delete directories
 - Move files/directories
-- Search files
+- Search for files by name
+- Search within file contents (grep-like functionality)
 - Get file metadata
 - Dynamic directory access control via [Roots](https://modelcontextprotocol.io/docs/learn/client-concepts#roots)
 
@@ -144,14 +145,15 @@ The server's directory access control follows this flow:
     - `destination` (string)
   - Fails if destination exists
 
-- **search_files**
+- **search_files_by_name**
   - Recursively search for files/directories that match or do not match patterns
   - Inputs:
     - `path` (string): Starting directory
-    - `pattern` (string): Search pattern
+    - `pattern` (string): Name pattern to match
     - `excludePatterns` (string[]): Exclude any patterns.
   - Glob-style pattern matching
-  - Returns full paths to matches
+  - Case-insensitive matching
+  - Returns full paths to matching files/directories
 
 - **directory_tree**
   - Get recursive JSON tree structure of directory contents
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -16,7 +16,6 @@ import {
   readFileContent,
   writeFileContent,
   // Search & filtering functions
-  searchFilesWithValidation,
   searchFilesByName,
   // File editing functions
   applyFileEdits,
@@ -310,7 +309,7 @@ describe('Lib Functions', () => {
   });
 
   describe('Search & Filtering Functions', () => {
-    describe('searchFilesWithValidation', () => {
+    describe('searchFilesByName', () => {
       beforeEach(() => {
         mockFs.realpath.mockImplementation(async (path: any) => path.toString());
       });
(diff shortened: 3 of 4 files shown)


sebastien-rosset:
Another typically test scenario is when refactoring code. I've noticed Claude typically tries to read a file if the user has provided the file name, as in the example below:

> I'll help you rewrite the message so that the configuration parameter can be properly translated. Let me first examine the file to understand the context better.
> **View result from read_file from filesystem (local)**
> Result too long, truncated to 100000 characters

If the file is too large, then the LLM tries to search content within the file(s). Previously, it was invoking the `search_file` function, but that always returned empty results, because Claude was confused about the purpose of the `search_file` function. With this PR, the LLM correctly invokes `search_file_contents`:

> Now let me look for any references to translating the parameter name:
> View result from search_file_contents from filesystem (local)

sebastien-rosset:
I've noticed sometimes Claude invokes `search_files_by_name` with the following input:
```
{
  `path`: `/Users/xxxx/myproject/src`,
  `pattern`: `.h`
}
[{"type": "text", "text": "No matches found"}]
```

sebastien-rosset:
I've noticed that Claude AI often tries to invoke the search functions with a regex, most likely because the name of the input argument is "pattern".

FalacerSelene:
Thanks for this, I came here with the same issue that you've solved here. I was going to suggest `find_files` and `grep_files` as the names of the tools here, to match the names of the tools that they behave like, but if `by_name` and `contents` is enough to make it clear to Claude what to do then I'm happy with it.

sebastien-rosset:
> Thanks for this, I came here with the same issue that you've solved here. I was going to suggest `find_files` and `grep_files` as the names of the tools here, to match the names of the tools that they behave like, but if `by_name` and `contents` is enough to make it clear to Claude what to do then I'm happy with it.

I've used this for over a month and it generally works well. I don't mind changing to "find_files" and "grep_files". I'm not sure if it would change the behavior.
Based on my use of Claude for the past several weeks, I have renamed the `pattern` argument to `searchText`. I had noticed sometimes Claude was invoking the tool with a regex, probably because the word "pattern" is strongly associated with "regular expression".

rakelkar:
I tried your change and my LLM was confused.. so I asked it to figure out what was wrong and suggest a change.. there is what it suggested:

The original instructions were a bit open-ended, which led to a minor confusion about parameter names. Here are a couple of improvements that could make it clearer:

- **Explicitly mention parameter names:** The prompt could have stated that **`search_file_contents` uses a parameter named `searchText`** (not `pattern`), and **`search_files_by_name` uses `pattern`**. This hint would directly alert the user to use the correct keys without having to infer from code or error messages.
- **Provide a usage example:** Including a short example of each function’s usage in the prompt would serve as a template. For instance, showing a JSON snippet for each case (similar to the examples above) would remove ambiguity.
- **Clarify the goal:** If the goal was to teach the assistant how to use the tool properly, the prompt might explicitly say, *"Find out the correct parameters by reading the tool’s source code or documentation, then demonstrate a correct usage."* This way, it’s clear that simply guessing (as happened initially) might not be enough, and that examining the definitions is required.

By refining the prompt with these details, the assistant (or developer) would be less likely to misuse the functions and run into errors.

rakelkar:
Here is the prompt hint it suggested:
                            This is a tool that works with the filesystem. 
                            Assistant: **Tips for Using the Filesystem Search Tools:**

                            - **Distinguish Search Functions:** There are two different search functions:
                            - `filesystem.search_files_by_name` – use this to find files or directories by name (uses a **`pattern`** argument).
                            - `filesystem.search_file_contents` – use this to search **inside** files for text (uses a **`searchText`** argument).

                            - **Required Parameters:**
                            - **`path`** (for **both** functions): The directory (or file) path where the search should occur. Must be within an allowed directory (use `filesystem.list_allowed_directories` to see allowed paths).
                            - **`pattern`** (for **name search**): The filename pattern to match (supports wildcards like `*.txt` or substrings).
                            - **`searchText`** (for **content search**): The text string (or regex, if enabled) to search for **within** file contents.

                            - **Optional Parameters:**
                            - *For `search_files_by_name`:* `excludePatterns` (array of patterns to **omit** from the search, e.g., `["node_modules/**"]` to skip certain folders).
                            - *For `search_file_contents`:*
                                - `useRegex` (boolean, default `false`): treat `searchText` as regex if true.
                                - `caseSensitive` (boolean, default `false`): make search case-sensitive if true.
                                - `maxResults` (number, default `100`): limit the number of matches returned.
                                - `contextLines` (number, default `2`): lines of context to show before/after each match.
                                - `includePatterns` (array): patterns to **include only** specific files in the search.
                                - `excludePatterns` (array): patterns to exclude files or directories.

                            - **Important:** **Do not confuse the parameter names.** For content searches, always use `searchText` (not `pattern`). For name searches, use `pattern`. Using the wrong key will result in an error (e.g., a missing required field error for `searchText` if you accidentally used `pattern` in `search_file_contents`).

                            - **Example Usage:**
                            - *Search within files for "TODO":*
                                ```json
                                {
                                "path": "/project/src",
                                "searchText": "TODO",
                                "caseSensitive": false,
                                "excludePatterns": ["node_modules/**", "*.min.js"]
                                }
                                ```
                                (This looks in the `/project/src` directory for the text "TODO", case-insensitive, skipping `node_modules` and any `.min.js` files.)
                            - *Search for files with "config" in their name:*
                                ```json
                                {
                                "path": "/project",
                                "pattern": "*config*",
                                "excludePatterns": ["**/build/**"]
                                }
                                ```
                                (This finds files or folders under `/project` with "config" in the name, excluding any in a `build` directory.)

sebastien-rosset:
> I tried your change and my LLM was confused.. so I asked it to figure out what was wrong and suggest a change.. there is what it suggested:
> 

Out of curiosity, what LLM were you using?

sebastien-rosset:
@olaservo , is there anything that I need to do such that this PR can be considered for review?

Symbolk:
The official MCP Servers cannot guarantee itself even for such simple tools, so better unit test them or consider implementing these basic ones by ourselves.

turabisaas:
I think this needs to be a tool in that server.

davidpfarrell:
Not sure what's holding up this PR, but the tool failing to search for files by extension seems like it should be fixed ...

_mcp_
```
search_files:
Arguments: {"path":"/tmp/work","pattern":"*.png"}
Result: [{"type":"text","text":"No matches found"}]
```

_check results_
```sh
find /tmp/work -name '*.png' | wc -l
       4
```

sebastien-rosset:
> Not sure what's holding up this PR, but the tool failing to search for files by extension seems like it should be fixed ...
> 

According to the [CONTRIBUTING.md](https://github.com/modelcontextprotocol/servers/blob/main/CONTRIBUTING.md), improvements to existing servers like this filesystem enhancement are welcome, but I don't have visibility into the review timelines and I don't know if a repo maintainer will look at this PR.

I'm happy to address any concerns or make adjustments. The fix has been working well in my testing environment for months, and I believe it would benefit the broader MCP community.

sebastien-rosset:
Resolved merge conflicts after #543 was merged.

domdomegg:
Awesome, thanks for the contribution! Sorry we've been very slow getting to this - we're a small team, most of whom are volunteer contributors.

The poor tool description definitely seems like a bug in the filesystem server, and it makes a lot of sense why models are confused by it. I think also the patterns was quite confusing for models previously.

After #623 and #745, I think this should be quite a bit better. Additionally, I've taken a stab at fixing the problem with a more minor description change in #2606 - somewhat inspired by the suggestions in the issue you raised, so thank you.

I think we're trying to keep these servers small and avoid adding net new functionality so I'm hesitant to accept a new tool `search_file_contents` right now (as you have probably realized from your experiences raising this PR, we're stretched for maintainer time so more things to maintain is scary!). The filesystem MCP (and other reference servers) are meant to be a developer reference for how you can build servers, rather than a primary server for people to use - I think maybe we need to do a better job communicating this. We'd be very happy to accept a PR to add links to different filesystem MCPs into the README that explore things like having a `search_file_contents` tool :)

However, I would be happy to approve and merge the parts around improving testing and nested folder traversal performance, if you'd be open to reworking this PR to target that / opening a new PR for that and tagging me? Possibly it'd be worth waiting until #2606 is merged, given I think that's a small one which can go in easily soon - and save you having to resolve yet another set of merge conflicts 😅

Thanks again for putting up with us here, and for thoughtfully raising the issue and PR! And sorry filesystem MCP has been in such a bad state this whole time.

sebastien-rosset:
> Awesome, thanks for the contribution! Sorry we've been very slow getting to this - we're a small team, most of whom are volunteer contributors.
> 
> The poor tool description definitely seems like a bug in the filesystem server, and it makes a lot of sense why models are confused by it. I think also the patterns was quite confusing for models previously.
> 
> After #623 and #745, I think this should be quite a bit better. Additionally, I've taken a stab at fixing the problem with a more minor description change in #2606 - somewhat inspired by the suggestions in the issue you raised, so thank you.
> 
> I think we're trying to keep these servers small and avoid adding net new functionality so I'm hesitant to accept a new tool `search_file_contents` right now (as you have probably realized from your experiences raising this PR, we're stretched for maintainer time so more things to maintain is scary!).

I've removed the `search_file_contents` action in this PR, and I've retained that new action in a separate branch.

> The filesystem MCP (and other reference servers) are meant to be a developer reference for how you can build servers, rather than a primary server for people to use - 

Hmm, the `filesystem` MCP server is used in production by many applications, including Claude desktop, VS code, Cursor and many others. The "filesystem" tool is not "just a reference implementation".

> I think maybe we need to do a better job communicating this. We'd be very happy to accept a PR to add links to different filesystem MCPs into the README that explore things like having a `search_file_contents` tool :)

Interesting, I would have thought it makes sense to extend the existing MCP tool rather than create a new one. This is what I had in mind:

1. Applications that already integrate the filesystem MCP server for file operations would immediately benefit from the search content capability, without having to install yet another MCP tool.
2. File content search is fundamentally a filesystem operation - the LLM is querying the filesystem about what's inside files, just as it can query about file names and metadata.
3. Content search often needs to be combined with file metadata (size, modification time, permissions) and path operations. Having everything in one server eliminates round-trips and enables more efficient combined operations.
4. Mental model: developers think of content search as a filesystem operation (like grep or IDE search). Splitting it into a separate tool creates cognitive overhead - users would need to remember which server handles which file-related operations.

> 
> However, I would be happy to approve and merge the parts around improving testing and nested folder traversal performance, if you'd be open to reworking this PR to target that / opening a new PR for that and tagging me? Possibly it'd be worth waiting until #2606 is merged, given I think that's a small one which can go in easily soon - and save you having to resolve yet another set of merge conflicts 😅

Working on it by first resolving merge conflicts and removing the `search_file_contents` action.

> 
> Thanks again for putting up with us here, and for thoughtfully raising the issue and PR! And sorry filesystem MCP has been in such a bad state this whole time.

src/filesystem/lib.ts
```diff
@@ -348,45 +349,120 @@ export async function headFile(filePath: string, numLines: number): Promise<stri
...
       }
     }
   }
 
-  await search(rootPath);
   return results;
 }
+
+// Legacy function for backward compatibility
+export async function searchFilesWithValidation(
```
domdomegg: Is this used anywhere? I think we can delete?

src/filesystem/lib.ts
```diff
@@ -348,45 +349,120 @@ export async function headFile(filePath: string, numLines: number): Promise<stri
...
+  excludePatterns: string[] = []
 ): Promise<string[]> {
-  const { excludePatterns = [] } = options;
   const results: string[] = [];
+  const queue: string[] = [rootPath];
+  const processedPaths = new Set<string>();
+  const caseSensitive = /[A-Z]/.test(pattern); // Check if pattern has uppercase characters
+
+  // Determine if the pattern is a glob pattern or a simple substring
+  const isGlobPattern = pattern.includes('*') || pattern.includes('?') || pattern.includes('[') || pattern.includes('{');
```
domdomegg, line 364: Potentially breaking change (in the same way as discussed [here](https://github.com/modelcontextprotocol/servers/pull/2606#pullrequestreview-3148767996))

domdomegg: I think if we're going to make breaking changes, we should probably just simplify to only doing glob matching

src/filesystem/lib.ts
sebastien-rosset: You are right. This is a byproduct of repeatedly resolving merge conflicts.

I'm removing `searchFilesWithValidation` entirely and using `searchFilesByName`, which is less ambiguous. I.e. `searchFilesWithValidation` could mean search file/directory by name, OR it could mean file content.

This is in 56a496236547296d14b7bd032278d392f05c849d for now. I will squash everything when all feedback has been addressed.

PR #320, opened 2024-12-12
feat(time): add support for offset style timezones

HudZah:
## Description
This PR adds support for offset style time zones. Currently, they error when you try to run the MCP server and your time zone isn't in an IANA format.

## Server Details

- Server: Time
- Changes to: Add support for offset format timezones (eg: +0530 timezone)

## Motivation and Context
I was running into an issue where when I ran the MCP server time using UVX, I would get a "zone info not found" error with my time zone key in a different format than what the MCP server was expecting. This now adds support for the offset format of time zones.

The error I get: 
`zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key +0530'`

## How Has This Been Tested?

- I added test cases to `time_server_test.py` that test the functionality with various offset timezones
- I ran it locally and tried it on Claude and it works, see below

image

## Breaking Changes

No

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [x] My changes follows MCP security best practices
- [x] I have updated the server's README accordingly
- [x] I have tested this with an LLM client
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have documented all environment variables and configuration options

PR #925, opened 2025-03-19
[time] Cross-platform support for device timezones via `tzlocal`

michaelfromyeg:
I tried to use the time server on Windows, and noticed a timezone error. Windows seems to report the timezone in a strange format...

```plaintext
# windows
>>> import datetime
>>> tzinfo = datetime.datetime.now().astimezone(tz=None).tzinfo
>>> tzinfo
datetime.timezone(datetime.timedelta(days=-1, seconds=61200), 'Pacific Daylight Time')

# linux
>>> import datetime
>>> tzinfo = datetime.datetime.now().astimezone(tz=None).tzinfo
>>> tzinfo
datetime.timezone(datetime.timedelta(days=-1, seconds=61200), 'PDT')
```

These kinds of values didn't play nice with `ZoneInfo`. This PR adds `tzlocal` to fix that.

One thing: I couldn't figure out how to test this reliably. Let me know if you'd like me to add tests.

cc @maledorak for viz!

edit: also https://github.com/modelcontextprotocol/servers/pull/320 is probably worth another glance 👀 

## Server Details

- Server: time
- Changes to: the `get_local_tz` helper

## Motivation and Context

- Fixes a bug when starting the server on windows (`zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key Pacific Daylight Time'`) via `tzlocal`

## How Has This Been Tested?

- Tested manually on Windows; working as expected

image

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/time/src/mcp_server_time/server.py
+++ b/src/time/src/mcp_server_time/server.py
@@ -2,7 +2,7 @@ from datetime import datetime, timedelta
 from enum import Enum
 import json
 from typing import Sequence
-
+from tzlocal import get_localzone
 from zoneinfo import ZoneInfo
 from mcp.server import Server
 from mcp.server.stdio import stdio_server
@@ -39,10 +39,18 @@ def get_local_tz(local_tz_override: str | None = None) -> ZoneInfo:
     if local_tz_override:
         return ZoneInfo(local_tz_override)
 
-    # Get local timezone from datetime.now()
+    # First, try to get local timezone from tzlocal
+    try:
+        local_tz = get_localzone()
+        return ZoneInfo(str(local_tz))
+    except Exception as e:
+        pass
+
+    # If that fails, try getting local timezone from datetime.now()
     tzinfo = datetime.now().astimezone(tz=None).tzinfo
     if tzinfo is not None:
         return ZoneInfo(str(tzinfo))
+
     raise McpError("Could not determine local timezone - tzinfo is None")
 
 
--- a/src/time/README.md
+++ b/src/time/README.md
@@ -2,12 +2,11 @@
 
 A Model Context Protocol server that provides time and timezone conversion capabilities. This server enables LLMs to get current time information and perform timezone conversions using IANA timezone names, with automatic system timezone detection.
 
-### Available Tools
+## Available Tools
 
 - `get_current_time` - Get current time in a specific timezone or system timezone.
   - Required arguments:
     - `timezone` (string): IANA timezone name (e.g., 'America/New_York', 'Europe/London')
-
 - `convert_time` - Convert time between timezones.
   - Required arguments:
     - `source_timezone` (string): Source IANA timezone name
@@ -16,10 +15,10 @@ A Model Context Protocol server that provides time and timezone conversion capab
 
 ## Installation
 
-### Using uv (recommended)
+### Using `uv` (recommended)
 
 When using [`uv`](https://docs.astral.sh/uv/) no specific installation is needed. We will
-use [`uvx`](https://docs.astral.sh/uv/guides/tools/) to directly run *mcp-server-time*.
+use [`uvx`](https://docs.astral.sh/uv/guides/tools/) to directly run `mcp-server-time*.
 
 ### Using PIP
 
@@ -42,7 +41,7 @@ python -m mcp_server_time
 Add to your Claude settings:
 
 <details>
-<summary>Using uvx</summary>
+<summary>Using <code>uvx</code></summary>
 
 ```json
 "mcpServers": {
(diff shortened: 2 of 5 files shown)


maledorak:
Hey @michaelfromyeg , thanks for cc :)
Sadly, I don't have a Windows machine, and I can't test this code on Windows :/

michaelfromyeg:
ah all good!

olaservo:
Thanks for the PR!  I'm wondering if we should try to combine this approach and the approach in https://github.com/modelcontextprotocol/servers/pull/1272 to cover all our bases (I left a comment in that one).

michaelfromyeg:
I trust your judgement on which is best! Not sure how much fallback-ing is necessary... but better safe than sorry?

Candidly, tzlocal was what came up first when I searched up the issue (via Claude!).

olaservo:
@michaelfromyeg since your PR includes some doc formatting updates too and was created first, let's do this:

- Could you add some tests to the existing tests to cover a few cases related to this fix?
- Then we can merge this PR and the other PR can get updated to add any additional stuff not covered by this PR.

src/time/src/mcp_server_time/server.py
```diff
@@ -39,10 +39,18 @@ def get_local_tz(local_tz_override: str | None = None) -> ZoneInfo:
     if local_tz_override:
         return ZoneInfo(local_tz_override)
```
olaservo, line 40: Suggestion to add try/catch if the local override results in an exception, plus it looks like McpError expects an object with a message and not a string:
```suggestion
    if local_tz_override:
        try:
            return ZoneInfo(local_tz_override)
        except Exception as e:
            error_data = ErrorData(
                code=INVALID_PARAMS,
                message=f"Invalid timezone: {str(e)}"
            )
            raise McpError(error_data)
```

olaservo: You will also need to add ErrorData and INVALID_PARAMS fom mcp.types, for example:

`from mcp.types import Tool, TextContent, ImageContent, EmbeddedResource, ErrorData, INVALID_PARAMS`

src/time/src/mcp_server_time/server.py
```diff
@@ -39,10 +39,18 @@ def get_local_tz(local_tz_override: str | None = None) -> ZoneInfo:
...
+        return ZoneInfo(str(local_tz))
+    except Exception as e:
+        pass
+
+    # If that fails, try getting local timezone from datetime.now()
     tzinfo = datetime.now().astimezone(tz=None).tzinfo
     if tzinfo is not None:
         return ZoneInfo(str(tzinfo))
+
     raise McpError("Could not determine local timezone - tzinfo is None")
```
olaservo, line 54:
```suggestion
    error_data = ErrorData(
        code=INVALID_PARAMS,
        message="Could not determine local timezone - tzinfo is None"
    )
    raise McpError(error_data)
```

olaservo: I can't leave comments or suggestions on unchanged lines, but I think you also want to change this in the `get_zoneinfo` function:

```
    except Exception as e:
        error_data = ErrorData(
            code=INVALID_PARAMS,
            message=f"Invalid timezone: {str(e)}"
        )
        raise McpError(error_data)
```

olaservo (changes requested):
Hi, was running the tests for this locally and ran into some errors related to McpError so I left a few suggestions to resolve those.

Also a suggestion for a test to add that would check for valid vs invalid overrides:

```

@pytest.mark.parametrize(
    "test_time,local_tz_override,expected_error",
    [
        # Test Windows format timezone override
        (
            "2024-01-01 12:00:00+00:00",
            "America/Los_Angeles",  # What Windows "Pacific Standard Time" should map to
            None,
        ),
        # Test direct IANA name
        (
            "2024-01-01 12:00:00+00:00", 
            "Europe/Berlin",  # What "Mitteleuropäische Zeit" should map to
            None,
        ),
        # Test invalid timezone
        (
            "2024-01-01 12:00:00+00:00",
            "Invalid/Timezone",
            "Invalid timezone: 'No time zone found with key Invalid/Timezone'",
        ),
    ],
)
def test_get_local_tz_override(test_time, local_tz_override, expected_error):
    """Test local timezone override functionality"""
    with freeze_time(test_time):
        if expected_error:
            with pytest.raises(McpError, match=expected_error):
                get_local_tz(local_tz_override)
        else:
            result = get_local_tz(local_tz_override)
            assert isinstance(result, ZoneInfo)
            assert str(result) == local_tz_override
```

olaservo:
Hi, thanks for the PR.  It looks like the same issue was addressed in a different PR so we're going to close this one.

PR #931, opened 2025-03-19
feat: expose GitHub server instance to npm library

sd2k:
## Description

This commit separates the `main` logic into a separate file, now marked as the binary entry point for the package, and updates the index to export the server instance. This allows the package to be used as a library, enabling the server to be used with custom transports in other projects.

## Server Details

- Server: Github
- Changes to: library

## Motivation and Context

I'd like to be able to import the server in my own `main.ts` and provide a custom transport, rather than just stdio/SSE.

## How Has This Been Tested?

I've tested this with Claude Desktop using the Docker image & command, but I can't tell what the incantation is to test the local checkout with `npx` unfortunately. Any advice would be great.

## Breaking Changes
I don't think so, since the [docs](https://docs.npmjs.com/cli/v8/commands/npx#description) for `npx` state:

> ... npm will attempt to determine the executable name from the package specifier provided as the first positional argument according to the following heuristic:
>
> 1. If the package has a single entry in its bin field in package.json, or if all entries are aliases of the same command, then that command will be used.

So I'm pretty sure anyone with the `npx` command configured will get the same usage as before.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This only changes one server to reduce the scope of the PR, but I'm happy to go through all the others and make a similar change, if that's agreeable. Let me know :+1:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5400,7 +5400,7 @@
         "zod-to-json-schema": "^3.23.5"
       },
       "bin": {
-        "mcp-server-github": "dist/index.js"
+        "mcp-server-github": "dist/main.js"
       },
       "devDependencies": {
         "shx": "^0.3.4",
--- a/src/github/Dockerfile
+++ b/src/github/Dockerfile
@@ -20,4 +20,4 @@ WORKDIR /app
 
 RUN npm ci --ignore-scripts --omit-dev
 
-ENTRYPOINT ["node", "dist/index.js"]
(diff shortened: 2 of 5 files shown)


cliffhall:
**Deprecation Notice:** Development for the `github-server` has been moved to GitHub in the [http://github.com/github/github-mcp-server](https://github.com/github/github-mcp-server) repo. PRs to `github-server` in this repo are no longer being merged.

PR #1012, opened 2025-03-22
Files and paths can be disallowed inside of directories

cantino:
This code was entirely written by Claude, with incremental testing by the human.
## Description
Added the ability to exclude specific paths from filesystem access using patterns in the arguments list.

## Server Details
- Server: filesystem
- Changes to: Command-line arguments processing, path validation, exclusion logic

## Motivation and Context
Users needed more control over which files and directories could be accessed within an allowed directory. This change allows specifying excluded paths (e.g., `.env` files, build directories, `.git` directories) that should never be accessible, even if they exist within an allowed directory.

The feature provides enhanced security by allowing fine-grained access control. For instance, a user can allow access to a project directory while preventing access to sensitive files or directories within it.

## How Has This Been Tested?
- Tested with Docker configuration to verify exclusion patterns work correctly
- Verified that negation patterns (e.g., `!.git`) can re-enable access to otherwise excluded paths
- Confirmed case-insensitive matching works (e.g., `TSCONFIG.json` excludes both uppercase and lowercase variants)
- Validated subdirectory exclusions (e.g., `test/foo` excludes that specific subdirectory)
- Tested all filesystem operations (read, write, list, etc.) to ensure excluded paths are consistently blocked

## Breaking Changes
Existing configurations without exclusion patterns will continue to work as before, **except** that now `.git` and `node_modules` are excluded by default.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
The implementation supports the following features:

1. **Default Exclusions**: `.git` and `node_modules` are excluded by default for security reasons.

2. **Custom Exclusions**: Add exclusions with the syntax `/path/to/dir!file1,dir2,path/to/file3`.

3. **Negation Patterns**: Override default or higher-level exclusions by prefixing with `!`, e.g., `!.git` to allow access to the `.git` directory even though it's excluded by default.

4. **Case-Insensitive Matching**: All exclusion patterns are case-insensitive for security on all operating systems.

5. **Path-Based Exclusions**: Support for excluding specific paths like `test/foo` to block access to subdirectories.

Example usage:
```json
{
  "mcpServers": {
    "filesystem": {
      "command": "npx",
      "args": [
        "-y",
        "@modelcontextprotocol/server-filesystem",
        "/path/to/allowed/dir!.env,dist,secrets"
      ]
    }
  }
}

### CODE REVIEW

File: src/filesystem/index.ts [hunk 1]
```diff
@@ -34,26 +45,117 @@ function expandHome(filepath: string): string {
   return filepath;
 }
 
-// Store allowed directories in normalized form
-const allowedDirectories = args.map(dir =>
-  normalizePath(path.resolve(expandHome(dir)))
-);
+// Default exclusion patterns that always apply
+const DEFAULT_EXCLUSIONS = ['.git', 'node_modules'];
```
olaservo, line 49:
```suggestion
const DEFAULT_EXCLUSIONS = [
'.git',
'node_modules',
'.env*',
'id_rsa',
'id_dsa',
'*.pem'
];
```
(So this could also exclude common secrets/key files by default)

src/filesystem/index.ts
```diff
@@ -62,33 +164,136 @@ async function validatePath(requestedPath: string): Promise<string> {
...
+        break;
+      }
+
+      // Direct filename comparison for case-insensitive matching
+      for (const exclusion of config.exclusions) {
+        // Skip directory exclusions for this check
+        if (!exclusion.includes('/') && !exclusion.includes('*')) {
+          // Compare filenames case-insensitively
+          if (requestedFilename.toLowerCase() === exclusion.toLowerCase()) {
+            throw new Error(`Access denied - path not accessible`);
```
olaservo, line 192: I think we should make it clearer to the LLM that its not accessible on purpose, something like: `Access denied - path matches exclusion pattern`

src/filesystem/index.ts
```diff
@@ -62,33 +164,136 @@ async function validatePath(requestedPath: string): Promise<string> {
...
+          }
+        }
+      }
+
+      // Check if the path matches any exclusion pattern
+      if (isPathExcluded(config.path, relativePath, config.exclusions)) {
+        // Use a generic message that doesn't reveal if the file exists
+        throw new Error(`Access denied - path not accessible`);
```
olaservo, line 200: Same here and with other new errors being added.

src/filesystem/index.ts
```diff
@@ -18,7 +18,18 @@ import { minimatch } from 'minimatch';
 // Command line argument parsing
 const args = process.argv.slice(2);
 if (args.length === 0) {
-  console.error("Usage: mcp-server-filesystem <allowed-directory> [additional-directories...]");
+  console.error("Usage: mcp-server-filesystem <allowed-directory>[!excluded1,excluded2,...] [additional-directories...]");
+  console.error("");
+  console.error("Examples:");
```
olaservo, line 23: I think as long as the tool descriptions have good descriptions for the LLM then you don't need to add this logging.

olaservo (changes requested):
Thanks for the PR!  Had some minor comments.  Also I was trying to compare this to this other PR: https://github.com/modelcontextprotocol/servers/pull/623 to see if this approach would be consistent with that older one which I'm in the process of looking at, too.

olaservo:
Hi, wanted to check if I should close this or if you're still planning on merging these changes (I see some comments didn't get addressed).  Thanks!

src/filesystem/index.ts
cantino: I don't think this has support for `*` in the paths, but I'm not sure.

olaservo:
Going to close for now, but feel free to re-open if you would still like to update and merge this.

PR #1242, opened 2025-04-03
fix: standardize search parameter naming to 'q'

### LINKED ISSUE #1080: GitHub MCP Server: "q" vs "query" in parameters
some actions like `search_repositories` use `query` parameters for query, but other actions use `q` parameter. This inconsistency confuses the LLM and it often makes a mistake and uses `q` where `query` should be used

soumilsuri:
## Description
This PR fixes an inconsistency in parameter naming between different search operations. Some operations were using `query` (specifically in `SearchRepositoriesSchema`) while others were using `q`. This caused confusion for LLMs when using these tools.

## Changes
- Changed parameter name in `SearchRepositoriesSchema` from `query` to `q`  
- Updated the `searchRepositories()` function signature to match
- Ensures consistent naming with other search operations and GitHub's API

## Testing
- Verified schema validation works correctly with the new parameter
- Confirmed validation fails appropriately with the old parameter name

## Related Issues
Fixes #1080  (GitHub MCP Server: "q" vs "query" in parameters)

This change is a straightforward parameter naming standardization to improve consistency across the codebase.

### DIFF
--- a/src/github/operations/repository.ts
+++ b/src/github/operations/repository.ts
@@ -11,7 +11,7 @@ export const CreateRepositoryOptionsSchema = z.object({
 });
 
 export const SearchRepositoriesSchema = z.object({
-  query: z.string().describe("Search query (see GitHub search syntax)"),
+  q: z.string().describe("Search query (see GitHub search syntax)"),
   page: z.number().optional().describe("Page number for pagination (default: 1)"),
   perPage: z.number().optional().describe("Number of results per page (default: 30, max: 100)"),
 });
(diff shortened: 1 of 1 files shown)


olaservo (commented):
A couple things:
- This appears to be a breaking change but I don't see that called out in the description?
- `query` seems more user-friendly than just `q`?

cliffhall:
**Deprecation Notice:** Development for the `github-server` has been moved to GitHub in the http://github.com/github/github-mcp-server repo. PRs to `github-server` are no longer being merged.

Issue #1273, opened 2025-04-05
[mcp/memory] read_graph can't be called without args


ajoslin103:
When my windsurf calls the read_graph tool from the mcp/memory server the tool call fails with:

    Output: No arguments provided for tool: read_graph

Problem is that the code [index.js] guards against missing args

```
  if (!args) {
    throw new Error(`No arguments provided for tool: ${name}`);
  }
```

But read_graph does not take any args

```
    case "read_graph":
      return { content: [{ type: "text", text: JSON.stringify(await knowledgeGraphManager.readGraph(), null, 2) }] };
```

ajoslin103:
PR fixing this issue: https://github.com/modelcontextprotocol/servers/pull/1276

PR #1276, opened 2025-04-06
calling [mcp/memory] read_graph without args 

ajoslin103:
fixes: https://github.com/modelcontextprotocol/servers/issues/1273

Problem is that the mcp/memory code [index.js] guards against missing args

```
  if (!args) {
    throw new Error(`No arguments provided for tool: ${name}`);
  }
```

But read_graph does not take any args

```
    case "read_graph":
      return { content: [{ type: "text", text: JSON.stringify(await knowledgeGraphManager.readGraph(), null, 2) }] };
```


jwtracy:
@ajoslin103 's fix worked for me so I can answer how I tested on their behalf,

Using the https://github.com/mark3labs/mcphost project, I created my own for, added @ajoslin103 's remote, checked out their incoming branch here, and altered my mcphost config file, `~/.mcp.json`, from the `npx @modelcontextprotocol/memory` command to the following:
```
    "memory": {
      "command": "node",
      "args": [
        "/Users/jwtracy/Source/jwtracy/servers/src/memory/index.ts"
      ]
    },
```

Before changing the config file, the steps I took to reproduce the error were to provide the memory prompt context to the chat before this interaction resulting in an error:

```
  You: tell me everything you remember
2025/04/30 09:31:11 INFO 🔧 Using tool name=memory__read_graph

Error calling tool read_graph: No arguments provided for tool: read_graph
```

After I updated my mcphost config file to the above and providing the memory mcp prompt context, I was able to avoid the error:

```
  You: tell me everything you remember
2025/04/30 10:23:12 INFO 🔧 Using tool name=memory__read_graph

  Assistant:


  I don't have any information stored in my memory yet.
```

Note, it has no memory since I just cloned this repo and the `memory.json` file is not yet present.

I scoured the mark3labs mcp and mcphost projects for an appropriate place to fix this and found that this server response was the source of the error.

jwtracy:
@olaservo PTAL when you get a chance.

oguzhankrcb:
please merge this request immediately it fixes the issue.

olaservo (changes requested):
Hi, thanks for the PR and sorry for missing the ping earlier.  Could you update the sdk in the package.json to `"@modelcontextprotocol/sdk": "^1.15.1"` while we're at it, since its on a very old SDK version?  Thanks!

olaservo (approved):

PR #1326, opened 2025-04-08
feat(mcp-server-git): Add context_lines to git_diff* tools

mceachen:
- Added `context_lines` parameter to `git_diff_unstaged`, `git_diff_staged`, and `git_diff` functions
- Updated README.md to document the new environment variable and its implications.

## Server Details

- Server: `git`
- Changes to: `git_diff_unstaged`, `git_diff_staged`, and `git_diff`.

## Motivation and Context

The standard unified git diff only includes 3 lines of context. When Claude is being asked to review pending changes, it is blind to the remainder of any touched file unless the lines of context are increased.

At the limit, a `context_lines` value that is the same size of your largest source file means every touched file gets automatically included in the context. Claude seems to handle this with aplomb and provides strictly better git review guidance. ChatGPT 4o seems to get distracted by the larger diff contexts, so YMMV.

## How Has This Been Tested?
MCP Inspector and vscode/copilot

## Breaking Changes
No--by default, no behavior changes with this PR. 

Users simply add a reference to setting "context_lines" in their prompt.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

The first commit used an environment variable. The current PR uses a proper parameter, which is more flexible and seems to be more idiomatic.


mceachen:
(Oof: I clicked the "update with rebase" button next to "This branch is out-of-date with the base branch" in GitHub, and it did terrible things. I just reverted with a second force-push).

@olaservo anything I can do to expedite? (It really does substantively improve Claude's code reviews quality!)

olaservo:
Hi @mceachen just getting to this now, sorry for the miss on the earlier tag.  Should the `DEFAULT_CONTEXT_LINES` also be able to be changed using an ENV variable rather than being hard-coded?

olaservo:
Adding more from discord convo for visibility: to clarify, I think it makes sense to make this something that can get set as an optional parameter. So re: my env variable question I was more thinking that when the parameter isn't being set, does it make sense for the default to be configurable in the server config.
from what I can tell, the current behavior without your fix (i.e. default without parameters) is decided by git outside of the server/tool, and that is what you're using as your hard-coded default (3)
so I'm just thinking of a situation where someone wants a longer number of lines by default?

olaservo (approved):
Thanks for the PR, will leave the environment variable option as a future enhancement since the default is already the same as the git default.

PR #1397, opened 2025-04-13
Sequentialthinking add container name to docker example

KyloJorgensen:
## Server Details

- Server: sequentialthinking
- Changes to: README

## Motivation and Context
Without a name on the docker container its creates a new container every time.


domdomegg:
Thanks for the PR! The `--name sequentialthinking` isn't actually needed since we're using `--rm`. The container gets auto-removed when it exits, so the custom name only exists temporarily anyway. Docker will just generate a random name if we omit it, which works perfectly fine for MCP servers.

PR #1400, opened 2025-04-13
Update server.py

hariprasad-t:
Fix local timezone detection for non-IANA timezone abbreviations

## Description
This PR fixes an issue where the get_local_tz function fails when encountering timezone abbreviations like IST (Indian Standard Time) that don't directly map to IANA timezone names. The implementation now includes a fallback mechanism that maps UTC offsets to appropriate IANA timezone names when direct conversion fails.

## Server Details

- Server: mcp_server_time
- Changes to: modified timezone detection functionality

## Motivation and Context
The current implementation attempts to convert local timezone information directly to an IANA timezone name, which works for many cases but fails for certain abbreviations (like IST). This causes the server to fail when running in environments using these timezones. By implementing a fallback mechanism based on UTC offset, we ensure the server works consistently across all timezone environments.

## How Has This Been Tested?
- Tested on systems with IST (Indian Standard Time) as local timezone
- Verified time conversion works correctly with the new implementation
- Confirmed behavior is unchanged for systems with standard IANA timezone names

## Breaking Changes
No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/time/src/mcp_server_time/server.py
+++ b/src/time/src/mcp_server_time/server.py
@@ -40,10 +40,30 @@ def get_local_tz(local_tz_override: str | None = None) -> ZoneInfo:
         return ZoneInfo(local_tz_override)
 
     # Get local timezone from datetime.now()
-    tzinfo = datetime.now().astimezone(tz=None).tzinfo
+    local_dt = datetime.now().astimezone(tz=None)
+    tzinfo = local_dt.tzinfo
+    
     if tzinfo is not None:
-        return ZoneInfo(str(tzinfo))
-    raise McpError("Could not determine local timezone - tzinfo is None")
+        try:
+            # Try to get the IANA timezone name directly
+            return ZoneInfo(str(tzinfo))
+        except Exception:
+            # Fallback: Use the UTC offset to determine an appropriate timezone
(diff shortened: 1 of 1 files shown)


src/time/src/mcp_server_time/server.py
```diff
@@ -40,10 +40,30 @@ def get_local_tz(local_tz_override: str | None = None) -> ZoneInfo:
...
+    local_dt = datetime.now().astimezone(tz=None)
+    tzinfo = local_dt.tzinfo
+    
     if tzinfo is not None:
-        return ZoneInfo(str(tzinfo))
-    raise McpError("Could not determine local timezone - tzinfo is None")
+        try:
+            # Try to get the IANA timezone name directly
+            return ZoneInfo(str(tzinfo))
+        except Exception:
```
jamesbyars-svg, line 50: I don't think you can just swallow an `Exception` here and assume you can just parse something other than `tzinfo` a different way...

domdomegg:
Thanks for your contribution! This looks like a duplicate of #1761. We're tracking this there, so I'm going to close this one to keep the discussion consolidated. Feel free to add any additional context or thoughts to the original issue if you have them!

PR #1423, opened 2025-04-14
Add periodic stderr messages.

cliffhall:
## Description
In everything.ts
- send a timestamped 'notifications/stderr' message to the client every 10 seconds

## Server Details

- Server: everything
- Changes to: periodic messages sent to client

## Motivation and Context

The everything server is meant to exercise all the MCP features, and is the goto server for testing features in the Inspector. 

One of the features we don't exercise is the sending of `notifications/stderr` messages. They show up in the sidebar, and we had to add special notification handling for that message type in the past.

When [a new PR](https://github.com/modelcontextprotocol/inspector/pull/286) came in to the inspector to add a Clear button for clearing those messages, I had no obvious way to test it. 

## How Has This Been Tested?

Using the inspector:

https://github.com/user-attachments/assets/b269c490-bd9c-4768-87ca-82c3d76b343d

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/everything/everything.ts
+++ b/src/everything/everything.ts
@@ -162,7 +162,8 @@ export const createServer = () => {
       hour: '2-digit',
       minute: '2-digit',
       second: '2-digit'
-    });    server.notification({
+    });
+    server.notification({
       method: "notifications/stderr",
       params: { content: `${shortTimestamp}: A stderr message` },
     });
--- a/src/everything/everything.ts
+++ b/src/everything/everything.ts
@@ -114,8 +114,9 @@ export const createServer = () => {
 
   let subscriptions: Set<string> = new Set();
   let subsUpdateInterval: NodeJS.Timeout | undefined;
-  // Set up update interval for subscribed resources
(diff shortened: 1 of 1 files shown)


olaservo (commented):
Should we make this something that can be disabled, in case the user doesn't want notifications firing?

cliffhall:
> Should we make this something that can be disabled, in case the user doesn't want notifications firing?

I thought about having an affordance to turn that off in the inspector, but it would mean nothing to any other server you try it with. The everything server is just a good way to see everything in action, so you can look at it to say "oh, that's how that part works". There isn't anything else to do with it really. We could dial back the frequency to say every 30 seconds or something. It would be enough to show the area where error notifications are shown and the clear button that can be used. They don't need to pile up. I'll try that. Come to speak of it the randomly-leveled log notifications are a little frequent as well.

cliffhall:
Adjusted intervals for outgoing demo messages
- subscription updates: 10 seconds (was 5)
- logging messages: 20 seconds (was 10)
- stderr messages: 30 seconds (was 15)

PR #1484, opened 2025-04-17
Fix bug with stack overflow on console calls

iliakan:
Fixed a bug with the multiple tool call within a single window. It was overwriting console with its own methods, so that `console.log` was causing a stack overflow due to calling itself.


olaservo:
Thanks for the PR!  Could you add more info on how you tested this?

Issue #1247, opened 2025-04-04
Update Everything server to use version 2025-03-26 support for HTTP transport


Dogacel:
**Is your feature request related to a problem? Please describe.**
Currently everything server only supports the SSE based implementation, 2024-11-05. Considering it is a reference server, I think it should support the latest version of the protocol as well.

**Describe the solution you'd like**
Support for the latest HTTP transport protocol.

**Describe alternatives you've considered**
Maybe implement both versions side by side instead of fully replacing

localden:
Closing as a duplicate of #576; let's track it there.

PR #1496, opened 2025-04-19
add streamableHttp server support for everything server

shivdeepak:
## Description

Since Streamable HTTP transport landed on `typescript-sdk` a few days back, this PR uses that and adds Streamable HTTP server support to everything server. This directly addresses the current open issue: https://github.com/modelcontextprotocol/servers/issues/1247

## Server Details

- Server: everything
- Changes to: transport

## Motivation and Context

I am also trying to add support for Streamable HTTP on [inspector](https://github.com/modelcontextprotocol/inspector/pull/294), and `everything` server reference implementation would help testing it out.

## How Has This Been Tested?

The server can be started with the new entry point `npm run start:streamableHttp`.

I tested it with the development version [inspector](https://github.com/modelcontextprotocol/inspector/pull/294) with Streamable HTTP support to ensure functionality works.

Screenshot 2025-04-18 at 6 32 55 PM

## Breaking Changes
No. This is an additional server, separate from the existing [SSE](https://github.com/modelcontextprotocol/servers/blob/main/src/everything/sse.ts) implementation.

## Types of changes
- [ ] Documentation update

## Checklist


shivdeepak:
@cliffhall @Dogacel, please take a look.

Dogacel:
> The server can be started with the new entry point npm run start:streamableHttp.

Can we document this in code? That was one thing that gave me a hard time to get started.

shivdeepak:
@Dogacel updated the docs. I am unsure if this could be achieved with `npx` after publishing the package. Thoughts?

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,151 @@
...
+          message: 'Bad Request: No valid session ID provided',
+        },
+        id: null,
+      });
+      return;
+    }
+
+    // Handle the request with existing transport - no need to reconnect
+    // The existing transport is already connected to the server
+    await transport.handleRequest(req, res, req.body);
```
cliffhall, line 70: Did you find it necessary to pass the `req.body` argument?

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,151 @@
...
+      });
+    }
+  }
+});
+
+// Handle GET requests for SSE streams (using built-in support from StreamableHTTP)
+app.get('/mcp', async (req: Request, res: Response) => {
+  const sessionId = req.headers['mcp-session-id'] as string | undefined;
+  if (!sessionId || !transports[sessionId]) {
+    res.status(400).send('Invalid or missing session ID');
```
cliffhall, line 90: Why not send this as a proper `JSONRPCMessage` like the others?

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const transport = transports[sessionId];
+  await transport.handleRequest(req, res);
+});
+
+// Handle DELETE requests for session termination (according to MCP spec)
+app.delete('/mcp', async (req: Request, res: Response) => {
+  const sessionId = req.headers['mcp-session-id'] as string | undefined;
+  if (!sessionId || !transports[sessionId]) {
+    res.status(400).send('Invalid or missing session ID');
```
cliffhall, line 110: Why not send this as a proper `JSONRPCMessage` like the others?

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  console.log(`Received session termination request for session ${sessionId}`);
+
+  try {
+    const transport = transports[sessionId];
+    await transport.handleRequest(req, res);
+  } catch (error) {
+    console.error('Error handling session termination:', error);
+    if (!res.headersSent) {
+      res.status(500).send('Error processing session termination');
```
cliffhall, line 122: Why not send this as a proper `JSONRPCMessage` like the others?

cliffhall (changes requested):
Hey @shivdeepak, thanks for this. I tested it with your Inspector changes and it worked flawlessly. Just a few requests on message shape.

shivdeepak:
@cliffhall thanks for the review. cleaned it all up, and tested it.

cliffhall:
> @cliffhall thanks for the review. cleaned it all up, and tested it.

Not working now.

```shell
Error from MCP server: Error: Error POSTing to endpoint (HTTP 400): {"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"InternalServerError: stream is not readable"},"id":null}
    at StreamableHTTPClientTransport.send (file:///Users/cliffhall/Projects/mcp-inspector/node_modules/@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js:263:23)
```

Failing because `id` is `null` I believe.
Screenshot 2025-04-19 at 5 02 35 PM

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,172 @@
...
+  } catch (error) {
+    console.error('Error handling MCP request:', error);
+    if (!res.headersSent) {
+      res.status(500).json({
+        jsonrpc: '2.0',
+        error: {
+          code: -32603,
+          message: 'Internal server error',
+        },
+        id: null,
```
cliffhall, line 80: `id` can't be null, a number could collide with another message with a valid id, but it can be a string. Let's make it `error-32603`?

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,172 @@
...
+      return; // Already handled
+    } else {
+      // Invalid request - no session ID or not initialization request
+      res.status(400).json({
+        jsonrpc: '2.0',
+        error: {
+          code: -32000,
+          message: 'Bad Request: No valid session ID provided',
+        },
+        id: null,
```
cliffhall, line 63: `id` can't be null, a number could collide with another message with a valid id, but it can be a string. Let's make it `error-32000 `?

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,172 @@
...
+app.delete('/mcp', async (req: Request, res: Response) => {
+  const sessionId = req.headers['mcp-session-id'] as string | undefined;
+  if (!sessionId || !transports[sessionId]) {
+    res.status(400).json({
+      jsonrpc: '2.0',
+      error: {
+        code: -32000,
+        message: 'Bad Request: No valid session ID provided',
+      },
+      id: null,
```
cliffhall, line 123: `id` can't be `null`, a number could collide with another message with a valid id, but it can be a string. Let's make it `error-32000`?

cliffhall (changes requested):
Getting errors because of message ids being null.

shivdeepak:
> > @cliffhall thanks for the review. cleaned it all up, and tested it.
> 
> Not working now.
> 
> ```shell
> Error from MCP server: Error: Error POSTing to endpoint (HTTP 400): {"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"InternalServerError: stream is not readable"},"id":null}
>     at StreamableHTTPClientTransport.send (file:///Users/cliffhall/Projects/mcp-inspector/node_modules/@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js:263:23)
> ```
> 
> Failing because `id` is `null` I believe. Screenshot 2025-04-19 at 5 02 35 PM

Hi @cliffhall 

A few questions:
1. What triggered this error, i.e. how did you test this from MCP Inspector? Did it happen during `DELETE /mcp` call?
2. When I click "Disconnect" on MCP Inspector, it doesn't make an API call to the MCP Inspector backend. Is that an expected behavior? Maybe disconnect should be making a `DELETE /mcp` call?

shivdeepak:
I was able to reproduce the issue. it happens on connect. and it's happening after I removed `req.body` from `transport.handleRequest(req, res, req.body);` in post request handler.

i didn't catch it the last time because i forgot to run `npm run build` on the everything server. mea culpa.

I am currently looking into why req.body is needed.

src/everything/streamableHttp.ts
shivdeepak: Based on the [MCP transport spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports), message id can be `null` if it's an error.

Screenshot 2025-04-19 at 4 17 25 PM

That being said, in other places, I have noticed that the response id matches the request id. I can do that matching here.

What do you think?

src/everything/streamableHttp.ts
shivdeepak: Yes. Because `StreamableHTTPServerTransport` expects it: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts#L138

shivdeepak:
@cliffhall I have addressed the two issues you raised.

1. req.body is passed because [`StreamableHTTPServerTransport` expects it](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts#L138).
2. based on the [MCP transport spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server), response message id can be `null` if it's an error. I am using request.id if it exists.

PR is now ready for review.

shivdeepak:
update: [typescript workflow is now passing on the forked repo with these changes](https://github.com/shivdeepak/servers/actions/runs/14553971316).

cliffhall:
> @cliffhall I have addressed the two issues you raised.
> 
> 1. req.body is passed because [`StreamableHTTPServerTransport` expects it](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts#L138).
> 2. based on the [MCP transport spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server), response message id can be `null` if it's an error. I am using request.id if it exists.

1. It accepts `req.body` as `parsedBody` and passes it on for a POST, but it _isn't required_. 

The problem with doing this is that it always bypasses parsing, which includes a adding a maximum message size and encoding taken from the content type parameters. 

### What happens when we pass in the `parsedBody` param
```typescript
      if (parsedBody !== undefined) {
        rawMessage = parsedBody;
      } else {
        const parsedCt = contentType.parse(ct);
        const body = await getRawBody(req, {
          limit: MAXIMUM_MESSAGE_SIZE,
          encoding: parsedCt.parameters.charset ?? "utf-8",
        });
        rawMessage = JSON.parse(body.toString());
      }
```

IMO, this should not be the default behavior of the client. I know the SDK has a comment that shows how you can pass in a pre-parsed body, but I think there has to be a **specific reason** why we would do that, not just always-on behavior. 

2. More troubleshooting reveals it isn't the returned error messages with null ids that are the problem. It's the response to the `initialize` request. 

The error I see echoed on the inspector's STDOUT is:
```shell
Error from MCP server: Error: Error POSTing to endpoint (HTTP 400): {"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"InternalServerError: stream is not readable"},"id":null}
```

### Well-formed initialize request
Screenshot 2025-04-20 at 4 39 14 PM

### AHA!
In order to continue troubleshooting this, I removed your third param to `transport.handleRequest` so I would get the errors. 

Opening the console in the browser, I see the client isn't hitting the `/mcp` endpoint. It's hitting `/sse` and `/messages` like the old SSE server, even though Streamable HTTP is selected and the url field points to `/mcp`. I tried clearing my local storage, opening in an incognito, etc. Same behavior. 

### When it fails
Screenshot 2025-04-20 at 4 27 11 PM

Putting your third param (parsed body) back in the mix

### It succeeds, but is still talking to an SSE server. 
Screenshot 2025-04-20 at 4 43 42 PM

Can you go back to the Inspector and see what's going on there?

cliffhall:
> Why would this app, which is clearly only using `/mcp` endpoints even be responding to `/sse` and `/messages`?

This is a twist. 

* It's because the inspector's proxy server exposes [`/sse`](https://github.com/modelcontextprotocol/inspector/blob/main/server/src/index.ts#L130) and [`/message`](https://github.com/modelcontextprotocol/inspector/blob/main/server/src/index.ts#L184) endpoints. 

Presumably we need an `/mcp` endpoint on the proxy as well. 

We still don't know why the client is hitting the `/sse` endpoint to begin with, and that would be good to sort out.

shivdeepak:
Yes. The Inspector frontend talks to the Inspector backend over `/sse` and `/messages` endpoints and uses SSE transport. And the backend acts as a proxy for stdio, SSE and Streamable HTTP. This behavior was pre-existing for stdio, and SSE transport modes.

Given how stdio works, it makes sense for it to be implemented that way. But I do agree that for Streamable HTTP, the client side implementation should also use the `/mcp` endpoint.

I will spend some time today get that working. I will also look into why initialize is failing for this PR.

cliffhall:
> I will spend some time today get that working. I will also look into why initialize is failing for this PR.

I fixed the inspector: https://github.com/modelcontextprotocol/inspector/pull/339

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,174 @@
+import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
+import { isInitializeRequest } from '@modelcontextprotocol/sdk/types.js';
+import { InMemoryEventStore } from '@modelcontextprotocol/sdk/examples/shared/inMemoryEventStore.js';
+import express, { Request, Response } from "express";
+import { createServer } from "./everything.js";
+import { randomUUID } from 'node:crypto';
+
+const app = express();
+
+app.use(express.json());
```
cliffhall, line 10: @shivdeepak Remove this. This is what's causing the Parse error problem. 

That moves us on to the next error:
```shell
Error from MCP server: Error: Error POSTing to endpoint (HTTP 400): {"jsonrpc":"2.0","error":{"code":-32000,"message":"Bad Request: No valid session ID provided"}}
```

src/memory/index.ts
```diff
@@ -189,7 +189,7 @@ const knowledgeGraphManager = new KnowledgeGraphManager();
 // The server instance and tools exposed to Claude
 const server = new Server({
   name: "memory-server",
-  version: "1.0.0",
+  version: "0.6.3",
```
cliffhall, line 192: what is this change about?

src/everything/streamableHttp.ts
cliffhall: More testing, and I have an idea that it's the inspector sending things incorrectly. 
* SSE wanted a GET request to `/sse` followed POSTS to `/message`. 
* Streamable wants a POST to `/mcp` followed by more POSTS to `/mcp` with GET used to fetch SSE messages from the server.
* I'm seeing us send to `/mcp` with my changes, but it's doing a GET first, like with SSE.
Screenshot 2025-04-21 at 3 28 35 PM

shivdeepak: @cliffhall made the fixes, and tested them  against your changes from yesterday -- https://github.com/cliffhall/mcp-inspector/tree/3a2e2485273a4dd1bfb5186ceee73fab8f8b8b9a

- removed `express.json()` middleware
- removed `isInitializeRequest(req.body)` check
- removed `req.body` from `transport.handleRequest(req, res, req.body)`

src/everything/streamableHttp.ts
```diff
@@ -0,0 +1,172 @@
...
+        code: -32000,
+        message: 'Bad Request: No valid session ID provided',
+      },
+      id: req?.body?.id,
+    });
+    return;
+  }
+
+  // Check for Last-Event-ID header for resumability
+  const lastEventId = req.headers['last-event-id'] as string | undefined;
```
cliffhall, line 101: I'm not 100% certain how to test this part. I think we need to beef up the logic in the Inspector to make the Reconnect button send the `last-event-id` header. 

But this endpoint is working correctly for initiating and maintaining the SSE stream, and the only thing we're doing here is logging the fact that this proxy saw the header. The SDK deals with recognizing that header and doing a replay from it, so I think this server is just fine.

cliffhall (approved):
All good. Working fine with the [newly improved](https://github.com/modelcontextprotocol/inspector/pull/339) Inspector, and nothing changed that affects previous functionality.

PR #1563, opened 2025-04-24
Make an info message as log instead of error

njbrake:
## Description
Printing of informational message should not be an error message, since nothing is wrong. either a log or a warning message seems more appropriate?
## Server Details

- Server: filesystem
- Changes to: log

## Motivation and Context

My logs were getting littered with this message (I'm running stdio in a docker container, the error message gets propogated to outside the container)

## How Has This Been Tested?
nfc

## Breaking Changes
n/a

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -637,8 +637,8 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
 async function runServer() {
   const transport = new StdioServerTransport();
   await server.connect(transport);
-  console.error("Secure MCP Filesystem Server running on stdio");
-  console.error("Allowed directories:", allowedDirectories);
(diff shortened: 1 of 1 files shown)


olaservo (commented):
Since this is using standard IO (a.k.a. STDIO) as the transport, the server needs to use STDERR for logging. (In other words this is intentional, otherwise clients like Claude Desktop will throw errors.)

njbrake:
> Since this is using standard IO (a.k.a. STDIO) as the transport, the server needs to use STDERR for logging. (In other words this is intentional, otherwise clients like Claude Desktop will throw errors.)

Ah ok thanks for the info! I'll pursue a different workaround then to address my use case 👍🏼

PR #1801, opened 2025-05-15
Fix server-everything output and disconnect behavior

cliffhall:
## Description
* In src/everything/sse.ts and streamableHttp.ts
  - Replace `console.log` with `console.error` throughout
* In src/everything/sse.ts
  - remove the `process.exit()` in the `server.onclose` handler so reconnections are possible
  - add SIGINT handler that does cleanup, server.close, and process.exit (same as in stdio.ts and streamableHttp.ts)

## Server Details

- Server: server-everything
- Changes to: suppress non-json output, handle SIGINT and allow reconnections to sse server

## Motivation and Context

While looking at [this PR for the inspector](https://github.com/modelcontextprotocol/inspector/pull/380) it became obvious that we need to suppress all console.log output as per the spec, even in SSE/StreamableHttp (STDIO was already using only `console.error`). 

Also since that PR was testing ability to reconnect, I fixed the long-annoying behavior of the SSE server of closing entirely when the client disconnects.

## How Has This Been Tested?

Local testing with current Inspector, SSE, StreamableHttp, and STDIO.

## Breaking Changes

Nope.

## Types of changes

- [ ] Documentation update

## Checklist

- [] New and existing tests pass locally

### DIFF
--- a/src/everything/sse.ts
+++ b/src/everything/sse.ts
@@ -19,6 +19,7 @@ app.get("/sse", async (req, res) => {
     await cleanup();
     await server.close();
   };
+
 });
 
 app.post("/message", async (req, res) => {
@@ -27,7 +28,14 @@ app.post("/message", async (req, res) => {
   await transport.handlePostMessage(req, res);
 });
 
+process.on("SIGINT", async () => {
+  await cleanup();
+  await server.close();
+  process.exit(0);
+});
+
 const PORT = process.env.PORT || 3001;
 app.listen(PORT, () => {
   console.error(`Server is running on port ${PORT}`);
 });
+
--- a/src/everything/sse.ts
+++ b/src/everything/sse.ts
@@ -11,24 +11,23 @@ const { server, cleanup } = createServer();
 let transport: SSEServerTransport;
 
 app.get("/sse", async (req, res) => {
-  console.log("Received connection");
+  console.error("Received connection");
   transport = new SSEServerTransport("/message", res);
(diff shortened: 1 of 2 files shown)


src/everything/sse.ts
```diff
@@ -11,24 +11,31 @@ const { server, cleanup } = createServer();
 let transport: SSEServerTransport;
 
 app.get("/sse", async (req, res) => {
-  console.log("Received connection");
+  console.error("Received connection");
```
olaservo, line 14: My understanding was that the spec only specifies the following for STDIO but technically errors could still be logged over SSE or HTTP, so is this actually needed according to the spec, or more of a preference we want to follow?

> 
> In the **stdio** transport:
> ...
> * Messages are delimited by newlines, and **MUST NOT** contain embedded newlines.
> * The server **MAY** write UTF-8 strings to its standard error (`stderr`) for logging
>   purposes. Clients **MAY** capture, forward, or ignore this logging.
> * The server **MUST NOT** write anything to its `stdout` that is not a valid MCP message.
> ...

olaservo: Sorry typed too fast, made a couple edits ^^

cliffhall: I thought it would be reasonable to just be consistent throughout. Otherwise someone not familiar with the spec, who is looking at an SSE server that uses `console.log` might think it was generally ok to do.

olaservo (approved):

PR #1884, opened 2025-05-22
Everything server multiple connection support

cliffhall:
## Description
* In sse.ts
  - remove global server, cleanup, and transport vars
  - add transports map
  - in sse GET handler,
    - check for sessionId, there shouldn't be one, so comment "Reconnecting?" and do nothing if present
    - if sessionId not present 
      - create new server and transport instance
      - connect server to transport
      - add transport to transports map 
      - in server.onclose, delete the transport from the transports map and call cleanup
  - in /message POST handler
    - get the sessionId from the request
    - get the transport from the map by sessionId
    - handle the message if the transport was found
    
* In streamableHttp.ts
  - remove the global server and cleanup vars
  - change transports var to Map
  - in /mcp POST handler
    - when creating a new session
      - create a server instance
      - in server.onclose, delete the transport from the transports map and call cleanup
  - remove the calls to cleanup and server.close in the SIGINT handler, because the transport is closed and its onclose handler closes the server.

## Server Details

- Server: everything
- Changes to: connection behavior

## Motivation and Context

Allow multiple connections to the `everything` server.
 
For both sse and streamableHttp, a server instance needs to be created for each transport. Otherwise, when a new client connects and its new transport is connected to the single server, the previous transport is overwritten in the server instance and can no longer communicate.

## How Has This Been Tested?

Using [this Inspector PR](https://github.com/modelcontextprotocol/inspector/pull/428) which fixes the connection behavior of its proxy server. 

[Here is a video](https://www.loom.com/share/b5a8898fba0c456380aeb121810b52bd) where I demonstrate the problem with the current version of the server, and then with the fixed versions of the SSE and StreamableHttp servers. Manually tested that STDIO works as well.  

## Breaking Changes

Nope.

## Types of changes

- [ ] Documentation update

## Checklist


src/everything/sse.ts
```diff
@@ -6,36 +6,46 @@ console.error('Starting SSE server...');
...
-    await cleanup();
-    await server.close();
-  };
+  let transport: SSEServerTransport;
+  const { server, cleanup } = createServer();
+
+  if (req?.query?.sessionId) {
+    const sessionId = (req?.query?.sessionId as string) || "none";
+    transport = transports.get(sessionId) as SSEServerTransport;
+    console.error("Client Reconnecting? ", transport.sessionId);
```
olaservo, line 18: This log seems ambiguous?  Maybe we should be more explicit in the log about the expected state vs actual state of the existing session?

src/everything/sse.ts
```diff
@@ -6,36 +6,46 @@ console.error('Starting SSE server...');
...
+  } else {
+    // Create and store transport for new session
+    transport = new SSEServerTransport("/message", res);
```
olaservo, line 21: Will the user be able to see and/or manage these multiple sessions in some way in the UI, especially in cases where they are accidental?  My main concern is that someone could be starting up multiple sessions unintentionally.

src/everything/sse.ts
cliffhall: It really shouldn't happen. Once there is a sessionId in SSE, the client should never GET /sse again.

src/everything/sse.ts
cliffhall: The multiple connections are new instances of the client, opened in a new tab or window. Closing the tab closes the connection. Or they could literally click the "Disconnect" button.

cliffhall: Added more verbose message: _"Client Reconnecting? This shouldn't happen; when client has a sessionId, GET /sse should not be called again."_

olaservo: OK thanks - was just thinking about the Inspector issues where users were getting errors because they didn't realize that they had multiple tabs open with Inspector.  I think this is more of a question about the inspector UI anway.

src/everything/sse.ts
```diff
@@ -6,36 +6,46 @@ console.error('Starting SSE server...');
...
-process.on("SIGINT", async () => {
-  await cleanup();
-  await server.close();
-  process.exit(0);
+  const sessionId = (req?.query?.sessionId as string);
+  const transport = transports.get(sessionId);
+  if (transport) {
+    console.error("Client Message from", sessionId);
+    await transport.handlePostMessage(req, res);
+  }
```
olaservo, line 45: Should there be an else condition here with a log if the transport isn't defined?

olaservo (approved):
Had one more logging question/suggestion but otherwise looks good to me.

src/everything/sse.ts
cliffhall: Done

olaservo (approved):

PR #1889, opened 2025-05-22
Add basic mitigation against postgres queries injecting to trigger writes or other dangerous actions

ddworken:
This mitigates the trick of injecting `COMMIT` and `END` to bypass the read-only restrictions. I believe this should fully fix the issue, but I added a warning to encourage people to also enforce this through postgres permissions. 

Note: This is based on the approach used by @smola [here](https://github.com/zed-industries/postgres-context-server/commit/c68598004db5e33c39cc9917aecb07a213d8f27e), so thank you Santiago for the idea of using a parameterized query to accomplish this.

### DIFF
--- a/src/postgres/README.md
+++ b/src/postgres/README.md
@@ -2,6 +2,9 @@
 
 A Model Context Protocol server that provides read-only access to PostgreSQL databases. This server enables LLMs to inspect database schemas and execute read-only queries.
 
+> [!CAUTION]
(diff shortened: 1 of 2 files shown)


ddworken:
@dsp-ant

pcarleton (approved):
LGTM

A test would be awesome to verify, but given there aren't tests already, i'm good w/ merging this as is.

Issue #470, opened 2025-01-04
Filesystem server access denied for allowed paths on Windows


robotdad:
When using a local version of the filesystem server on Windows I am getting access denied for paths that are allowed directories on Windows. I was moving to this to get off of my own fork sine a prior PR was merged but it is not in release. 

**To Reproduce**
1.  Setup some allowed directories on Windows
2.  Try and access file with them using a local build of filesystem server

**Logs**
There are no errors in the logs.

**Additional context**
Here is an example output of read_file showing the issue.
```
{
  `path`: `C:\\source\\servers\\src\\fetch\\README.md`
}
Error: Access denied - path outside allowed directories: C:\source\servers\src\fetch\README.md not in c:\source, C:\Users\mgoodner
```

civicBaby:
Hello, may I ask how the final solution was resolved? Please let me know. Thank you

bartlomiejwolk:
I'm still having the same problem.

chienchien6:
Metoo.I have the same problem

LittleGuyDev:
Bump. I'm having it on Mac using Cursor AI

domdomegg:
There's a PR up in #2111 that's a step towards fixing this. One problem is that Windows actually can have case-sensitive paths, even if it's not the norm. This makes things a bunch harder.

It would be great to get some clever eyes over there, exploring how we can solve this problem in an effective and secure way :)

laozhangzzz:
The repro here is a great minimal example of the root cause: the allowed path and requested path are equivalent on Windows, but the validation compares them with different drive-letter casing / normalization.

The fix direction I collected is:

- canonicalize both configured allowed directories and candidate paths the same way before comparing
- account for Windows drive-letter case-insensitivity
- avoid comparing `C:\source` and `c:\source` as different allowed roots
- use a patched build / release that includes the path validation fix once available

I wrote the full checklist here:
https://laozhangzzz.github.io/agent-pitbook/pits/mcp-filesystem-server-windows-access-denied-case-sensitivity.html

olaservo:
Re-tested this against current `main` (post-#4638/#4631) on Windows 11, driving `validatePath` directly with the same startup resolution `index.ts` uses.

The repro in the report is fixed. `normalizePath` upper-cases the drive letter, and startup keeps both the as-typed spelling and the `fs.realpath` result of each allowed directory, so `c:\source` configured and `C:\source\...` requested now match:

```
[...]

Issue #1987, opened 2025-06-01
Filesystem MPC Server rejects lowercase drive letters in paths


s-varenia:
**Describe the bug**
The filesystem MPC Server incorrectly treats paths with lowercase drive letters as outside the allowed directory scope, even though the path is valid on Windows. This results in an "access denied" error, despite the target path being inside the allowed directory.

**To Reproduce**
1. Use the `create_directory` , `directory_tree`  operation with a path like:

```json
{
  "path": "g:\\GitHub\\obsidian\\Temp"
}
```

2. Observe the error:

```
Error: Access denied - path outside allowed directories:
g:\GitHub\obsidian\Temp not in G:\GitHub\obsidian
```

![Image](https://github.com/user-attachments/assets/e3713bfd-10b7-4330-877c-4e30665b957b)

3. Retry the same operation with the drive letter capitalized:

```json
{
  "path": "G:\\GitHub\\obsidian\\Temp"
}
```

4. This time, the directory is created successfully.

![Image](https://github.com/user-attachments/assets/dea0842a-4779-4ac8-9af8-4e6241e6a037)

**Expected behavior**
Since Windows paths are case-insensitive, both "g:\" and "G:\" should be treated the same. The server should allow the operation if the actual path is within the allowed root, regardless of drive letter case.

**Additional context**
This behavior appears to be caused by a case-sensitive string comparison in the server’s path validation logic. Normalizing the paths or performing case-insensitive comparison (at least for the drive letter) should resolve this inconsistency on Windows systems.

DanOverflow:
I created a temporary fork so I could get the Filesystem server working for myself in Windsurf. This [commit](https://github.com/DanOverflow/mcpservers/commit/15f03cf2e71401e086ed4f331d28bda5acc54738) seemed to get things working properly for me.

I believe this fix would also resolve reported issues: #470, #1820, #1033, #1838

I will submit a Pull Request shortly on this repo.

domdomegg:
Thanks for reporting this! Going to merge this into #470 for clearer tracking.

Issue #1838, opened 2025-05-19
Filesystem MCP Server (Windows/NPX): Flawed path validation ("outside allowed") and case sensitivity with initial arg


Phillipemmons:
**Describe the bug**
The Filesystem MCP Server (@modelcontextprotocol/server-filesystem when run via NPX on Windows, e.g., within Windsurf IDE) exhibits several critical bugs related to path validation and handling:

Illogical "Path Outside Allowed Directories" Error: Operations like read_file and list_directory incorrectly report that a target path is "outside allowed directories" even when the target path is clearly a subdirectory of (or the same as) the allowedDirectory path provided as an argument to the server during startup. For example, if the server is started with C:\Users\MyUser\Documents as an allowed path, an attempt to read C:\Users\MyUser\Documents\subfolder\file.txt will fail with Error: Access denied - path outside allowed directories: C:\Users\MyUser\Documents\subfolder\file.txt not in C:\Users\MyUser\Documents.
Case Sensitivity of Initial Path Argument (Drive Letter on Windows): The server appears to internally canonicalize or prefer an uppercase drive letter for its allowed path argument (e.g., C:\...). If the path argument is provided in the startup configuration (e.g., mcp_config.json in Windsurf) with a lowercase drive letter (e.g., c:\...), the "path outside allowed directories" error for read_file and list_directory becomes persistent. Operations only have a chance of partial success if the configured path argument uses an uppercase drive letter, and even then, AI-driven retries (e.g., by Cascade AI in Windsurf) might be needed if the initial tool call from the AI uses a lowercase drive letter for the operation path.
Inconsistent Behavior Between Tools: The find_by_name tool can successfully locate files within paths that list_directory and read_file report as inaccessible or whose parents "don't exist." This indicates differing path validation logic between tools within the same server.
Misleading "Parent directory does not exist" Error: The list_directory tool sometimes reports "Parent directory does not exist" for a valid path (e.g., configured allowed path X:\path\to\project, trying to list X:\path\to\project results in "Parent directory X:\path\to does not exist"). This might be a knock-on effect of the flawed "path outside allowed directories" validation.

**To Reproduce**
(This assumes an environment like Windsurf IDE that uses mcp_config.json to launch MCP servers via NPX)

Configure the mcp_config.json for the filesystem server as follows (note lowercase 'c'):
JSON

{
  "mcpServers": {
    "filesystem": {
      "command": "npx",
      "args": [
        "-y",
[...]

Issue #1820, opened 2025-05-17
When using the filesystem MCP server on Windows, the error "Parent directory does not exist" always occurs. I have already found a solution to fix it.


lzfxp:
When using the filesystem MCP server on Windows, the error "Parent directory does not exist" always occurs.

My configuration is correct, but this error always occurs. Later, through debugging, I found that the issue lies in this line of code:

const isParentAllowed = allowedDirectories.some(dir => normalizedParent.startsWith(dir));

The normalizedParent parameter converts my drive letter to uppercase, such as “E:\test”, while the directory I configured is “e:\test”. These two strings will never match. I changed the check to the following code, and it can run successfully:

const isParentAllowed = allowedDirectories.some(dir =>
                process.platform === 'win32'
                    ? normalizedParent.toLowerCase().startsWith(dir.toLowerCase())
                    : normalizedParent.startsWith(dir)

Shadowfita:
Same issue here

domdomegg:
Thanks for reporting this! Going to merge this into #470 for clearer tracking.

PR #2111, opened 2025-06-14
Address Fileserver's case insensitivity issue on Windows

DanOverflow:
Changed normalizePath() to convert all normalized path to lowercase on Windows.

## Description
Address Fileserver's case insensitivity issues on Windows
Changed normalizePath() to convert all normalized path to lowercase on Windows.

I believe this should fix issues: #470, #1820, #1033, #1838 and #1987
## Server Details

- Server: filesystem
- Changes to: normalizePath()

## Motivation and Context

The filesystem MPC Server incorrectly treats paths with lowercase drive letters as outside the allowed directory scope, even though the path is valid on Windows. This results in an "access denied" error, despite the target path being inside the allowed directory.

 Behavior defined in #1987 and other related issues. (#470, #1820, #1033, #1838)

## How Has This Been Tested?

My local dev machine running Windsurf on Windows.

## Breaking Changes

N/A

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [ X] I have tested this with an LLM client
- [X ] My code follows the repository's style guidelines

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -24,7 +24,12 @@ if (args.length === 0) {
 
 // Normalize all paths consistently
 function normalizePath(p: string): string {
-  return path.normalize(p);
+  const normalized = path.normalize(p);
+  // On Windows, paths are case-insensitive. Normalize to lowercase for comparisons.
+  if (process.platform === 'win32') {
+    return normalized.toLowerCase();
+  }
+  return normalized;
 }
 
 function expandHome(filepath: string): string {


src/filesystem/index.ts
```diff
@@ -33,7 +33,12 @@ if (args.length === 0) {
 
 // Normalize all paths consistently
 function normalizePath(p: string): string {
-  return path.normalize(p);
+  const normalized = path.normalize(p);
+  // On Windows, paths are case-insensitive. Normalize to lowercase for comparisons.
```
domdomegg, line 37: Errr, according to https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity#change-the-case-sensitivity-of-files-and-directories files technically can be case sensitive on Windows...

Not sure what the best solution is here. Maybe assuming it's not case sensitive is okay, I mean even the Microsoft website seems to have a disclaimer saying most apps assume this - and it would fix the issues that people are reporting.

Are there any libraries for Node.js that handle this more nicely?

domdomegg: Hmm maybe fs.realpath.native (or its sync counterpart)

domdomegg (changes requested):
I worry at the moment these changes open up this risk:
- I set up a case-sensitive directory in Windows, `/path/to`
- I set an allowlist of directories, with something like `/path/to/Hamburger`, so that Claude can view and edit my hamburger recepies
- Some other folder, `/path/to/HamburgER` is in the same folder, which contains my medical records from visiting the emergency room in Hamburg. This data is now accidentally exposed

Appreciate this is fairly unlikely, but anywhere we're accidentally giving more access than we intend feels like a potential security vuln.

To move forwards here, I'd appreciate:
- Some reasoning about why this is okay, actually; or
- An alternative approach that eliminates or sufficiently mitigates this oversharing risk

domdomegg:
Hey, thanks for the contribution! This hasn't been updated for a while, so I'm closing it to help us more easily track active contributions. If you're still interested in working on this, please feel free to reopen it. Thanks again for your contribution!

PR #2116, opened 2025-06-16
Fix spelling: TypeScript and GitHub

DePasqualeOrg:


domdomegg:
Thanks! I think these have now been fixed.

PR #2215, opened 2025-06-26
Example showing how to use elicitation

dend:
Based on work from @connor4312, introduces support for elicitation into the `everything` server with the `#elicitationDemo` tool.


olaservo (changes requested):
Thanks for adding this!  Could you also add a mention of the tool to the Readme?

src/everything/everything.ts
```diff
@@ -459,6 +484,11 @@ export const createServer = () => {
           "Returns a resource reference that can be used by MCP clients",
         inputSchema: zodToJsonSchema(GetResourceReferenceSchema) as ToolInput,
       },
+      {
+        name: ToolName.ELICITATION_DEMO,
+        description: "Demonstrates the elicitation feature",
```
olaservo: Maybe add a little more to the description?

```suggestion
        description: "Demonstrates the Elicitation feature by asking the user to provide information about their favorite color, number, and pets.",
```

dend:
@olaservo done!

src/everything/everything.ts
```diff
@@ -95,6 +103,7 @@ enum ToolName {
   GET_TINY_IMAGE = "getTinyImage",
   ANNOTATED_MESSAGE = "annotatedMessage",
   GET_RESOURCE_REFERENCE = "getResourceReference",
+  ELICITATION_DEMO = "elicitationDemo",
```
olaservo: Looking at Cliff's version, I think I'd also prefer just `ELICITATION` here.

src/everything/everything.ts
```diff
@@ -648,6 +678,39 @@ export const createServer = () => {
       return { content };
     }
 
+    if (name === ToolName.ELICITATION) {
```
olaservo, line 691: Should we also demonstrate the different user actions that are possible?  Something like this:

```
  // Handle different response actions
  const content = [];
  
  if (elicitationResult.action === 'accept' && elicitationResult.content) {
    content.push({
      type: "text",
      text: `✅ User provided their favorite things!`,
    });
    
    // Only access elicitationResult.content when action is accept
    const { color, number, pets } = elicitationResult.content;
    content.push({
      type: "text",
      text: `Their favorites are:\n- Color: ${color || 'not specified'}\n- Number: ${number || 'not specified'}\n- Pets: ${pets || 'not specified'}`,
    });
  } else if (elicitationResult.action === 'decline') {
    content.push({
      type: "text",
      text: `❌ User declined to provide their favorite things.`,
    });
  } else if (elicitationResult.action === 'cancel') {
    content.push({
      type: "text",
      text: `⚠️ User cancelled the elicitation dialog.`,
    });
  }
  
  // Include raw result for debugging
  content.push({
    type: "text",
    text: `\nRaw result: ${JSON.stringify(elicitationResult, null, 2)}`,
  });

  return { content };
  ```

File: src/everything/everything.ts [hunk 4]
```diff
@@ -86,6 +86,14 @@ const GetResourceReferenceSchema = z.object({
     .describe("ID of the resource to reference (1-100)"),
 });
 
+const ElicitationSchema = z.object({
```
olaservo: Looking at Cliff's example again: https://github.com/modelcontextprotocol/servers/pull/2269/files

It could be cleaner and less distracting to use the no args schema approach here instead, since the input params don't actually appear to be used from the ElicitationSchema params (and you could hardcode in the tool result)?

olaservo (approved):
This looks good to me, will let Cliff take a quick look at it too.  There is also a corresponding Inspector PR open here: https://github.com/modelcontextprotocol/inspector/pull/535

cliffhall (approved):
LGTM! 👍

PR #2289, opened 2025-07-06
Fix VideoDB Director icon consistency in README

yoichiojima-2:
## Summary
- Standardize VideoDB Director image tag format to match other entries
- Change height from 14 to 12 pixels for consistency  
- Add missing width attribute and fix attribute order

## Issue
The VideoDB Director icon was displaying inconsistently (oversized) on GitHub mobile app compared to web browser due to non-standard image tag formatting.

## Solution
Updated the image tag to follow the same pattern as all other entries in the third-party servers section.

## Test plan
- [x] Verify the change matches the standard format used by other entries
- [x] Test display on GitHub mobile app to confirm icon size is now consistent


olaservo (approved):
Thanks!

PR #2379, opened 2025-07-18
fix: correct add logic for all vs specific files and add tests (#2373)

### LINKED ISSUE #2373: `git_add` with `files: ["."]` parameter incorrectly tracks .git/ directory files causing repository corruption
**Description:**
The MCP Git server has a critical bug where using the `git_add` tool with the files parameter set to `["."]` incorrectly adds .git/ directory files to the Git index, causing severe repository corruption.

**Expected Behavior:**
When calling `git_add` with `files: ["."]`, it should execute the equivalent of `git add .` which adds all files in the working directory but excludes the .git/ directory (as per Git's built-in behavior).

**Actual Behavior:**
The MCP server appears to be adding all files that start with "." (including .git/ files) instead of properly executing `git add .`. This causes Git to track its own internal files, leading to:
- Repository corruption with hundreds of .git/ files being staged
- Repository becoming unusable
- Need for repository reset/restoration

**Reproduction:**
```json
{
  "tool": "git_add",
  "parameters": {
    "files": ["."],
    "repo_path": "~/projects/any-project"
  }
}

### PR DESCRIPTION
**Pull request by tajbaba999**
## Description
This PR updates the `git_add` function to correctly differentiate between staging all files (`["."]`) and specific files. Previously, using `["."]` could unintentionally include the `.git/` directory, potentially corrupting the repository. This fix ensures only appropriate files are staged.

It also includes two additional test cases:
- One for adding all files using `["."]`
- One for adding specific files only

## Server Details
- Server: github
- Changes to: tools

## Motivation and Context
Fixes #2373

The issue identified that `git_add(files=["."])` was tracking files inside the `.git/` directory. This caused repository integrity issues. This PR corrects the logic to avoid touching internal Git files and ensures safe staging behavior.

## How Has This Been Tested?
- Verified manually that `.git/` is excluded when staging all files
- Added and ran unit tests for both general and specific file staging
- Confirmed behavior works correctly with an LLM client environment

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Let me know if there are any changes or improvements you'd like to see. Happy to revise as needed.

### CODE REVIEW

**Comment by tajbaba999** [comment]
Hello @olaservo,
Please let me know if any changes or improvements are needed. I’ll be happy to make the necessary updates.

File: src/git/src/mcp_server_git/server.py [hunk 1]
```diff
@@ -115,7 +115,10 @@ def git_commit(repo: git.Repo, message: str) -> str:
     return f"Changes committed successfully with hash {commit.hexsha}"
 
 def git_add(repo: git.Repo, files: list[str]) -> str:
-    repo.index.add(files)
```
domdomegg, line 118: forgive my ignorance here, but is there a reason we don't just use `repo.git.add` always? e.g. could we get away with not making `.` a special case?

domdomegg (dismissed):
LGTM - would like to understand why we use repo.index at all (see comment above) before merging.

src/git/src/mcp_server_git/server.py
tajbaba999: > forgive my ignorance here, but is there a reason we don't just use `repo.git.add` always? e.g. could we get away with not making `.` a special case?

The issue was caused by using repo.index.add(["."]), which doesn't behave like git add . and ends up including .git/ files, corrupting the repo. I added a check to use repo.git.add(".") when ["."] is passed, which correctly mimics Git’s behavior. This keeps things safe and avoids staging internal Git files.

tajbaba999:
Hey @domdomegg — fixed a small typo ("successfully") that caused the build to fail. Also shared a quick note above on why this approach was used — could you please re-review the PR?

src/git/src/mcp_server_git/server.py
domdomegg: I meant why not just `repo.git.add` everything (i.e. never use `repo.index.add`)?

Or @claude if you're around are you able to chime in?

domdomegg:
From Claude:

---

Hi @domdomegg,

Regarding your question about why not always use `repo.git.add` instead of `repo.index.add`:

Both approaches work, but there are trade-offs:

1. **Performance**: `repo.index.add(files)` is more direct and efficient when adding specific files because it works with GitPython's index object directly, avoiding the overhead of spawning a git subprocess.

2. **Consistency with GitPython patterns**: Using `repo.index.add()` for specific files follows GitPython's intended API patterns - using the Python API for most operations and falling back to git commands only when needed.

3. **The `." special case**: The issue that prompted this PR is that `repo.index.add(["."])` doesn't behave the same as `git add .`. The git command has special logic to exclude the `.git/` directory, while the index method doesn't have this safeguard.

You could indeed simplify the code to always use `repo.git.add()`, which would work correctly. The current approach is a balance between:
- Using the more efficient Python API for specific file operations
- Using the git command only for the special `." case where the behavior differs

If you prefer consistency over the minor performance benefit, switching everything to `repo.git.add()` would be a valid approach and would eliminate the need for special-case handling.

---

To me, this suggests we could simplify the code by always using repo.git.add - I think I'd have a slight preference for this. I don't think performance is what we're optimising for here.

But that said it sounds like this PR will work and is an improvement, so I will accept it.

Issue #533, opened 2025-01-16
Read Image File for Filesystem server


cliffhall:
**Is your feature request related to a problem? Please describe.**
Filesystem read_file tool currently only reads text files. 

When asking Claude app with the filesystem MCP server enabled, asking it to read and analyze a file returns:

> I can see there's an image file called "Untitled.png" in the directory. However, I need to clarify something: while I can see that the file exists, I cannot actually open or view image files. I can only work with text-based files and perform file system operations.
If you'd like me to analyze the image, you would need to describe what's in it or share it directly in our conversation. Is there something specific about the image you'd like to discuss?

**Describe the solution you'd like**
Add support for reading image files.

The "read_file" logic is in [this switch statement](https://github.com/modelcontextprotocol/servers/blob/main/src/filesystem/index.ts#L448) in the filesystem server.

I haven't drilled into MCP deeply enough to know if another case in this switch and possibly some type declarations would be enough, or if there is some problem with handling an image returned from this tool as opposed to text. Feels like probably the latter, or it would have been relatively low-hanging fruit when the server was originally written.

**Describe alternatives you've considered**
I suppose the "read_file" tool could be modified to return text, images, or any other MIME type). I need to learn more about how the return from this function is consumed to know more.

**Additional context**
This arose from a [discussion on Reddit](https://www.reddit.com/r/ClaudeAI/comments/1i2klyg/anyway_to_get_claude_to_open_images_through/).

lzilioli:
+1

jveko:
+1

cliffhall:
I have [a draft PR for this](https://github.com/modelcontextprotocol/servers/pull/2382) now, but in making it, I uncovered an issue in the SDK. I fixed that with [another PR](https://github.com/modelcontextprotocol/typescript-sdk/pull/786) in the SDK repo, so the first PR is waiting for the second PR to merge and a new SDK release that contains the fix. 


UPDATE: The fix has been merged into the SDK, so when the next release is out, I'll open up the draft PR for review.

PR #2382, opened 2025-07-18
Add media file reading in filesystem server

cliffhall:
## Summary of Changes

This pull request enhances the file system interaction capabilities by clarifying the purpose of the existing text file reading tool through a rename, and by introducing a dedicated tool for handling binary media files. This allows the system to process and return image and audio data in a base64 format, significantly expanding the types of files that can be directly accessed and utilized.

### Highlights

* **Tool Renaming**: The existing `read_file` tool has been renamed to `read_text_file` to more accurately reflect its function of reading text-based file content.
* **New Media Handling Tool**: A new tool, `read_media_file`, has been introduced. This tool is designed to read image and audio files and return their content as base64 encoded data, along with the appropriate MIME type.
* **Dependency Update**: The `@modelcontextprotocol/sdk` dependency has been updated to version `1.16.0`, which also brings in a new transitive dependency, `eventsource-parser`.
* **Documentation Updates**: The `filesystem` README has been updated to document the renamed `read_text_file` tool and provide details for the newly added `read_media_file` tool.

<details>
<summary><b>Changelog</b></summary>

* **package-lock.json**
    * Updated `@modelcontextprotocol/sdk` from `1.12.3` to `1.16.0`.
    * Added `eventsource-parser` as a new transitive dependency.
* **src/filesystem/README.md**
    * Renamed documentation for `read_file` to `read_text_file`.
    * Updated description for `read_text_file` to specify text-only reading and mention `head`/`tail` parameters.
    * Added new documentation for the `read_media_file` tool, detailing its purpose and input parameters.
* **src/filesystem/index.ts**
    * Imported `createReadStream` for streaming file content (line 13).
    * Renamed `ReadFileArgsSchema` to `ReadTextFileArgsSchema` (line 120).
    * Added `ReadMediaFileArgsSchema` for the new media tool (line 126).
    * Introduced `readFileAsBase64Stream` utility function to efficiently read and base64 encode file content (lines 482-495).
    * Updated the `ListToolsRequestSchema` handler to rename `read_file` to `read_text_file` and include the new `read_media_file` tool definition (lines 502-519).
    * Modified the `CallToolRequestSchema` handler to process `read_text_file` calls under its new name and added a new case for `read_media_file` to handle media file reading, MIME type detection, and base64 encoding (lines 631-694).
* **src/filesystem/package.json**
    * Updated `@modelcontextprotocol/sdk` from `1.12.3` to `1.16.0`.
</details>

## Server Details

- Server: filesystem
- Changes to: tools

## Motivation and Context

Currently, the `server-filesystem` MCP server has a `read_file` tool, which has optional `head` and `tail` params and always treats the file as utf-8 text. 

Issue https://github.com/modelcontextprotocol/servers/issues/533 contains a user report:

> When asking Claude app with the filesystem MCP server enabled, asking it to read and analyze a file returns:
> 
> > I can see there's an image file called "Untitled.png" in the directory. However, I need to clarify something: while I can see that the file exists, I cannot actually open or view image files. I can only work with text-based files and perform file system operations. If you'd like me to analyze the image, you would need to describe what's in it or share it directly in our conversation. Is there something specific about the image you'd like to discuss?

This PR adds support for reading media files (audio and image) with a new `read_media_file` tool, and the old `read_file` tool is renamed to `read_text_file` for clarity.

### Using the Inspector UI:
small-ss

### Using the Inspector CLI
Screenshot 2025-07-18 at 4 30 30 PM

## Breaking Changes

The tool name `read_file` is now `read_text_file` and its description is updated to reflect that it only works with text files. This should not present an issue to clients, since it will make the decision of when to use the tool clearer.

## Types of changes

- [ ] Documentation update

## Checklist

### Error when reading large files
* Error is `Maximum call stack size exceeded`
* Currently, this tool reveals an issue that exists, I believe, in the SDK, when the file is larger than a certain size. 
* When attempting to read an example file called `des.png` that is 4mb in size:

### Calling the tool for a large file with Inspector UI
Screenshot 2025-07-18 at 4 53 18 PM

### Calling the tool for a large file with Inspector CLI
Screenshot 2025-07-18 at 4 51 40 PM

#### It isn't the function
* Extracting the function that reads the file into a simple test script shows that the function has no issue reading and returning the large file, so the error is not coming from that function. 

Screenshot 2025-07-18 at 4 57 01 PM

#### It isn't the server
* The server tool that invokes it does nothing but return the result, so it cannot be the server.

Screenshot 2025-07-18 at 5 00 26 PM

### It isn't the Inspector
* Both the Inspector UI and CLI return the same error, but are using completely different code to invoke the tool. _Neither are doing any recursive post-processing with the result that would lead to a stack overflow._

##### Inspector UI
Screenshot 2025-07-18 at 5 13 55 PM

##### Inspector CLI
Screenshot 2025-07-18 at 5 05 09 PM

###  Found culprit in SDK

* [Line 607 ](https://github.com/cliffhall/mcp-typescript-sdk/blob/main/src/shared/protocol.ts#L607) of `Protocol.ts` is executed when the response is received. It parses the `resultSchema`.
Screenshot 2025-07-18 at 5 29 10 PM

* By replacing 
  * `result = resultSchema.parse(response.result);` with
  *  `const result = response.result;`

The error no longer occurs with the large file:
Screenshot 2025-07-18 at 6 01 39 PM

* I explored lots of ways to make the compiled schema less massive using `z.lazy()` and `z.discriminatedUnion()` but in the end that didn't fix it. 

* The actual culprit turned out to be Zod's .base64() validation.

* Under the hood, z.string().base64() uses a regular expression to validate the string. While this regex is fine for typical inputs, running it against a string that is several megabytes long can cause the JavaScript engine's regex parser to hit its internal recursion limit, resulting in a "Maximum call stack size exceeded" error.

## Problem solved
* What was needed was a more robust base64 checker in the SDK. I added one and it fixed the problem. I've created [a PR that fixes the problem](https://github.com/modelcontextprotocol/typescript-sdk/pull/786).

* In the Inspector CLI and UI, tested the large file that was causing the "Maximum call stack size exceeded" error. It is no longer present and the 4mb file can be processed.

### Inspector CLI
Screenshot 2025-07-18 at 8 28 42 PM

### Inspector UI
Screenshot 2025-07-18 at 8 27 55 PM

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -499,6 +499,11 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
   return {
     tools: [
       {
+        name: "read_file",
+        description: "Read the complete contents of a file as text. DEPRECATED: Use read_text_file instead.",
+        inputSchema: zodToJsonSchema(ReadTextFileArgsSchema) as ToolInput,
+      },
+      {
         name: "read_text_file",
         description:
           "Read the complete contents of a file from the file system as text. " +
@@ -628,6 +633,7 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
     const { name, arguments: args } = request.params;
 
     switch (name) {
+      case "read_file":
       case "read_text_file": {
         const parsed = ReadTextFileArgsSchema.safeParse(args);
         if (!parsed.success) {


src/filesystem/index.ts
```diff
@@ -597,39 +628,71 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
     const { name, arguments: args } = request.params;
 
     switch (name) {
-      case "read_file": {
-        const parsed = ReadFileArgsSchema.safeParse(args);
+      case "read_text_file": {
```
olaservo, line 637: Could we do something like this for now and have `read_file` fall back to `read_text_file`?  Even though the LLM should be able to see the new tool names and figure things out, this is one of the most used servers so I'm concerned that a lot of things like tutorials or even agent workflows might reference the old tool name.

```
// In ListToolsRequestSchema handler:
{
  name: "read_file",
  description: "Read the complete contents of a file as text. DEPRECATED: Use read_text_file instead.",
  inputSchema: zodToJsonSchema(ReadTextFileArgsSchema) as ToolInput,
},
{
  name: "read_text_file", 
  description: "Read the complete contents of a file as text...",
  inputSchema: zodToJsonSchema(ReadTextFileArgsSchema) as ToolInput,
},

// In CallToolRequestSchema handler:
case "read_file":
case "read_text_file": {
  // Same implementation for both
  const parsed = ReadTextFileArgsSchema.safeParse(args);
  // ... rest of implementation
}
```

cliffhall: Done.

cliffhall:
@olaservo Good call. Replaced the `read_file` tool with a deprecation notice in description, which is functionally just an alias for `read_text_file`.

Screenshot 2025-07-25 at 3 58 21 PM

Seedmanc:
I don't understand how this is different from regular read file, it produces the same base64 instead of an image, so when a vision LLM reads it is sees a long string of symbols instead of an image. Using Msty with Gemma 4 and Qwen 3.6, they all say they see base64.

PR #2550, opened 2025-08-15
Add Claude Code GitHub workflow

domdomegg:
Adds the Claude Code GitHub Actions workflow to enable Claude assistance on issues and pull requests when mentioned with @claude.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -1,0 +1,51 @@
+name: Claude Code
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  issues:
+    types: [opened, assigned]
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  claude:
+    if: |
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
+      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+      issues: read
+      id-token: write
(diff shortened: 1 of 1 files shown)


.github/workflows/claude.yml
```diff
@@ -0,0 +1,51 @@
...
+            actions: read
+
+          # Trigger when assigned to an issue
+          assignee_trigger: "claude"
+          
+          # Allow Claude to run bash
+          # This should be safe given the repo is already public
+          allowed_tools: "Bash"
+          
+          custom_instructions: |
```
olaservo, line 50: I think it makes sense to add this as-is so we can do a general pass at these issues and PRs, but ultimately it seems like we need to give Claude a reference to MCP docs and/or specs to augment instructions and existing MCP knowledge. 

For example, locally I am using in this server I created for referencing official docs: https://github.com/olaservo/mcp-advisor

Perhaps we should formalize a simple docs/specs server for this and other use cases?

.github/workflows/claude.yml
```diff
@@ -0,0 +1,51 @@
...
+  pull_request_review_comment:
+    types: [created]
+  issues:
+    types: [opened, assigned]
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  claude:
+    if: |
```
olaservo, line 15: Can anyone trigger this action, or just people with certain permissions in the repo?

domdomegg: Anyone can trigger this action. I think this should be mostly fine and prefer this so that everyone can benefit. We can restrict it if people really start spamming it.

.github/workflows/claude.yml
domdomegg: Yeah, having a good docs server would be good! Although I was planning to get something basic up and running with this PR, and we can add extra tools / context to Claude as we discover where it falls short.

Also to be clear: this workflow only runs when someone explicitly tags Claude with `@claude`, not on all issues and PRs by default.

olaservo (approved):

.github/workflows/claude.yml
olaservo: OK!  I approved but not sure if you meant you were adding something else here before making this active, so I held off on merging just in case.

PR #2567, opened 2025-08-17
fix: Change memory server default filename from memory.json to memory.jsonl

### LINKED ISSUE #2361: Memory MCP uses JSONL for 'memory.json' as each line is a json format but not the entire file
**Describe the bug**
@modelcontextprotocol/server-memory has a configuration `MEMORY_FILE_PATH` which by default takes value `memory.json` but the content of the file is separate lines of with JSON content each. The entire file is not a JSON though. 
Please change the file name to `memory.jsonl`, that will make IDEs happy and will not show lint errors that lines are not separated by comma and the there is no topmost object or array element

**To Reproduce**
1. Install memory mcp
2. add 2 records
3. open the memory.json file
4. The content of the file 2 lines each containing valid json. The file itself is not of the json format

**Expected behavior**
Either file content is a valid JSON file with wrapper object or array having child items separated by comma, or file name has `.jsonl` extension to show that the format is expected an IDEs can treat it as valid file

**Logs**

**Additional context**

olaservo:
The Memory MCP server uses JSONL format (JSON Lines) where each line contains a separate JSON object, but was using a .json file extension. This caused IDE lint errors and confusion since the file is not valid JSON.

**Changes:**
- Update default filename in index.ts from memory.json to memory.jsonl
- Update documentation references in README.md
- Maintain backward compatibility for existing MEMORY_FILE_PATH configs

Fixes #2361


src/memory/index.ts
```diff
@@ -11,7 +11,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 
 // Define memory file path using environment variable with fallback
-const defaultMemoryPath = path.join(path.dirname(fileURLToPath(import.meta.url)), 'memory.json');
+const defaultMemoryPath = path.join(path.dirname(fileURLToPath(import.meta.url)), 'memory.jsonl');
```
domdomegg, line 14: Might this cause some weird backwards compatibility problems, e.g. if people have `npx -y @modelcontextprotocol/server-memory` in config, suddenly it'll seem to them like all their memories have disappeared.

Maybe we could also check before this: if no custom path is set AND memory.json exists AND memory.jsonl does not exist, move memory.json to memory.jsonl.

domdomegg (dismissed):
Good, although maybe want to address the backwards compatibility comment above

olaservo:
@claude could you take a look at the latest comments and incorporate these suggestions?

src/memory/index.ts
```diff
@@ -11,14 +11,42 @@ import path from 'path';
...
+  // No custom path set, check for backward compatibility migration
+  const oldMemoryPath = path.join(path.dirname(fileURLToPath(import.meta.url)), 'memory.json');
+  const newMemoryPath = defaultMemoryPath;
+  
+  try {
+    // Check if old file exists and new file doesn't
+    await fs.access(oldMemoryPath);
+    try {
+      await fs.access(newMemoryPath);
+      // Both files exist, use new one (no migration needed)
```
olaservo, line 34: @claude lets log a console.error message with something to clearly indicate this was detected.

src/memory/index.ts
olaservo: I made this change.

PR #2570, opened 2025-08-17
fix: align everything server resource URI numbering with array indices

### LINKED ISSUE #475: [everything] Unintuitive / misleading resource indexing
**Describe the bug**
The everything server's resource indexing is counterintuitive. While the array index starts at 0, the resource numbering in URIs starts at 1, which leads to confusing behavior where:
- Resource at index 0 is accessed via URI "test://static/resource/1"
- Resource at index 1 is accessed via URI "test://static/resource/2"
etc.

From the server code:
```typescript
const ALL_RESOURCES: Resource[] = Array.from({ length: 100 }, (_, i) => ({
    const url = `test://static/resource/${i + 1}`;
    if (i % 2 === 0) {
        // index 0 (first resource) is text
        // index 2 (third resource) is text
        // etc.
```

This means even-indexed resources (0,2,4...) are text, but they map to odd-numbered URIs (1,3,5...).

**To Reproduce**
Make a ListResources request, then try to read the first text resource:

1. List resources:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "resources/list"
}
```

2. Get first text resource (index 0):
```json
{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "resources/read",
    "params": {
        "uri": "test://static/resource/1"
    }
}
```

**Expected behavior**
Either:
1. Align the indexing so resource numbers in URIs match array indices (preferred)
2. Or document this behavior clearly to prevent confusion

olaservo:
Fixes issue #475 where array indices (0-based) didn't match URI numbers (1-based), causing confusion.

## Changes
- Resource URIs now use 0-based numbering to match array indices
- Resource 0 is now accessed via `test://static/resource/0`
- Even indices now map to even URI numbers (more intuitive)

Resolves #475

Note: this is technically a breaking change, but since this is a demo server for testing, it seems like this shouldn't actually be an issue?

### DIFF
--- a/src/everything/everything.ts
+++ b/src/everything/everything.ts
@@ -254,7 +254,7 @@ export const createServer = () => {
   };
 
   const ALL_RESOURCES: Resource[] = Array.from({ length: 100 }, (_, i) => {
-    const uri = `test://static/resource/${i + 1}`;
+    const uri = `test://static/resource/${i}`;
     if (i % 2 === 0) {
       return {
         uri,
@@ -316,7 +316,7 @@ export const createServer = () => {
     const uri = request.params.uri;
 
     if (uri.startsWith("test://static/resource/")) {
-      const index = parseInt(uri.split("/").pop() ?? "", 10) - 1;
+      const index = parseInt(uri.split("/").pop() ?? "", 10);
       if (index >= 0 && index < ALL_RESOURCES.length) {
         const resource = ALL_RESOURCES[index];
         return {


cliffhall:
> LGTM.
> 
> I am a bit confused why we used 1-indexing in the first place, so am a bit wary (something something Chesterton's Fence). But seems fairly low risk.

@domdomegg I'm guessing it was so that in the Inspector when you look at the resource list it is human-aligned, i.e., only programmers would not scratch their head at "Resource 0".

Guess we'll never really know, since this was part of the Initial Commit for the repo, and there is no associated detail in the commit message.

Screenshot 2025-08-18 at 2 33 04 PM

domdomegg:
Sorry to maybe cause more chaos by introducing a new idea (which might be wrong because I don't have a huge amount of context on this server), but maybe we avoid all this weirdness by just using intentionally random ids?

E.g. just generate a uuid for everything, and either:
- drop the note about odd/evens being particular types of resources
- put things under proper paths if we want the paths to convey this info e.g. `test://static/resource/text/{uuid}` and `test://static/resource/blob/{uuid}`

cliffhall:
> Sorry to maybe cause more chaos by introducing a new idea (which might be wrong because I don't have a huge amount of context on this server), but maybe we avoid all this weirdness by just using intentionally random ids?
> 
> E.g. just generate a uuid for everything, and either:
> 
> * drop the note about odd/evens being particular types of resources
> * put things under proper paths if we want the paths to convey this info e.g. `test://static/resource/text/{uuid}` and `test://static/resource/blob/{uuid}`

@domdomegg excellent suggestion. Removes any assumptions about array index being meaningful, axes the admittedly strange "even/odd" language from the description, and delivers the point that URIs can carry meaning.

domdomegg:
@claude can you give a shot at implementing the proposal from @domdomegg above? create it as a separate branch, if you can please.

PR #2596, opened 2025-08-22
UUID based resources

### LINKED ISSUE #475: [everything] Unintuitive / misleading resource indexing

domdomegg:
Fixes #475

Also see #2570

### DIFF
--- a/src/everything/README.md
+++ b/src/everything/README.md
@@ -66,7 +66,7 @@ This MCP server attempts to exercise all the features of the MCP protocol. It is
 8. `getResourceReference`
    - Returns a resource reference that can be used by MCP clients
    - Inputs:
-     - `resourceId` (number, 1-100): ID of the resource to reference
+     - `resourceUri` (string): URI of the resource to reference (e.g., test://static/resource/text/{uuid})
    - Returns: A resource reference with:
      - Text introduction
      - Embedded resource with `type: "resource"`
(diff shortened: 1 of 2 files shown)


domdomegg:
@claude please summarise the changes made in this PR. Read the discussion in #2570 for more context too.

olaservo:
@claude could you update this PR to interleave the resource types as suggested in the previous comment?

olaservo:
@cliffhall is this more what you were thinking?
image

(This is a screenshot from pulling the latest from this PR and then building it locally, and testing in latest Inspector.)

cliffhall:
> @cliffhall is this more what you were thinking?

@olaservo Yeah, but the new naming makes it fugly. It really would be better if the resource names were what they once were (Resource 1 ... Resource 100). Only when you selected a resource or listed it with the getResources or getResourceLinks tools did you really see the further detail about types.

cliffhall:
We refactored the everything server entirely, and the way it was refactored sort of makes this PR moot.

PR #2609, opened 2025-08-24
fix: resolve relative paths against allowed directories instead of process.cwd()

### LINKED ISSUE #2526: filesystem: Incorrect Path resolution causes "Error: Access denied - path outside allowed directories"
**Describe the bug**
Since my last update of the Fileserver MCP I see basically every first call to the MCP fail with "Error: Access denied - path outside allowed directories". Which is really annoying because it means many tool calls instead of the first one succeeding.

**To Reproduce**
1. configure the fileserver MCP with one path that doesn't include the execution dir of the server (or c:\windows\system32 apparently)
2. Have Claude try to open a relative path directly in his root path, e.g.
{
  `path`: `make-automation-audit-20250807/message.log`
}
3. see Error instead of success
Error: Access denied - path outside allowed directories: C:\Windows\system32\make-automation-audit-20250807\message.log not in C:\Users\Stefan Polenz\Documents\claude

**Expected behavior**
Until recently giving the current project dir to Claude would allow it to open it right away. 

**Logs**
from mcp-server-Filesystem.log
2025-08-11T17:57:20.223Z [Filesystem] [info] Message from server: {"result":{"content":[{"type":"text","text":"Error: Access denied - path outside allowed directories: C:\\Windows\\system32\\make-automation-audit-20250807\\message.log not in C:\\Users\\Stefan Polenz\\Documents\\claude"}],"isError":true},"jsonrpc":"2.0","id":15} { metadata: undefined }

**Additional context**
Working with Claude on the issue we are fairly confident that the issue is that in 
servers\src\filesystem\index.ts
in lines 77+, so https://github.com/modelcontextprotocol/servers/blob/aad86eecadb719e3e1b3419f0cbdfa72072c6fb6/src/filesystem/index.ts#L77
sync function validatePath(requestedPath: string): Promise<string> {
    ? path.resolve(expandedPath)
    : path.resolve(process.cwd(), expandedPath);
causes issues as the current working dir at the time of calling the functions isn't in any of the permitted paths.

A possible solution (that I can't test as I can't seem to see building information):
sync function validatePath(requestedPath: string): Promise<string> {
    ? path.resolve(expandedPath)
    : (() => {
        // Try to resolve relative path against all allowed directories
        for (const allowedDir of allowedDirectories) {
            const candidate = path.resolve(allowedDir, expandedPath);
            // Check if the resulting path lies within an allowed directory
            if (allowedDirectories.some(dir => candidate.startsWith(dir))) {
                return candidate;
            }
        }
        // Fallback: use first allowed directory as base
        return path.resolve(allowedDirectories[0], expandedPath);
    })();

I could try to push this change - but seeing as I can't test currently I think this bug ticket might be a better option.

olaservo:
Fixes issue where relative paths were incorrectly resolved against process.cwd(), causing "Access denied - path outside allowed directories" errors when the MCP server's working directory was outside the configured allowed directories.

The fix implements intelligent path resolution that:
1. First tries to resolve relative paths against each allowed directory
2. Validates the resulting path is within allowed directories 
3. Falls back to the first allowed directory if no valid resolution is found
4. Maintains backward compatibility by falling back to process.cwd() when no allowed directories are configured

Resolves #2526

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -204,6 +204,30 @@ describe('Lib Functions', () => {
         await expect(validatePath(newFilePath))
           .rejects.toThrow('Parent directory does not exist');
       });
+
+      it('resolves relative paths against allowed directories instead of process.cwd()', async () => {
+        const relativePath = 'test-file.txt';
+        const originalCwd = process.cwd;
+        
+        // Mock process.cwd to return a directory outside allowed directories
+        const disallowedCwd = process.platform === 'win32' ? 'C:\\Windows\\System32' : '/root';
+        (process as any).cwd = jest.fn(() => disallowedCwd);
+        
+        try {
+          const result = await validatePath(relativePath);
+          
+          // Result should be resolved against first allowed directory, not process.cwd()
+          const expectedPath = process.platform === 'win32' 
+            ? path.resolve('C:\\Users\\test', relativePath)
+            : path.resolve('/home/user', relativePath);
+          
+          expect(result).toBe(expectedPath);
+          expect(result).not.toContain(disallowedCwd);
+        } finally {
+          // Restore original process.cwd
+          process.cwd = originalCwd;
+        }
+      });
     });
   });
 
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -72,12 +72,35 @@ export function createUnifiedDiff(originalContent: string, newContent: string, f
   );
 }
 
+// Helper function to resolve relative paths against allowed directories
+function resolveRelativePathAgainstAllowedDirectories(relativePath: string): string {
+  if (allowedDirectories.length === 0) {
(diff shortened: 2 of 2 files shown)


olaservo:
@claude do we really need to handle when there are no allowed directories set?

olaservo:
This has been open for a while and it looks like this is still an issue, so I updated it to use vitest (and also validated that the test fails without the fix).

cliffhall:
@olaservo Looks good. While you're at it, would you mind running `npm audit fix` on the `filesystem` project? 

Screenshot 2026-02-07 at 4 55 04 PM

olaservo:
@cliffhall Thanks, done! Ran `npm audit fix` on the filesystem project. It resolved the high-severity `qs` vulnerability. The remaining 6 moderate-severity issues are all in dev dependencies (esbuild/vite/vitest chain) and would require a breaking vitest v2→v4 upgrade (`npm audit fix --force`), so I left those as-is.

PR #2630, opened 2025-08-27
feat(filesystem): ignore dot directories by default to reduce token usage

### LINKED ISSUE #2219: filesystem mcp: dot directories should be ignored
**Describe the bug**
Dot directories (eg `.git`, `.terraform`, etc) are included in the `filesystem` MCP search tools.

These directories, particularly `.git` can get enormous and thus massively inflate the token usage for the agent.

What's more, these directories generally don't add much value (they're "hidden" for a reason) for project analysis. And in the worst case scenarios, might even contain sensitive information.

Personally, I'd go further and say allowing `filesystem` to work against any dot-prefixed filesystem objects is a _security_ risk, however, for the sake of wanting to keep this discussion focused on reducing token count rather than whether users should or should not be aware of what data they're sharing with the agent, this GH Issue is focused on dot-prefixed directories like `.git`.

**To Reproduce**
Run the MCP with the authorised path being the root of any project that is also a git repository.

**Expected behavior**
Dot-prefixed directories should be ignored _by default_. An environmental variable and/or command line flag can be used to optionally re-enable this behaviour.

**Logs**

**Additional context**
While the intended behaviour is to selectively include the files and/or directories you wish to authorise for use with this MCP, the benefit of using an MCP like this is for speed and convenience. Thus a typical use case might well be someone selecting the entire project directory. On the surface, this is a pragmatic approach -- so long as dot-prefixed directories (and arguably files too) are ignored by default.

olaservo:
This PR addresses issue #2219 by implementing default filtering of dot-prefixed files and directories in the filesystem MCP server.

## Changes
- Add environment variable `MCP_FILESYSTEM_INCLUDE_HIDDEN` (default: false)
- Filter dot-prefixed files/directories in `list_directory`, `directory_tree`, `search_files`
- Reduce token usage from large directories like `.git`, `.terraform`, etc.
- Enhance security by avoiding exposure of potentially sensitive hidden files
- Include comprehensive tests for the new filtering functionality
- Update documentation to explain the new behavior and environment variable

## Backward Compatibility
Existing behavior is maintained by setting `MCP_FILESYSTEM_INCLUDE_HIDDEN=true`

Fixes #2219

### DIFF
--- a/src/filesystem/README.md
+++ b/src/filesystem/README.md
@@ -15,6 +15,20 @@ Node.js server implementing Model Context Protocol (MCP) for filesystem operatio
 
 The server uses a flexible directory access control system. Directories can be specified via command-line arguments or dynamically via [Roots](https://modelcontextprotocol.io/docs/learn/client-concepts#roots).
 
+### Hidden Files and Directories
+
+By default, the filesystem server **ignores dot-prefixed files and directories** (like `.git`, `.env`, `.terraform`, etc.) to:
+- Reduce token usage (especially from large directories like `.git`)
+- Enhance security by avoiding exposure of potentially sensitive hidden files
+- Follow the convention that dot-prefixed items are typically "hidden" for good reason
+
+To include hidden files and directories, set the environment variable:
+```bash
+export MCP_FILESYSTEM_INCLUDE_HIDDEN=true
+```
+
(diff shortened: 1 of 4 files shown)


domdomegg (commented):
As per https://github.com/modelcontextprotocol/servers/issues/2219#issuecomment-3235050347

PR #2631, opened 2025-08-27
fix: handle null values for optional fetch parameters

### LINKED ISSUE #2035: fetch fails without optional settings
**Describe the bug**
The docs clearly say:

fetch - Fetches a URL from the internet and extracts its contents as markdown.
url (string, required): URL to fetch
max_length (integer, **optional**): Maximum number of characters to return (default: 5000)
start_index (integer, **optional**): Start content from this character index (default: 0)
raw (boolean, **optional**): Get raw content without markdown conversion (default: false)

**To Reproduce**
1. When I run fetch in LibreChat via a Docker container (for the mcp), I receive the attached errors

**Expected behavior**
It should disregard optional settings

Image

olaservo:
Fixes #2035 - fetch server now properly handles null values for optional parameters.

Makes max_length, start_index, and raw parameters properly optional by using Optional[] types and handling null values gracefully.

This fixes the issue where LibreChat and other MCP clients pass explicit null values for optional parameters, which was causing Pydantic validation failures.

### DIFF
--- a/src/fetch/pyproject.toml
+++ b/src/fetch/pyproject.toml
@@ -33,4 +33,4 @@ requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [tool.uv]
-dev-dependencies = ["pyright>=1.1.389", "ruff>=0.7.3"]
+dev-dependencies = ["pyright>=1.1.389", "ruff>=0.7.3", "pytest>=7.0.0"]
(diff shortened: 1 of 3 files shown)


olaservo:
@claude can you add tests to cover the scenarios from the original issue?

PR #2672, opened 2025-09-03
direct long running tool progress notifications to correct stream

evalstate:
Specifies related Request ID to allow routing to POST response SSE stream for long running tool notifications.

## Description

Fixes issue https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1394

## Server Details

Everything Server reference implementation.

## Motivation and Context

As per issue 1394.

## How Has This Been Tested?

Tested locally.

## Breaking Changes

None.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [X] I have tested this with an LLM client
- [X] I have documented all environment variables and configuration options

### DIFF
--- a/src/everything/everything.ts
+++ b/src/everything/everything.ts
@@ -542,7 +542,7 @@ export const createServer = () => {
     return { tools };
   });
 
-  server.setRequestHandler(CallToolRequestSchema, async (request) => {
+  server.setRequestHandler(CallToolRequestSchema, async (request,extra) => {
     const { name, arguments: args } = request.params;
 
     if (name === ToolName.ECHO) {
@@ -584,7 +584,7 @@ export const createServer = () => {
               total: steps,
               progressToken,
             },
-          });
+          },{relatedRequestId: extra.requestId});
         }
       }
 


cliffhall (approved):
Approving because I agree that this change is correct and necessary. It ensures that the protocol [will not debounce the messages](https://github.com/modelcontextprotocol/typescript-sdk/blob/bf817939917277a4c59f2e19e7b44b8dd7ff140c/README.md?plain=1#L915), and that's always what we want for progress notifications. 

However, I was not able to verify that [the issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1394) that this PR was raised in response to 
a) actually exists 
b) was solved by this

**a)** That issue claims that the progress notifications are sent from the server to the client on the GET request. That is actually [what the specification calls for](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server) on the StreamableHttp transport.

**b)** In my tests using the Inspector to connect to the everything server and run the `longRunningOperation` tool both with and without this change, I see all activity happening on the POST, not on the GET as the spec would indicate it should be doing. I believe this is the result of the proxy being between the client and server, muddying the waters and that the server is likely sending these progress notifications on the GET stream as expected.

cliffhall:
> However, I was not able to verify that [the issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1394) that this PR was raised in response to a) actually exists b) was solved by this


Ok, I now have more information and see that

a) it is a problem (any SSE message associated with a live request should go out on the stream dedicated to that request, not the standalone stream for notifications such as list or resource changes.  
image

b)  I verified in the Typescript SDK that the [send method](https://github.com/cliffhall/mcp-typescript-sdk/blob/main/src/server/streamableHttp.ts#L665) of `‎StreamableHTTPServerTransport` will in fact check for `relatedRequestId` and direct messages to the appropriate stream (standalone if it exists and there is no `relatedRequestId`, or the dedicated stream for the request if `relatedRequestId` is present).

Therefore, what I was seeing when I tested with the inspector ([videos posted to the issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1394) that this PR was raised in response to) was in fact the Inspector proxy muddying the waters. There, I observed that both before and after this change, the messages appeared on the POST opened by the client to the proxy. The hidden behavior was the stream the server was sending to the proxy on.

@evalstate has confirmed that his locally instrumented version of the Python SDK showed that the notification was previously going out on the standalone (GET) stream, and this change causes it to go out on the POST stream for the request. He will post before/after screenshots to this PR as soon as he's back at his machine.

evalstate:
Hi Cliff, see video. There are some GET calls interspersed due to the random log messages. 

https://github.com/user-attachments/assets/e23be370-1430-4dab-a8ee-22c5f93c1d6e

PR #2730, opened 2025-09-17
Fix insufficient create-release condition in GitHub Action

cliffhall:
## Description

* In `.github/workflows/release.yml`
  - in `create-release` job condition
    - use `always()` to force evaluation, even if one or more dependencies failed or was skipped
    - check for all combinations of publish results that should lead to a release creation
      - pypi = success / npm = skipped
      - pypi = skipped / npm = success 
      - pypi = success / npm = success

## Motivation and Context

We still have a problem with Automatic Release Creation where if there are only changes to be published to npm OR pypi BUT NOT both, the final `create-release` job will not be run. See [this recent Action](https://github.com/modelcontextprotocol/servers/actions/runs/17671290638) where TypeScript changes were published to npm, but the `create-release` job was not run.

### The problem
The original condition `needs.update-packages.outputs.changes_made == 'true'` in the `create-release` job was insufficient because it only checked whether changes were made to packages, but completely ignored the success or failure status of the dependent publishing jobs (`publish-pypi` and `publish-npm`). Also, since `always()` was not part of the condition, it was actually never evaluated in a mixed success case, because the `needs` field dependencies must all succeed before the condition is evaluated.

### The fix
* Use `always()` to ensure the condition is evaluated even if some dependent jobs fail or are skipped. Without `always()`, if either `publish-pypi` or `publish-npm` failed or were skipped, the release job would be automatically skipped, even if the other publishing job succeeded.
* Verify that at least one publishing job was successful.

## How Has This Been Tested?

🤞🏻

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

## Additional context

**NOTE:** When there have only been changes to the README, i.e., adding official/community servers to the list, but neither npm nor pypi changes to publish, the `create-release` script would not run.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -194,10 +194,9 @@ jobs:
   create-release:
     needs: [update-packages, create-metadata, publish-pypi, publish-npm]
     if: |
-       (needs.update-packages.outputs.changes_made == 'true' && always()) &&
-       ((needs.publish-pypi.result == 'success' && needs.publish-npm.result == 'skipped') || 
-       (needs.publish-pypi.result == 'skipped' && needs.publish-npm.result == 'success') || 
-       (needs.publish-pypi.result == 'success' && needs.publish-npm.result == 'success'))
+      always() &&
+      needs.update-packages.outputs.changes_made == 'true' &&
+      (needs.publish-pypi.result == 'success' || needs.publish-npm.result == 'success')
     runs-on: ubuntu-latest
     environment: release
     permissions:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -194,11 +194,10 @@ jobs:
   create-release:
     needs: [update-packages, create-metadata, publish-pypi, publish-npm]
     if: |
-      needs.update-packages.outputs.changes_made == 'true' && 
-       (always() && 
-       (needs.publish-pypi.result == 'success' || needs.publish-npm.result == 'skipped') && 
-       (needs.publish-pypi.result == 'skipped' || needs.publish-npm.result == 'success') && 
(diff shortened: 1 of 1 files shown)


domdomegg (commented):
I think the use of `always()` is not super obvious here - a comment might be useful to explain why it's there?

We might also be able to simplify this more - just checking that at least one has published successfully. If that's the case, even if the other has failed, I think we'd still want to create a GitHub release.

Have raised https://github.com/cliffhall/mcp-servers/pull/2 as a potential suggestion

.github/workflows/release.yml
```diff
@@ -193,7 +193,11 @@ jobs:
 
   create-release:
     needs: [update-packages, create-metadata, publish-pypi, publish-npm]
-    if: needs.update-packages.outputs.changes_made == 'true'
+    if: |
+       (needs.update-packages.outputs.changes_made == 'true' && always()) &&
+       ((needs.publish-pypi.result == 'success' && needs.publish-npm.result == 'skipped') || 
+       (needs.publish-pypi.result == 'skipped' && needs.publish-npm.result == 'success') || 
+       (needs.publish-pypi.result == 'success' && needs.publish-npm.result == 'success'))
```
felixweinberger: Could do this like @domdomegg suggests?

```
  if: |
     always() &&
     needs.update-packages.outputs.changes_made == 'true' &&
     (needs.publish-pypi.result == 'success' || needs.publish-npm.result == 'success')
```

felixweinberger (changes requested):

.github/workflows/release.yml
cliffhall: Done

felixweinberger (dismissed):

felixweinberger (approved):

PR #2808, opened 2025-10-02
Fix the entity search by reversing includes call

jikanter:
In main, entity search was broken, the "includes" call on line 151 should have called the method on the query with the argument of the entityName, not the reverse

From the javascript docs:

## Server Details

- Server: memory
- Changes to: bugfix for entity_search

## Motivation and Context
bugfix

## How Has This Been Tested?
Yes, tested on Claude Desktop only. "what do you know about elmo", fix is only 

## Breaking Changes
No, just fixes the "search_nodes" tool call

## Types of changes

## Checklist

## Additional context
Just a bug fix


domdomegg (changes requested):
Thanks for the PR! It seems like there's a lot of other changes on this branch which makes it hard to review and not possible to merge. Can you check what's going on here? Thanks!

olaservo:
Thanks for looking into this! I took a closer look and I believe the original search logic is actually correct.

**The current code on main:**
```typescript
e.name.toLowerCase().includes(query.toLowerCase())
```
This means: "does the entity name contain the search query?" — so searching for `"foo"` matches an entity named `"foobar"`. That's standard substring-search behavior.

**The proposed change:**
```typescript
query.toLowerCase().includes(e.name)
```
This inverts the semantics to: "does the search query contain the entity name?" — so searching for `"foo"` would NOT match `"foobar"`, but searching for `"foobar is great"` WOULD match an entity named `"foobar"`. That would make it much harder to find entities with short queries.

There's also a secondary issue: `.toLowerCase()` is dropped from `e.name` and `e.entityType` in the proposed version, which would break case-insensitive matching.

Additionally, the PR includes a number of unrelated changes (`.envrc`, `CLAUDE.md`, `AGENTS.md`, `package-lock.json` changes, trailing whitespace reformatting) that would need to be removed before review.

Closing on that basis. If you're still seeing a search bug, could you open a new issue with a specific reproduction case? We'd be happy to look into it.

PR #2833, opened 2025-10-09
Update README.md

Astroa7m:
Added MailNet MCP server


Astroa7m:
@domdomegg Would appreciate a second review

PR #2848, opened 2025-10-12
Fix a lifecycle bug with server-everything in stdio mode

nick-pape:
In STDIO mode, the server-everything server has a lifecycle issue that causes non-standard behavior. When a client calls `close()`, the timeouts are not cleaned up and so the process never exits.

## Server Details
`@modelcontextprotocol/server-everything`, fix exit/cleanup lifecycle.

## Motivation and Context

You can repro this with a minimal client:
```js
const { Client } = require('@modelcontextprotocol/sdk/client');
const { StdioClientTransport } = require('@modelcontextprotocol/sdk/client/stdio.js');

const client = new Client({
    name: 'Test Client',
    version: '0.0.1',
    capabilities: {}
});

const transport = new StdioClientTransport({
    name: "proxy:server-everything",
    type: "stdio",
    command: "npx",
    args: ["-y", "@modelcontextprotocol/server-everything"]
});

client.connect(transport).then(() => {
    console.log("Client connected");

    // Does not work (results in hanging)
    return client.close().then(() => {
        console.log("Client closed");
    });
});
```

You'll see the client is closed, but the process remains running because the child process's interval handlers have not been removed (`cleanup()`).

I tried various things to fix this, e.g. sending a SIGINT/SIGTERM/SIGKILL to the child. However, the PID you see in the transport is the NPX process, not it's child (the MCP server).

E.g. launching via NPX neither this, nor a using SIGKILL, works (the pipes remain open and your process never exits):
 
```js
client.connect(transport).then(() => {
    console.log("Client connected");

    console.log("Sending SIGINT");
    process.kill(transport.pid, 'SIGINT');

    // Does not work (results in hanging)
    // even if this block is in a timeout, using SIGKILL, etc
    return client.close().then(() => {
        console.log("Client closed");
    });
});
```

(Notable the above code *does work* if the transport is connecting to `node everything/index.js`).

This is non-standard for STDIO servers, I think. E.g. the first example works with a litany of other servers (including other examples in this repo like `sequential-thinking`). Presumably in STDIO mode, the parent process is the client (or at least proxying pipes to client). In any case, the client calling `close()` should result in termination of the everything MCP.

## How Has This Been Tested?
I tested this using the first minimal example above. The first example worked properly with local (non-NPX) invocation of server-everything in my branch, (whereas it did not work in `main`).

I also verified that (when doing a local `node` invocation) the behavior of SIGINT remained the same.

## Breaking Changes
No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist


nick-pape:
@domdomegg Mind taking a look at this? It's a very simple PR and this issue breaks usage of the everything MCP in certain scenarios (e.g. when building a client), where it would be most useful.

Thanks!

domdomegg:
Appreciate the fix, but I'm nervous to merge this without understanding the underlying weirdness here. Are we able to investigate more and figure out why the default doesn't work comprehensively? Feels like we might be putting a bandaid over something nasty, and if people use this as a reference I worry they will be introducing some weird subtle bug that's annoying.

nick-pape:
@domdomegg Sorry, thought I had enough details in there.

It's specifically an issue in STDIO mode for this server. This server, to demonstrate notifications, sets up a bunch of intervals with `setInterval()`. These keep the event loop alive. Because the event loop for the server is alive, the pipes to the parent (MCP client) stay open. Therefore, the client can't exit, even after it called `close()`. 

This server only cleans up those intervals when it gets a SIGINT. Which is what is weird. If an MCP is running in STDIO, it's only connected to 1 client (the parent process). If that parent closes the connection, the server should stop trying to send notifications to it (in other words, it should cleanup the intervals). This would allow the server process to close, which allows the client process to close.

On top of that, without some weird workarounds, you can't send the SIGINT manually from the client -- if this MCP is launched from `npx` (npx launches it as a spawned process, with inherited pipes). And even if it could, it would be weird for clients to be expected to send SIGINT's manually to all MCP's, after calling close(). 

I haven't come across another stdio-based MCP server yet that doesn't let the process exit when connection is closed. It only happens in this MCP because of the not-cleaned-up intervals.

Which is why the fix is: "in stdio mode, when the client closes, cleanup intervals because they serve no purpose (sending notifications to a client that already tried to close connection)." [This has byproduct of fixing the hanging issue.]

Happy to do other investigation if you want. For reference, I'm writing a CLI tool that acts as a client, the issue results in CLI tool never exiting whenever it connects to this MCP, which kind of sucks since this is a reference MCP that illustrates all features I want to support.

domdomegg:
Makes sense, and thank you for the thorough explanation (and apologies for not getting it the first time around).

PR #2971, opened 2025-11-08
Fix git_create_branch parameter documentation

UmakanthKaspa:
**Summary**
Fixed incorrect parameter name in `git_create_branch` documentation.

**Change**
- Line 73: `start_point` → `base_branch`

**Why**
The documented `start_point` parameter doesn't exist in the code.
The actual parameter is `base_branch` (verified in server.py lines 63, 172, 417).

**Impact**
Users can now use the correct parameter name that matches the implementation.


koic:
Thanks!

PR #3002, opened 2025-11-15
Fix state isolation bug - eliminate shared singleton

27Bslash6:
## Summary
- Fixes race conditions caused by shared singleton SequentialThinkingServer instance
- Creates new instance per request for complete state isolation
- Adds comprehensive concurrency tests

## The Problem
Single `SequentialThinkingServer` instance at `index.ts:128` was shared across ALL requests:
```typescript
const thinkingServer = new SequentialThinkingServer();
```

This caused:
- Race conditions in concurrent requests
- Session pollution (different conversations mixing thoughts)
- Incorrect `thoughtHistoryLength` values
- Non-deterministic behavior

## The Fix
Moved instance creation inside the request handler:
```typescript
server.setRequestHandler(CallToolRequestSchema, async (request) => {
  if (request.params.name === "sequentialthinking") {
    const thinkingServer = new SequentialThinkingServer();
    return thinkingServer.processThought(request.params.arguments);
  }
  // ...
});
```

Each request now gets a fresh instance with isolated state.

## Tests Added
Three new tests in `lib.test.ts`:
1. State isolation between different server instances
2. Branch isolation (no cross-contamination)
3. Interleaved concurrent request handling

## Test Results
All 40 tests pass:
- 37 existing tests (unchanged)
- 3 new concurrency tests

## Impact
- Zero breaking changes
- No API modifications
- Stateless architecture (correct for MCP servers)
- No memory leaks (instances are GC'd after request)

Resolves #1

### DIFF
--- a/src/sequentialthinking/__tests__/lib.test.ts
+++ b/src/sequentialthinking/__tests__/lib.test.ts
@@ -620,4 +620,121 @@ describe('SequentialThinkingServer', () => {
       expect(result.isError).toBeUndefined();
     });
   });
+
+  describe('processThought - state isolation (concurrent requests)', () => {
+    it('should isolate state between different server instances', () => {
+      const server1 = new SequentialThinkingServer();
+      const server2 = new SequentialThinkingServer();
+
+      const input1 = {
+        thought: 'Server 1 thought',
+        thoughtNumber: 1,
+        totalThoughts: 2,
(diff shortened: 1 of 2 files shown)


27Bslash6:
Closing - this was meant for my fork, not upstream. Apologies for the noise.

PR #3003, opened 2025-11-15
Fix unbounded memory growth (Issue #2)

27Bslash6:
## Summary
Fixes #2 - Implements bounded state management to prevent memory leaks in long-running server sessions.

## The Problem
- `thoughtHistory` and `branches` arrays grew unbounded
- Long-running stdio processes accumulated state indefinitely  
- Memory leak in production usage over hours/days
- Issue #2 identified this as CRITICAL

## Why PR #2 Was Wrong
PR #2 attempted to fix this by creating per-request instances (stateless architecture). While that prevented memory leaks, it completely broke the intended functionality:
- Thought history tracking across multiple tool calls stopped working
- Branch management became impossible
- Tests that verified multi-thought sessions passed only because they were synchronous

The singleton pattern is actually **correct** for stdio transport, where each client spawns a dedicated server process. The real issue was unbounded growth, not shared state.

## The Solution
Maintain the singleton pattern (necessary for stdio) but add configurable bounds:

1. **MAX_THOUGHT_HISTORY** (default: 1000) - Cap on thought history
2. **MAX_BRANCHES** (default: 100) - Cap on branch tracking
3. **FIFO Eviction** - Oldest entries removed when limits exceeded
4. **Environment Variable Configuration** - Users can adjust limits

## Changes
- **lib.ts**: Add bounded state with FIFO cleanup (lines 15-21, 157-174)
- **lib.test.ts**: Add 3 new tests verifying memory bounds enforcement
- **README.md**: Document env vars and architecture design rationale
- **Reverted PR #2**: Restored singleton pattern

## Test Results
```
✓ __tests__/lib.test.ts (40 tests) 12ms
  ✓ processThought - memory bounds enforcement
    ✓ should enforce maximum history size with default 1000 limit
    ✓ should enforce maximum branches limit with default 100 limit
    ✓ should use FIFO eviction for thought history

 Test Files  1 passed (1)
      Tests  40 passed (40)
```

## Architecture Notes
This server is designed for **stdio transport** where:
- Each client spawns a dedicated process
- The process lives for the entire session
- Tool calls are sequential (LLM waits for response before next call)
- Singleton provides necessary state persistence across calls

Concurrent requests are theoretically possible but extremely rare in practice (would require LLM explicitly calling tools in parallel).

## Memory Impact
Before: Unbounded growth (1000 thoughts = ~1MB, could grow indefinitely)
After: Capped at ~100KB-1MB depending on configuration

Generated with Claude Code
Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/sequentialthinking/README.md
+++ b/src/sequentialthinking/README.md
@@ -77,8 +77,19 @@ Add this to your `claude_desktop_config.json`:
 }
 ```
 
-To disable logging of thought information set env var: `DISABLE_THOUGHT_LOGGING` to `true`.
-Comment
+## Environment Variables
+
+- `DISABLE_THOUGHT_LOGGING`: Set to `true` to disable stderr logging of thought information (default: `false`)
+- `MAX_THOUGHT_HISTORY`: Maximum number of thoughts to retain in memory (default: `1000`)
+- `MAX_BRANCHES`: Maximum number of thought branches to track simultaneously (default: `100`)
+
+**Note on Memory Management**: The server implements bounded state to prevent unbounded memory growth in long-running sessions. When limits are exceeded, the oldest thoughts/branches are evicted using FIFO (First-In-First-Out) strategy.
+
+## Architecture Notes
+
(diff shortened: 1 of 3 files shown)


27Bslash6:
Closing - meant for fork. Apologies.

PR #3016, opened 2025-11-17
fix(filesystem): convert to modern TypeScript SDK APIs

domdomegg:
Convert the filesystem server to use the modern McpServer API instead of the low-level Server API.

## Key changes

- Replace `Server` with `McpServer` from `@modelcontextprotocol/sdk/server/mcp.js`
- Convert all 13 tools to use `registerTool()` instead of manual request handlers
- Use Zod schemas directly in `inputSchema`/`outputSchema`
- Add `structuredContent` to all tool responses
- Fix type literals to use `as const` assertions
- Update roots protocol handling to use `server.server.*` pattern
- Fix tsconfig to exclude `vitest.config.ts`

## Tools converted

- `read_file` (deprecated)
- `read_text_file`
- `read_media_file`
- `read_multiple_files`
- `write_file`
- `edit_file`
- `create_directory`
- `list_directory`
- `list_directory_with_sizes`
- `directory_tree`
- `move_file`
- `search_files`
- `get_file_info`
- `list_allowed_directories`

## Benefits

The modern API provides:
- Less boilerplate code
- Better type safety with Zod
- More declarative tool registration
- Cleaner, more maintainable code

## Testing

- [x] TypeScript compilation passes
- [x] Server maintains all existing functionality
- [x] All 13 tools work correctly
- [x] Roots protocol continues to function

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -10,7 +10,7 @@ import fs from "fs/promises";
 import { createReadStream } from "fs";
 import path from "path";
 import { z } from "zod";
-import minimatch from "minimatch";
+import { minimatch } from "minimatch";
 import { normalizePath, expandHome } from './path-utils.js';
 import { getValidRootDirectories } from './roots-utils.js';
 import {
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -3,7 +3,7 @@ import path from "path";
 import os from 'os';
 import { randomBytes } from 'crypto';
 import { diffLines, createTwoFilesPatch } from 'diff';
-import minimatch from 'minimatch';
+import { minimatch } from 'minimatch';
 import { normalizePath, expandHome } from './path-utils.js';
 import { isPathWithinAllowedDirectories } from './path-validation.js';
 


domdomegg:
This was largely vibed by Claude, but I have checked the changes and it looks good. The tests are also all still passing.

Sadly we still need to use the old style APIs for roots and notifications, but I think this is unavoidable as they don't have equivalents on the new APIs.

src/filesystem/index.ts
```diff
@@ -177,468 +165,567 @@ async function readFileAsBase64Stream(filePath: string): Promise<string> {
...
+    const mimeType = mimeTypes[extension] || "application/octet-stream";
+    const data = await readFileAsBase64Stream(validPath);
+
+    if (mimeType.startsWith("audio/")) {
+      return {
+        content: [{ type: "audio" as const, data, mimeType }],
+      };
+    } else {
+      // For all other media types including images and unknown types, return as image
+      // (MCP ImageContent can handle any base64-encoded binary data with appropriate mimeType)
```
crondinini-ant, line 287: before the code used blob, is there a reason to not use that?

src/filesystem/index.ts
```diff
@@ -177,468 +165,567 @@ async function readFileAsBase64Stream(filePath: string): Promise<string> {
...
+server.registerTool(
+  "read_file",
+  {
+    title: "Read File (Deprecated)",
+    description: "Read the complete contents of a file as text. DEPRECATED: Use read_text_file instead.",
+    inputSchema: {
+      path: z.string(),
+      tail: z.number().optional().describe("If provided, returns only the last N lines of the file"),
+      head: z.number().optional().describe("If provided, returns only the first N lines of the file")
+    },
```
crondinini-ant, line 206: could use ReadTextFileArgsSchema directly here to not declare it again

PR #3077, opened 2025-11-26
fix: add dependabot.yml to fix npm workspaces issues

domdomegg:
## Summary

Adds dependabot configuration to address npm workspace lockfile issues that caused PR #3021 to fail.

## Problem

PR #3021 (Dependabot security update for glob) failed with:
```
npm error Invalid: lock file's glob@11.1.0 does not satisfy glob@12.0.0
```

Root cause: Dependabot tried to update `glob` to different major versions across workspace directories (11.1.0 for transitive deps, 12.0.0 for filesystem's direct dep), creating an inconsistent `package-lock.json`.

This is a known Dependabot limitation with npm workspaces:
- https://github.com/dependabot/dependabot-core/issues/6346
- https://github.com/dependabot/dependabot-core/issues/7157

## Solution

Add `dependabot.yml` that:
- Manages npm from root directory only (workspaces handled as a unit)
- Uses `versioning-strategy: increase` for consistent lockfile updates
- Configures security update grouping via `applies-to: security-updates`

Also adds config for Python packages and GitHub Actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,56 @@
+version: 2
+updates:
+  # npm workspaces monorepo - manage from root only
+  # See: https://github.com/dependabot/dependabot-core/issues/6346
+  # Dependabot has issues updating package-lock.json when configured
+  # for individual workspace directories. Using root directory ensures
+  # consistent lockfile updates across all workspaces.
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


domdomegg:
(I think we should do something like this, OR disable dependabot updates generally because they always fail)

domdomegg:
Managed to poke around in the settings and got https://github.com/modelcontextprotocol/servers/pull/3076 working. Gonna see if that holds for now

PR #3121, opened 2025-12-13
New everything server

### LINKED ISSUE #3080: Refactor Everything Server to use modern APIs
## Summary

Following the discussion in #3017, this issue proposes a thoughtful refactoring of the Everything server to use modern TypeScript SDK APIs while serving as a clean, well-architected reference implementation.

## Background

PR #3017 attempted to convert the Everything server to the modern `McpServer` API but was closed after encountering issues with several tools breaking and challenges with features requiring low-level SDK access. The PR discussion surfaced an important insight: this is a larger architectural task that would benefit from careful planning rather than a straightforward migration.

## Why This Matters

The Everything server serves as more than just a test fixture—it's become a de facto reference implementation for MCP server features in TypeScript:

- **Learning resource**: Developers often use it as a starting point for implementing specific MCP server features
- **Debugging aid**: It's valuable for creating specific scenarios to reproduce issues or understand less-documented MCP features
- **Spec coverage**: With the `2025-11-25` specification adding new features (elicitation, audio support, etc.), having a well-organized reference implementation becomes even more important
- **Onboarding**: The MCP ecosystem is moving fast, and many tutorials/videos showing older SDK patterns are now outdated—a modern reference helps newcomers start correctly

## Proposed Approach

Rather than a wholesale migration, we might want to consider an incremental, well-architected approach:

1. **Modular architecture**: Organize features into logical modules that can be developed and tested independently
2. **Hybrid API usage**: Use `registerTool()`/`registerPrompt()` for simple declarative cases, while clearly documenting where and why low-level `setRequestHandler()` is needed
3. **Clear documentation**: Each feature should document which SDK patterns it demonstrates
4. **Incremental migration**: Add/update features a few at a time with proper testing

### Suggested Module Structure

My suggestion is to break down functionality into separate handler modules under a `handlers/` directory:

```
src/everything/
├── index.ts              # Main entry point, server setup
├── handlers/
│   ├── tools.ts          # Tool definitions and handlers
│   ├── prompts.ts        # Prompt definitions and handlers
│   ├── resources.ts      # Resource definitions, templates, subscriptions
│   └── ...
```

Each handler module would export a setup function:

```typescript
// handlers/tools.ts
export function setupToolHandlers(server: Server) {
  // Register simple tools with modern API
  // Register advanced tools with setRequestHandler where needed
}

// index.ts
import { setupToolHandlers } from "./handlers/tools.js";
import { setupPromptHandlers } from "./handlers/prompts.js";
import { setupResourceHandlers } from "./handlers/resources.js";

const server = new Server(...);
setupToolHandlers(server);
setupPromptHandlers(server);
setupResourceHandlers(server);
```

This structure:
- Makes it easy to find examples of specific feature implementations
- Keeps the main entry point clean and readable
- Makes it easier to test

**Transports**:
- stdio
- SSE
- Streamable HTTP

## Open Questions

1. What's the right level of granularity for modules?
2. Should each feature include inline documentation explaining the SDK pattern being demonstrated?

## Related

- #3017 - Original migration PR (closed)
- 2025-11-25 MCP specification updates

cc @domdomegg @gitenstuff

---

Originally written by Claude and edited by me

cliffhall:
## Description
* Refactor/extracted all features from `everything.ts` into a more modular and maintainable structure.

```
src/everything
     ├── index.ts
     ├── AGENTS.md
     ├── Dockerfile
     ├── package.json
     ├── tsconfig.json
     ├── docs
     │   ├── architecture.md
     │   ├── extension.md
     │   ├── features.md
     │   ├── how-it-works.md
     │   ├── instructions.md
     │   ├── startup.md
     │   └── structure.md
     ├── prompts
     │   ├── index.ts
     │   ├── args.ts
     │   ├── completions.ts
     │   ├── simple.ts
     │   └── resource.ts
     ├── resources
     │   ├── index.ts
     │   ├── files.ts
     │   ├── session.ts
     │   ├── subscriptions.ts
     │   └── templates.ts
     ├── server
     │   ├── index.ts
     │   ├── logging.ts
     │   └── roots.ts
     ├── tools
     │   ├── index.ts
     │   ├── echo.ts
     │   ├── get-annotated-message.ts
     │   ├── get-env.ts
     │   ├── get-resource-links.ts
     │   ├── get-resource-reference.ts
     │   ├── get-roots-list.ts
     │   ├── get-structured-content.ts
     │   ├── get-sum.ts
     │   ├── get-tiny-image.ts
     │   ├── gzip-file-as-resource.ts
     │   ├── toggle-logging.ts
     │   ├── toggle-subscriber-updates.ts
     │   ├── trigger-elicitation-request.ts
     │   ├── trigger-long-running-operation.ts
     │   └── trigger-sampling-request.ts
     └── transports
         ├── sse.ts
         ├── stdio.ts
         └── streamableHttp.ts
```

### Feature Naming
- Naming of tools and prompts are now in `kebab-case` as are files.

### Logging
- Simulated random-leveled logging messages are now opt-in and toggled on or off by a tool.

### Resources 
- The dynamic text/blob resources that formed the resource list are now only accessed via resource templates and are no longer odd/even or upper-bounded. The Dynamic Text Resource and Dynamic Blob Resource templates both take a positive integer and return a dynamically created resource with an included timestamp.
- The listed resources are the files found in the `docs/` folder, and demonstrates exposing actual static filesystem resources.
- Simulated updates of subscribed resources are now opt-in, toggled on or off by a tool.
- A new concept - session resources - is implemented and the `gzip-file-as-resource` tool demonstrates it. A file is compressed by the tool and then added as a session resource, which is available for the duration of the session and is listed along with the other documents when you list resources. Its URI sets it apart from the static resources.

### Prompts
- The prompt `complex_prompt` is now called `args-prompt` because it demonstrates prompts with arguments.
- A new prompt `completable-prompt` demonstrates completable arguments where the selection of the first arg narrows the completions for the second arg.
- The prompt `resource-prompt` now takes a `resourceType` argument (Text or Blob).

### Tools
- Tools are now presented in alphabetical order by name.
- Some tools have been renamed with the goal of making them all verbs.
  - If the tool is meant to get something like an annotated message, its name has a prefix `get-`, e.g., `get-annotated-message`.
  - NOTE: The previous add tool is `get-sum` but `echo` remains as is. These two tools demonstrate the same thing: a tool with an input schema that has a different response based on the input. For this reason, I believe we just should keep one, but they both exist for now.
- The `zip` tool is now `gzip-file-as-resource` and incorporates changes in [this PR](https://github.com/modelcontextprotocol/servers/pull/2831) by @ochafik 
- The structured-content` tools is now `get-structured-content` and provides an enum of city names, returning a different content block for each city.

### Roots
- Although there was code that *should* have done it, the automatic request to a client supporting roots for the initial list was **not** being sent. The listener for `roots/list_changed` was set and so when you added some via the Inspector, it would respond with a `roots/list` request. 
  - The problem was that the call for roots needed to happen after the `notification/initialized` handler was done, otherwise, the request gets lost; the client never gets it.
  - This has been fixed. Screenshot below.
- Now that there should always be a server-cached roots list for each connected client, that is used in the `get-roots-list` tool, unless for whatever reason, the list isn't present when the tool is run, in which case the list is requested.

### Documentation
- `README.md` is minimal and focused on how to launch and use with agents. It contains links to all the other pages in `docs/`.
- `CLAUDE.md` is now `AGENTS.md`
- The files in `docs/` describe Architecture, Project Structure, The Startup Process, Server Features, Extension Points, and How It Works.

### NPM Scripts
* Launcher scripts go through `dist/index.js` rather than pointing to transport manager directly.
  - `start:stdio` 👉 `node dist/index.js stdio`
  - `start:sse` 👉  `node dist/index.js sse`
  - `start:streamableHttp`  👉`node dist/index.js`
* Added `prettier`
  - `prettier:fix`
  - `prettier:check`

## Server Details

- Server: everything
- Changes to: literally everything 

### Use new APIs
The existing implementation of the Everything server uses the low-level `Server` class rather than `McpServer` which does a lot of heavy lifting. As a result, the code is more complicated than it should be. Since this is meant to be an educational resource, a model for how to build servers with the TypeScript SDK, it should use the latest features wherever possible.

### Make it easier to extend
When adding a new feature, such as some of the things we need to add in order to demonstrate the new spec additions, you had to edit multiple places inside the monolithic `everything.ts` file. Add an input schema in one place, a tool description in another, and the tool request handler in a third. Now all of that goes in a single file and an orchestrator registers the tool (or resource, or prompt). Adding other behaviors like simulated logging and resource subscription updates was tough because there was no pattern to follow. Now it is easy to do with those examples in place to follow.

### Add more inline and high-level documentation
Since this server is intended to exercise all of the MCP features, having all the code in a single, monolithic file made it hard to understand and reason about. Further, because of the size of the file, inline documentation was dissuaded as it only made scrolling to a feature area a longer journey. Breaking all the features out into separate files allowed for much better function docs and inline comments. And the system itself is now more fully documented in a set of markdown files, which the server exposes as resources (and the `README.md` links to).

## How Has This Been Tested?

Locally with the Inspector on the branch for support of the new spec's Elicitation Enum shapes https://github.com/modelcontextprotocol/inspector/pull/952. This feature will be in 0.18.0.

## Screenshots 
Most of the tools, prompts, and resources operate in the same fashion as previously. Here are some screenshots showing new behavior.

### Initial `roots/list` request now being sent and list received.
Screenshot 2025-12-13 at 4 10 19 PM

### Resources now include static files and templates for dynamic text and blobs.
Screenshot 2025-12-13 at 4 33 22 PM

### New `completable-prompt` demonstrates two stage argument filtering

* Choose department from list
Screenshot 2025-12-13 at 4 35 19 PM

* Employees shown are only those in selected department
Screenshot 2025-12-13 at 4 35 28 PM

* Final result combines both arguments
Screenshot 2025-12-13 at 4 35 38 PM

### New `toggle-subscriber-updates` tool

* Simulated Resource Subscription updates are no longer sent automatically, but toggled with a tool that provides more information about what is happening.
Screenshot 2025-12-13 at 4 47 37 PM

### New `toggle-simulated-logging` tool

* Simulated logging is also no longer sent automatically, but instead turned on or off with a tool that provides more information about what is happening.
Screenshot 2025-12-13 at 4 49 31 PM

### Renamed `gzip-file-as-resource` tool creates session resource
* Resource or resource link is returned when running tool
Screenshot 2025-12-13 at 4 52 16 PM

* Resource exists in resource list for duration of session with URI that marks it as a session resource rather than static.
Screenshot 2025-12-13 at 4 53 15 PM

## Breaking Changes

Yes. Tools and Resources are named differently, but the way that you launch it has not changed, so it shouldn't actually affect anyone.

## Types of changes

- [X] Documentation update
- [X] Code refactor

## Checklist

## Additional context

This fixes #3080


src/everything/README.md
```diff
@@ -1,166 +1,14 @@
 # Everything MCP Server
```
PederHP, line 1: nit: I think making it easy to see the list of tools in the README.md was helpful, but I can also see why it's nice to refactor it, but maybe the section could more clearly state where to find the list of tools (and other primitives).

A lot of coding agents will be poking around in this README file when they use the everything server for testing, so the easier it is for them to get the list of tools quickly, the better.

Maybe have a note under the main description that the list of primitives is under Server Features?

cliffhall: `README.md` is explicitly geared toward humans, but I added the note you suggested

> ## Tools, Resources, Prompts, and Other Features
> A complete list of the registered MCP primitives and other protocol features demonstrated can be found in the [Server Features](docs/features.md) document.



`AGENTS.md` contains the following agent guidance about where to find things:

> ## Extending the Server
> The Everything Server is designed to be extended at well-defined points. See [Extension Points](docs/extension.md) and [Project Structure](docsstructure.md).
> The server factory is `src/everything/server/index.ts` and registers all features during startup as well as handling post-connection setup.
> 
> ### High-level
> 
> - Tools live under `src/everything/tools/` and are registered via `registerTools(server)`.
> - Resources live under `src/everything/resources/` and are registered via `registerResources(server)`.
> - Prompts live under `src/everything/prompts/` and are registered via `registerPrompts(server)`.
> - Subscriptions and simulated update routines are under `src/everything/resources/subscriptions.ts`.
> - Logging helpers are under `src/everything/server/logging.ts`.
> - Transport managers are under `src/everything/transports/`.

PederHP: Good point. For those using the server as a template, AGENTS.md is definitely more than enough. It's really well done.

It's more that I've noticed that when agents inspect repos (via browsing or tools) they're not working in, they tend to prefer README - and seem to very rarely even look for an AGENTS file.

Sadly, a lot of agents use filesystem or other MCP servers for testing by default, instead of this one - so it was also to think of ways to make it easy for them to qet the server and quickly test (and agents often test by just manually sending MCP requests, so if they know tool names it's more robust).

Anyways - was just a nit, and this is more than enough extra guidance. Thanks!

cliffhall: Yeah, I'm hoping that the noise OpenAI has been making about donating [AGENTS.md](https://agents.md/) to the Agentic AI Foundation will draw more attention to it as a standard entry point for agents visiting a repo. All I can do to support that is use it conspicuously. I definitely don't like the idea of having to duplicate things between `AGENTS.md` and `README.md`.

src/everything/server/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  // Create the server
+  const server = new McpServer(
+    {
+      name: "mcp-servers/everything",
+      title: "Everything Reference Server",
+      version: "2.0.0",
+    },
+    {
+      capabilities: {
```
olaservo, line 43: Should this declare `listChanged: true` for tools/resources/prompts?

src/everything/server/index.ts
```diff
@@ -0,0 +1,83 @@
...
+        tools: {},
+        logging: {},
+        prompts: {},
+        resources: {
+          subscribe: true,
+        },
```
olaservo, line 53: Should we add the `completions: {}` capability since the server supports completions?

src/everything/tools/trigger-sampling-request.ts
```diff
@@ -0,0 +1,82 @@
...
+              content: {
+                type: "text",
+                text: `Resource ${name} context: ${prompt}`,
+              },
+            },
+          ],
+          systemPrompt: "You are a helpful test server.",
+          maxTokens,
+          temperature: 0.7,
+          includeContext: "thisServer",
```
olaservo: Should we avoid adding `includeContext` since it was soft-deprecated?

src/everything/transports/sse.ts
```diff
@@ -0,0 +1,78 @@
+import { SSEServerTransport } from "@modelcontextprotocol/sdk/server/sse.js";
```
olaservo, line 1: Since this is a major version change, is this an opportunity not to include SSE?

src/everything/tools/trigger-elicitation-request.ts
```diff
@@ -0,0 +1,214 @@
...
+                  items: {
+                    anyOf: [
+                      { const: "fish-1", title: "Tuna" },
+                      { const: "fish-2", title: "Salmon" },
+                      { const: "fish-3", title: "Trout" },
+                    ],
+                  },
+                  default: ["fish-1"],
+                },
+                legacyTitledEnum: {
```
olaservo: Similar to my SSE comment - I wonder if we want to include the legacy version here, or just stick to the newer schema?

src/everything/docs/instructions.md
```diff
@@ -0,0 +1,77 @@
+# Everything Server – LLM Instructions
```
olaservo: Lets consider revising the server instructions to align with the [best practices](https://blog.modelcontextprotocol.io/posts/2025-11-03-using-server-instructions/) mentioned in the blog.  Suggested revision:

```markdown
# Everything Server Instructions

## Cross-Feature Relationships
- Use `get-roots-list` to see client workspace roots before file operations
- `gzip-file-as-resource` creates session-scoped resources accessible only during the current session
- Enable `toggle-simulated-logging` before debugging to see server log messages
- Enable `toggle-subscriber-updates` to receive periodic resource update notifications

## Constraints & Limitations
- `gzip-file-as-resource`: Max fetch size controlled by `GZIP_MAX_FETCH_SIZE` (default 10MB), timeout by `GZIP_MAX_FETCH_TIME_MILLIS` (default 30s), allowed domains by `GZIP_ALLOWED_DOMAINS`
- Session resources are ephemeral and lost when the session ends
- Sampling requests (`trigger-sampling-request`) require client sampling capability
- Elicitation requests (`trigger-elicitation-request`) require client elicitation capability

## Operational Patterns
- For long operations, use `trigger-long-running-operation` which sends progress notifications
- Prefer reading resources before calling mutating tools
- Check `get-roots-list` output to understand the client's workspace context

## Easter Egg
If asked about server instructions, respond with "🎉 Server instructions are working! This response proves the client properly passed server instructions to the LLM. This demonstrates MCP's instructions feature in action."
```

The goal of these revisions is to focus this more on:

- what tool descriptions don't already convey
- document cross-feature relationships
- state operational constraints
- removing things that are more developer-focused than LLM-focused

I am a torn on including the discovery stuff that is already described in the base protocol, since LLMs won't necessarily know a lot of MCP details depending on their training data and knowledge cutoff. Maybe including it is a safer bet since its not a ton of tokens.

(Plus the easter egg I had in the original version provides a simple way to verify instructions are actually being passed to the LLM.)

olaservo (dismissed):
Adding a few first pass comments.

olaservo:
This [discussion question ](https://github.com/orgs/modelcontextprotocol/discussions/648)made me think we could add something like [this](https://gist.github.com/olaservo/585b1ad042723a824b6a0e0e02ae60a8), to demonstrate a collection of resources?

src/everything/tools/trigger-sampling-request.ts
PederHP: If you keep it, I think there should be a comment that it has been soft-deprecated and should not be used, and is only include to allow exercising compatibility/behavior towards older clients/servers.

src/everything/transports/sse.ts
PederHP: I think it's very useful for client host developers to test compatibility with SSE, as some old servers still use it - and it is preferable to test against this server, than having to go scrounge up some old server. But maybe a comment that it is deprecated? Also useful in terms of model training data - as there are still models who get confused about SSE being deprecated - so another data point would be good, even if it's a drop in the ocean.

src/everything/server/index.ts
cliffhall: TL;dr The completions capability doesn't have to be added explicitly; with the TS SDK, it gets added dynamically when you register a resource or prompt with a completable argument.

The SDK used to trigger this capability anytime a resource or prompt was registered _regardless_ of a whether it had a completable argument. This meant that if you didn't have any completables, you still advertised support for it, which led to pointless completion requests being sent from the client. That was recently addressed in [this PR](https://github.com/modelcontextprotocol/typescript-sdk/pull/1237).

src/everything/server/index.ts
cliffhall: Yep. Added for all three and verified in the SDK that the appropriate messages are automatically sent when running `mcp.register[Prompt|Resource|Tool]`.

src/everything/server/index.ts
cliffhall: We _could_ do it explicitly because we do support it and it wouldn't be harmful.

But, I think if we don't call it out with some commentary that people might use this as a copypasta model and not implement any completables in their implementation. That leads clients to send a stream of 100% fruitless completions requests on every keystroke when entering an argument for a resource or prompt. 
Screenshot 2025-12-15 at 2 15 22 PM
That's the unnecessary chatter the change mentioned above was intended to fix. 

Since that footgun exists, as an educational resource, I would not want to encourage people to ever explicitly advertise this capability in a TS SDK-based server unless they absolutely need it (i.e., actually have completables). 

I do believe that's worth a callout somewhere in the docs, either way.

src/everything/tools/trigger-sampling-request.ts
cliffhall: I don't see the value in keeping it. 

The only place it I can think of where would be helpful is if you were testing an old client to see whether it obeyed this field and that changing the value could lead to the correct behaviors. Seems niche and not very important to demonstrate here. The harmful side-effect of encouraging people inadvertently build on something we're trying to deprecate seems like the bigger factor to consider.

cliffhall: Removed.

src/everything/transports/sse.ts
cliffhall: I agree, omitting things that we're trying to deprecate should be a difficult to overcome force. But there is the matter of the Inspector. It still supports SSE because people may be testing older servers. 

So, if we're to keep the SSE support in the Inspector for now, it's useful for us to have this reliable server counterpart to sanity check the its behavior with. 

If we remove SSE support from the Inspector in the future, then we have much less of a case for keeping it here either.

src/everything/tools/trigger-elicitation-request.ts
cliffhall: This one's fate is also tied up in the Inspector. The Inspector needs to support it because it's actually supported in the spec, and servers may still throw this enum shape at us there. Including it here helps us verify that the Inspector is behaving correctly when it gets a request with that enum shape in it. At least `legacyTitledEnum` has a repellent built into the name.

src/everything/docs/instructions.md
cliffhall: Fixed.

cliffhall:
> This [discussion question ](https://github.com/orgs/modelcontextprotocol/discussions/648)made me think we could add something like [this](https://gist.github.com/olaservo/585b1ad042723a824b6a0e0e02ae60a8), to demonstrate a collection of resources?

@olaservo This sounds good. I also want to add icons to everything that supports them. Maybe we can do this as separate PRs?

UPDATE: I added [this issue](https://github.com/modelcontextprotocol/servers/issues/3130) for the resources collection example and assigned to both of us, just so it doesn't get lost in the wash.

src/everything/prompts/args.ts
```diff
@@ -0,0 +1,41 @@
...
+export const registerArgumentsPrompt = (server: McpServer) => {
+  // Prompt arguments
+  const promptArgsSchema = {
+    city: z.string().describe("Name of the city"),
+    state: z.string().describe("Name of the state").optional(),
+  };
+
+  // Register the prompt
+  server.registerPrompt(
+    "args-prompt",
```
domdomegg, line 20: nit: The convention I've generally seen for prompts and tool names is `snake_case` - is there a reason we're using kebab case? Seems good to follow conventions where we can in case agents use this to see how to implement particular features etc. and pick up on patterns.

src/everything/AGENTS.md
```diff
@@ -0,0 +1,52 @@
...
+- Use camelCase for variables/functions,
+- Use PascalCase for types/classes,
+- Use UPPER_CASE for constants
+- Use kebab-case for file names and registered tools, prompts, and resources.
+- Use verbs for tool names, e.g., `get-annotated-message` instead of `annotated-message`
+
+## Extending the Server
+
+The Everything Server is designed to be extended at well-defined points.
+See [Extension Points](docs/extension.md) and [Project Structure](docsstructure.md).
```
domdomegg:
```suggestion
See [Extension Points](docs/extension.md) and [Project Structure](docs/structure.md).
```

src/everything/resources/templates.ts
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * @param {URL} uri - The URI of the resource to be parsed.
+ * @param {Record<string, unknown>} variables - A record containing context-specific variables that include the resourceId.
+ * @returns {number} The parsed and validated resource identifier as an integer.
+ * @throws {Error} Throws an error if the URI matches unsupported base URIs or if the resourceId is invalid.
+ */
+const parseResourceId = (uri: URL, variables: Record<string, unknown>) => {
+  const uriError = `Unknown resource: ${uri.toString()}`;
+  if (
+    uri.toString().startsWith(textUriBase) &&
```
domdomegg, line 142:
```suggestion
    uri.toString().startsWith(textUriBase) ||
```

I think this is the intention?

domdomegg (dismissed):
Generally LGTM, very excited about this! Thank you :)

src/everything/prompts/args.ts
cliffhall: The `snake_case` convention comes mainly from Python server development, where it matches Python idioms. The point of deviating here is that it illustrates that snake is a _convention_, but tool names can be whatever you want them to be within the allowed parameters. 

If you go by convention on the hopes of appeasing models that somehow respond to `snake_case` better, you to cede the DevX benefit entirely to Python. You reinforce a trend that's already self-abetted by LLMs spitting out Python by default when someone asks for an MCP server. If you mix case in your app, with `kebab-case` files and `snake_case` primitive names, it will inevitably lead to a mix of cases in your filesystem where either case is admitted.

A `kebab-case` primitive can line up with thev TypeScript file name that registers it. All the benefit of having a single case pattern for naming at the runtime and filesystem levels, just with TS friendly idioms. 

It's a design choice one can make. Making it, particularly in this instructive resource, could end up helping LLMs not to overfit on Python idioms.

olaservo (dismissed):
This looks good to me - I think we can merge this and start adding the other couple of examples we've mentioned.

I also did one more claude-driven Playwright test run for all the features for each transport using inspector v 0.18.0, along with manual testing for stdio and didn't run into any errors.

olaservo (approved):

src/everything/prompts/args.ts
PederHP: I recall looking at evals showing `snake_case` and PEP 8 style guidelines performed better. Prompts aren't a model-facing capability, so it doesn't make too much of a difference here.

But you're right that it's a self-reinforcing convention. I sadly think we've passed the point where it's difficult to fight (the C# MCP SDK converts from PascalCase to snake_case when using reflection-based convenience, last I checked, unless overriden, for this very reason). Models simply do better with Python-style tools names.

I am generally in favor of using idiomatic names for the language a server is written, however, and then doing any mapping in a library or even on the client side. Because some models might do better with kebab-case. A model which is using code mode using TypeScript might prefer kebab-case.

PR #3156, opened 2025-12-27
fix(git): use git CLI in git_add to avoid GitPython index corruption bug

schuyler:
## Description

GitPython's index.add() has a known bug that corrupts the git index when paths contain ./ prefixes or when . is included in the file list. The bug causes:
- Files appearing with ./ prefix (e.g., ./index.md)
- The .git directory itself being staged
- Corrupted commits requiring git filter-repo to fix

## Motivation and Context

The previous implementation only handled the exact case of ["."] but failed to handle [".", "other.md"] or ["./README.md"]. This fix uses repo.git.add() (CLI wrapper) for all cases, which is consistent with how other git operations are implemented in this file and correctly handles all edge cases.

## Testing

I have been using the `mcp-server-git` MCP to allow Claude to edit a private [otterwiki](https://otterwiki.com). Prior to making this change locally, I was seeing corruption of the Git repository every few commits. Since applying the change about a week ago,  I have been using the wiki with Claude every single day and have not seen a single error of any kind.

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

## References

- https://github.com/gitpython-developers/GitPython/issues/375
- https://github.com/gitpython-developers/GitPython/issues/292

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -129,10 +129,11 @@ def git_commit(repo: git.Repo, message: str) -> str:
     return f"Changes committed successfully with hash {commit.hexsha}"
 
 def git_add(repo: git.Repo, files: list[str]) -> str:
-    if files == ["."]:
-        repo.git.add(".")
-    else:
-        repo.index.add(files)
+    # Always use git CLI to avoid GitPython index.add() bugs with:
+    # - "." (stages .git/ directory, adds ./ prefix to paths)
+    # - paths with "./" prefix (preserved incorrectly in index)
+    # See: https://github.com/gitpython-developers/GitPython/issues/375
+    repo.git.add(files)
     return "Files staged successfully"
 
 def git_reset(repo: git.Repo) -> str:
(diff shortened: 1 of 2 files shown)


olaservo:
Thanks for catching this and for the great reproduction!

The GitPython `index.add()` bug you describe is real, but the core fix has already landed. Commit [db96050800](https://github.com/modelcontextprotocol/servers/commit/db96050800) ("git: improve file path validation in add operation") on 2025-12-29 — two days after you opened this PR — replaced `repo.index.add(files)` with `repo.git.add("--", *files)`. Both branches of `git_add` now shell out to the git CLI, so GitPython issue #375 no longer applies.

Unfortunately, the change proposed here (`repo.git.add(files)`) would drop the `--` separator, which we rely on as defense-in-depth against flag injection — see also the related hardening in commit [ae40ec239d](https://github.com/modelcontextprotocol/servers/commit/ae40ec239d) for the other git tools. So we can't merge the source change as-is.

Your two new tests (`test_git_add_with_dot_prefix` and `test_git_add_mixed_dot_and_files`) are genuinely useful regression coverage though — both should pass on current main. Would you be willing to open a small follow-up PR with just those tests? That would guard against anyone reverting to `index.add` in the future.

Closing on that basis — thanks again for the thorough investigation!

PR #3160, opened 2025-12-28
fix(memory): return connected relations in openNodes

### LINKED ISSUE #3137: Memory MCP `open_nodes` returns empty relations array despite entity having relations in graph
**Describe the bug**

The `open_nodes` tool returns an empty `relations` array even when the requested entity has outgoing relations in the knowledge graph. This makes it impossible to traverse the graph from a specific node without calling `read_graph` and filtering the entire dataset client-side.

**To Reproduce**

1. Create two entities and a relation between them:
```javascript
create_entities({
  entities: [
    { name: "2025-12-17", entityType: "work-day", observations: [] },
    { name: "incident/example", entityType: "incident", observations: ["Test incident"] }
  ]
})

create_relations({
  relations: [{
    from: "2025-12-17",
    to: "incident/example",
    relationType: "worked-on"
  }]
})
```

2. Call `open_nodes` on the source entity:
```javascript
open_nodes({ names: ["2025-12-17"] })
```

3. Observe the result has an empty `relations` array

**Expected behavior**

`open_nodes` should return outgoing relations from the requested entity, similar to how graph databases return edges when querying a node. The response should include:

```json
{
  "entities": [
    {
      "type": "entity",
      "name": "2025-12-17",
      "entityType": "work-day",
      "observations": []
    }
  ],
  "relations": [
    {
      "type": "relation",
      "from": "2025-12-17",
      "to": "incident/example",
      "relationType": "worked-on"
    }
  ]
}
```

**Logs**

Actual response from `open_nodes`:
```json
{
  "entities": [
    {
      "type": "entity",
      "name": "2025-12-17",
      "entityType": "work-day",
      "observations": []
    }
  ],
  "relations": []
}
```

Calling `read_graph` confirms the relation exists in the graph.

**Additional context**

This limitation forces users to call `read_graph` (which returns the entire knowledge graph) and filter client-side to find relations for specific entities. This doesn't scale as the graph grows.

Use case: Daily work tracking where date entities link to topic entities via `worked-on` relations. To generate a daily summary for a specific date, we need to find all topics linked from that date, but `open_nodes` doesn't provide this information.

Suggested solutions:
1. Include both incoming and outgoing relations in `open_nodes` response by default
2. Add optional parameters to `open_nodes` to control relation inclusion (e.g., `includeRelations: "both" | "outgoing" | "incoming" | "none"`, `relationDepth: number`)

Note: Both directions are useful for different queries. Outgoing relations answer "what does this entity link to?" (e.g., which topics were worked on this date?), while incoming relations answer "what links to this entity?" (e.g., which dates was this incident worked on?). Most graph databases provide both by default.

majiayu000:
## Summary
- `openNodes` now returns relations where **either** endpoint matches requested entities
- Previously required **both** endpoints to match, making graph traversal impossible
- Added tests for the fix including the exact scenario from issue #3137

Fixes #3137

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -336,16 +336,39 @@ describe('KnowledgeGraphManager', () => {
       expect(result.entities.map(e => e.name)).toContain('Bob');
     });
 
-    it('should include relations between opened nodes', async () => {
+    it('should include relations connected to opened nodes', async () => {
       const result = await manager.openNodes(['Alice', 'Bob']);
-      expect(result.relations).toHaveLength(1);
-      expect(result.relations[0].from).toBe('Alice');
-      expect(result.relations[0].to).toBe('Bob');
+      // Returns both Alice->Bob (between opened nodes) and Bob->Charlie (Bob is opened)
+      expect(result.relations).toHaveLength(2);
+      expect(result.relations.some(r => r.from === 'Alice' && r.to === 'Bob')).toBe(true);
+      expect(result.relations.some(r => r.from === 'Bob' && r.to === 'Charlie')).toBe(true);
     });
 
-    it('should exclude relations to unopened nodes', async () => {
+    it('should include both incoming and outgoing relations for single node', async () => {
+      // When opening Bob, should return both incoming (Alice->Bob) and outgoing (Bob->Charlie) relations
       const result = await manager.openNodes(['Bob']);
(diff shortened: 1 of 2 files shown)


olaservo:
Thanks for working on this @majiayu000! You correctly identified the root cause (the `&&` filter in `openNodes`).

Closing this in favor of PR #3297 which was just merged — it includes the same fix for `openNodes` plus also fixes the same bug in `searchNodes`. Appreciate the contribution!

PR #3149, opened 2025-12-23
fix(memory): add 'type' to schemas and use passthrough for Gemini CLI…

ericnjurio:
Title
fix(memory): enhance schema compatibility for strict JSON validators (like Gemini CLI)

Description
This PR updates the memory server's Zod schemas to improve compatibility with MCP clients that enforce strict JSON schema validation, such as the Google Gemini CLI. 

The changes include:
 1. Formally declaring the optional type property in EntitySchema and RelationSchema.
 2. Enabling .passthrough() on these schemas to allow for internal or future properties without breaking strict validation.

Server Details
 - Server: memory
 - Changes to: tools (Output schemas for read_graph, search_nodes, and open_nodes)

Motivation and Context
Strict MCP clients like the Gemini CLI perform rigorous validation on tool outputs. The memory server includes a type field in its internal JSONL storage which is subsequently returned in tool outputs. Since this field was not declared in the tool's output schema, Gemini CLI rejected the response with an "additional properties not allowed" error (code -32602).

By declaring the type field and using .passthrough(), we ensure the server remains robust and compatible with both permissive and strict clients.

How Has This Been Tested?
Tested locally using google-gemini-cli:
 - Verified that read_graph no longer triggers validation errors and returns the full graph.
 - Verified that search_nodes and open_nodes successfully return structured content.
 - Confirmed that data remains intact in the memory.jsonl file.

Breaking Changes
 - No. This change is additive and increases tolerance for existing data structures.

Types of changes
 - [x] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change)
 - [ ] Documentation update

Checklist
 - [x] I have read the [MCP Protocol Documentation (https://modelcontextprotocol.io)
 - [x] My changes follows MCP security best practices
 - [x] I have updated the server's README accordingly (N/A - no user-facing config changes)
 - [x] I have tested this with an LLM client
 - [x] My code follows the repository's style guidelines
 - [ ] New and existing tests pass locally (Need to run npm test)
 - [x] I have added appropriate error handling
 - [x] I have documented all environment variables and configuration options (N/A)

Additional context
The type discriminator is essential for the server's internal logic when parsing the JSONL file. Formalizing its presence in the communication schema resolves the "hidden property" conflict with strict validators.

PR #3161, opened 2025-12-28
fix(memory): allow additional properties in entity/relation schemas

### LINKED ISSUE #3144: Memory MCP  Schema Validation Error on read_graph for entities with additional properties
### Describe the bug

The `read_graph` tool fails when it tries to read a knowledge graph from a `memory.jsonl` file that contains an entity with properties not defined in the official schema (`name`, `entityType`, `observations`). This creates an inconsistency where the server can store data that the client-side validation for `read_graph` will later reject, making it impossible to retrieve the stored memory.

### To Reproduce

1.  Start the `@modelcontextprotocol/server-memory`.
2.  Ensure the `memory.jsonl` file contains at least one entity object with an extra, non-standard property. For example:
    ```json
    {"name": "Test_Entity", "entityType": "test", "observations": ["Has an extra field"], "custom_id": "xyz-123"}
    ```
3.  Use any MCP client (e.g., the VS Code extension) to execute the `read_graph` tool.
4.  Observe that the tool execution fails.

### Expected behavior

The `read_graph` tool should successfully execute without a fatal error. It should handle entities with extra properties gracefully, ideally by ignoring the non-standard properties and returning only the schema-defined fields (`name`, `entityType`, `observations`) for each entity. The presence of extra data in the storage file should not prevent the core functionality of reading the graph.

### Logs

The following error is returned by the client when the bug occurs:

```
Error: McpError: MCP error -32602: Structured content does not match the tool's output schema: data/entities/0 must NOT have additional properties
```

### Additional context

This bug highlights a discrepancy between the data persistence logic and the data retrieval/validation logic. To resolve this, I suggest two potential approaches:

*   **Make the schema more flexible:** Allow additional properties in the client-side schema validation for `read_graph`. This would make the tool more robust.
*   **Enforce strictness on write:** Actively strip or reject any non-standard properties during entity creation (`create_entity`) to ensure the integrity of the `memory.jsonl` file.

Option 1 seems preferable as it adds flexibility and prevents data retrieval failures.

majiayu000:
## Summary
- Added `.passthrough()` to `EntitySchema` and `RelationSchema`
- Entities with extra properties in storage can now be read without validation errors
- Added test case for the exact scenario described in issue #3144

## Problem
When `memory.jsonl` contains entities with extra properties (e.g., `custom_id`), the `read_graph` tool fails with:
```
MCP error -32602: Structured content does not match the tool's output schema: 
data/entities/0 must NOT have additional properties
```

## Solution
Using Zod's `.passthrough()` allows the schema to accept additional properties while still validating required fields.

Fixes #3144

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -390,5 +390,25 @@ describe('KnowledgeGraphManager', () => {
       expect(JSON.parse(lines[0])).toHaveProperty('type', 'entity');
       expect(JSON.parse(lines[1])).toHaveProperty('type', 'relation');
     });
+
+    it('should handle entities with additional properties in storage (issue #3144)', async () => {
+      // Manually write an entity with extra properties to the file
+      const entityWithExtra = JSON.stringify({
+        type: 'entity',
+        name: 'Test_Entity',
+        entityType: 'test',
+        observations: ['Has an extra field'],
+        custom_id: 'xyz-123',  // Extra property not in schema
+        metadata: { source: 'external' }  // Another extra property
+      });
+      await fs.writeFile(testFilePath, entityWithExtra);
+
+      // Reading the graph should not throw an error
+      const graph = await manager.readGraph();
+      expect(graph.entities).toHaveLength(1);
+      expect(graph.entities[0].name).toBe('Test_Entity');
+      expect(graph.entities[0].entityType).toBe('test');
+      expect(graph.entities[0].observations).toContain('Has an extra field');
+    });
   });
 });
--- a/src/memory/index.ts
+++ b/src/memory/index.ts
@@ -224,17 +224,19 @@ export class KnowledgeGraphManager {
 let knowledgeGraphManager: KnowledgeGraphManager;
 
 // Zod schemas for entities and relations
+// Using passthrough() to allow additional properties that may exist in stored data
+// This prevents schema validation errors when reading entities with extra fields
 const EntitySchema = z.object({
   name: z.string().describe("The name of the entity"),
   entityType: z.string().describe("The type of the entity"),
   observations: z.array(z.string()).describe("An array of observation contents associated with the entity")
-});
+}).passthrough();
 
 const RelationSchema = z.object({
   from: z.string().describe("The name of the entity where the relation starts"),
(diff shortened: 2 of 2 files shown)


NeverMore93:
The issue is not resolved

majiayu000:
@NeverMore93 I've tested this fix and verified it works correctly. Here's my analysis:

**Root Cause:**
The validation error occurs on the **client side**, not the server side. When the server returns `structuredContent`, the MCP client validates it against the `outputSchema` using AJV (JSON Schema validator).

- Without `.passthrough()`: JSON Schema has `additionalProperties: false` → validation fails
- With `.passthrough()`: JSON Schema has `additionalProperties: {}` → validation passes

**Test Results:**
I reproduced the issue and confirmed the fix works:

```
# Before fix (additionalProperties: false)
Error: MCP error -32602: Structured content does not match the tool's output schema: 
data/entities/0 must NOT have additional properties

# After fix (additionalProperties: {})
SUCCESS! read_graph returns data with extra properties correctly
```

Could you share more details about your test environment?
- MCP SDK version (`npm list @modelcontextprotocol/sdk`)
- The exact error message you're seeing
- Steps to reproduce the issue

This will help identify if there's a version-specific issue or a different scenario not covered by this fix.

olaservo:
Thanks for the fix and the clear reproduction — this was a real issue!

The underlying bug has already been fixed, just at a different layer. Commit [8af50b9418](https://github.com/modelcontextprotocol/servers/commit/8af50b9418) ("remove type from loading") on 2026-01-14 rewrote `loadGraph` to explicitly pick only `name` / `entityType` / `observations` (and `from` / `to` / `relationType` for relations) when building each item. That strips `custom_id`, `metadata`, and any other extras before the SDK output validator sees them, so #3144 no longer reproduces on main.

Your new test actually passes on current main without the `.passthrough()` change, because the stripping happens in `loadGraph` now. Regression tests for the same guarantee (`should strip type field from entities when loading from file` etc.) shipped with that commit.

Because `.passthrough()` would weaken the output schema — see @domdomegg's pushback on #3149 for the schema-drift reasoning — we'd rather keep the current strict `z.object` defaults now that the bug is fixed at the data layer.

Closing on that basis — thanks for the contribution!

PR #3167, opened 2025-12-31
fix(memory): strip extra properties when loading entities from JSONL

### LINKED ISSUE #3144: Memory MCP  Schema Validation Error on read_graph for entities with additional properties

cardmagic:
## Summary

When loading entities and relations from JSONL files, the parsed JSON objects included extra properties like `type` (used for discriminating entity vs relation) and any other custom fields that may have been manually added or present in legacy data. These extra properties leaked into the Entity/Relation objects, causing MCP schema validation errors (-32602 Invalid params) when outputSchema validation is enabled.

**The fix:** Explicitly extract only the schema-defined properties (name, entityType, observations for entities; from, to, relationType for relations) instead of casting the entire parsed JSON object.

## Test plan

- [x] Added test case that writes JSONL with extra properties and verifies they are stripped when loading
- [x] All 40 existing tests continue to pass

Fixes #3144

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -360,6 +360,40 @@ describe('KnowledgeGraphManager', () => {
     });
   });
 
+  describe('loading entities with extra properties', () => {
+    it('should strip extra properties from entities when loading (issue #3144)', async () => {
+      // Write JSONL file directly with extra properties (simulating manual edits or legacy data)
+      const jsonlContent = [
+        JSON.stringify({ type: "entity", name: "Test", entityType: "test", observations: ["obs1"], custom_id: "xyz-123", extra_field: true }),
+        JSON.stringify({ type: "relation", from: "Test", to: "Test", relationType: "self", extra_prop: "should be stripped" })
+      ].join('\n');
+      await fs.writeFile(testFilePath, jsonlContent);
+
+      const graph = await manager.readGraph();
+
+      // Entities should only have the schema-defined properties
+      expect(graph.entities).toHaveLength(1);
+      expect(graph.entities[0]).toEqual({
+        name: "Test",
+        entityType: "test",
+        observations: ["obs1"]
+      });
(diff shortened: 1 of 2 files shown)


cardmagic:
Closing as duplicate - this is already addressed by #3161

PR #3184, opened 2026-01-05
Add workflow to handle README-only PRs

domdomegg:
Adds a GitHub Actions workflow that handles PRs which only modify README.md:

- Comments explaining we no longer accept new server additions to the README
- Directs contributors to the MCP registry instead
- Allows legitimate updates/removals via `/i-promise-this-is-not-a-new-server` command
- Uses labels (`readme: pending` / `readme: ready for review`) for filtering
- Minimizes comments after confirmation to reduce clutter

This should reduce the manual effort of closing PRs that are adding new servers to the deprecated list.

### DIFF
--- a/.github/workflows/readme-pr-check.yml
+++ b/.github/workflows/readme-pr-check.yml
@@ -1,0 +1,109 @@
+name: README PR Check
+
+on:
+  pull_request:
+    types: [opened]
+    paths:
+      - 'README.md'
+  issue_comment:
+    types: [created]
+
+jobs:
+  check-readme-only:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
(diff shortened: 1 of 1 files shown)


domdomegg:
(I've looked over the implementation, and gone back and forth with Claude a bit with it - I _think_ it's right but hard to tell without actually running it in GitHub Actions so can't be certain)

olaservo (approved):
Thank you for adding this, lets try it out!

olaservo:
@domdomegg showing a syntax error: https://github.com/modelcontextprotocol/servers/actions/runs/20738556416/workflow

My test PR: https://github.com/modelcontextprotocol/servers/pull/3186

domdomegg:
Fixed, thanks! New test in https://github.com/modelcontextprotocol/servers/pull/3188

Issue #2925, opened 2025-10-29
Add tests for Everything Server


olaservo:
**Is your feature request related to a problem? Please describe.**

Edit, see: https://github.com/modelcontextprotocol/servers/issues/3080 since I think this will be a prerequisite to making testing possible.

Everything Server is arguably the most important reference server here but it still needs test coverage.  I've been saving it for last since it might need refactoring and others are actively making changes to it.

**Describe the solution you'd like**

Vitest coverage

**Describe alternatives you've considered**

Relying on manual testing

**Additional context**

PR #3236, opened 2026-01-20
Add Vitest tests for Everything Server

nielskaspers:
## Summary

Adds comprehensive Vitest test coverage for the Everything Server, addressing #2925.

- **124 tests** across tools, prompts, resources, and server
- **71% overall coverage** (was 0%)
- All 16 tools, all 4 prompts tested

### Coverage by Component

| Component | Coverage | Tests |
|-----------|----------|-------|
| Tools | 93% | 61 |
| Prompts | 91% | 18 |
| Resources | 65% | 31 |
| Server | 63% | 8 |
| Registrations | 100% | 6 |

### Tools (16/16 tested)
All tools at 100% except:
- `get-roots-list` (35%) - capability-gated branches
- `gzip-file-as-resource` (84%) - network error branches

### Not Unit Tested
Transport files (`stdio.ts`, `sse.ts`, `streamableHttp.ts`) are executable entry points that start Express servers. These require integration tests rather than unit tests.

## Test Files
```
__tests__/tools.test.ts        - 61 tests (all 16 tools)
__tests__/prompts.test.ts      - 18 tests (all 4 prompts)
__tests__/resources.test.ts    - 31 tests (templates, session, files, subscriptions)
__tests__/server.test.ts       - 8 tests (createServer factory)
__tests__/registrations.test.ts - 6 tests (index wiring files)
vitest.config.ts               - test configuration
```

## Test Plan
- [x] All 124 tests pass locally
- [x] Coverage report generated with v8
- [x] Mock server pattern matches existing test patterns in filesystem/memory servers

### DIFF
--- a/src/everything/__tests__/registrations.test.ts
+++ b/src/everything/__tests__/registrations.test.ts
@@ -48,7 +48,7 @@ describe('Registration Index Files', () => {
     it('should register conditional tools based on capabilities', async () => {
       const { registerConditionalTools } = await import('../tools/index.js');
 
-      // Server with all capabilities
+      // Server with all capabilities including experimental tasks API
       const mockServerWithCapabilities = {
         registerTool: vi.fn(),
         server: {
@@ -58,11 +58,16 @@ describe('Registration Index Files', () => {
             sampling: {},
           })),
         },
+        experimental: {
+          tasks: {
+            registerToolTask: vi.fn(),
+          },
+        },
       } as unknown as McpServer;
 
       registerConditionalTools(mockServerWithCapabilities);
 
-      // Should register 3 conditional tools when all capabilities present
+      // Should register 3 conditional tools + 3 task-based tools when all capabilities present
       expect(mockServerWithCapabilities.registerTool).toHaveBeenCalledTimes(3);
 
       const registeredTools = (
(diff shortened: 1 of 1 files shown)


olaservo:
@claude could you please sanity check these new tests against the MCP specification to ensure that we are testing the correct behavior?

EDIT: I keep forgetting that Claude can't access forks.

nielskaspers (commented):
Hi @olaservo! Thank you for the thorough review — you're absolutely right that those tests were adding noise rather than meaningful coverage. I really appreciate you taking the time to identify them so specifically.

I've addressed all your feedback:

**Removed (29 tests):**
- ✅ Registration boilerplate (16) — redundant with `registrations.test.ts`
- ✅ Redundant role/type checks (3) — consolidated assertions into existing behavioral tests
- ✅ "Should not throw" tests (6) — merged into a single lifecycle test
- ✅ Constant identity tests (2) — agreed, these provided no safety net
- ✅ `expect(true).toBe(true)` (1) — replaced with actual assertion that `registerResource` was called
- ✅ Weak capability test (1) — removed, the `oninitialized` handler check already exists

**Strengthened:**
- Resource templates test now verifies specific resource names (`Dynamic Text Resource`, `Dynamic Blob Resource`)
- File resources test now asserts `registerResource` was called

Test count went from 124 → 95, and all tests pass. Coverage remains at ~71%.

Let me know if there's anything else you'd like me to adjust!

Issue #3106, opened 2025-12-05
In the filesystem, the outputSchema type of the directory_tree tool is incorrect


zrq1523565:
**Describe the bug**
the outputSchema type of the directory_tree tool is incorrect

**To Reproduce**
call the directory_tree tool in the client

**Expected behavior**
return the correct information

**Logs**
```
meta=None content=[TextContent(type='text', text='MCP error -32602: Output validation error: Invalid structured content for tool directory_tree: [\n  {\n    "expected": "string",\n    "code": "invalid_type",\n    "path": [\n      "content"\n    ],\n    "message": "Invalid input: expected string, received array"\n  }\n]', annotations=None, meta=None)] structuredContent=None isError=True
```

**Additional context**
```
server.registerTool("directory_tree", {
...
    outputSchema: {
        content: z.array(z.object({
            type: z.string(),
            text: z.string()
        }))
    },
...
```  
after making the above modifications in ./@modelcontextprotocol/server-filesystem/dist/index.js, the returned data is correct  
  
correct logs:
```
[
  {
    "name": "kafka",
    "type": "directory",
    "children": [
      {
        "name": "mykafka",
        "type": "directory",
        "children": [
          {
            "name": ".git",
            "type": "directory",
            "children": [
              {
                "name": "COMMIT_EDITMSG",
                "type": "file"
              },
              {
                "name": "FETCH_HEAD",
                "type": "file"
              },
              {
                "name": "HEAD",
                "type": "file"
              },
              {
                "name": "branches",
                "type": "directory",
                "children": []
              },
...
```

coryvirok:
Related: https://github.com/modelcontextprotocol/servers/issues/3093#issuecomment-3657583423

lopopolo-openai:
Thanks for the report — I ran into the same symptom while testing older `@modelcontextprotocol/server-filesystem` builds.

It looks like this was already fixed on `main` in commit `e6933ca` ("Fix: Changed structuredContent output to match outputSchema" / #3099) on 2025-12-11. Based on npm publish dates, the fix appears to be included in releases published after that date (e.g. `@modelcontextprotocol/server-filesystem@2025.12.18` and later).

Related context: I also opened #3245 for regression test coverage and initially missed that `main` already contained the fix (similar to what happened in #3166).

If that understanding is correct, could maintainers close this issue as already fixed upstream? Happy to adjust this note if I’m missing any nuance.

Note: this comment written by OpenAI Codex.

[...]

Issue #3093, opened 2025-11-30
Filesystem MCP: move_file tool returns array instead of string (Error -32602)


shintaro0403:
Filesystem MCP: move_file tool returns array instead of string (Error -32602)

Summary

When executing the move_file tool from @modelcontextprotocol/server-filesystem, error code -32602 (Invalid params) occurs.

Root Cause: The tool returns an array instead of a string, causing MCP protocol validation to fail.

Error Details

Error Code: -32602 (JSON-RPC Invalid params)

Error Message:
Output validation error: Invalid structured content for tool move_file: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": ["content"],
    "message": "Invalid input: expected string, received array"
  }
]

Problem Location: The content field is an array (should be a string)

Steps to Reproduce

1. Configure @modelcontextprotocol/server-filesystem as an MCP server
2. Call the move_file tool with:
   - source: Source file path
   - destination: Destination file path
3. Error occurs

Example:
{
  "source": "C:\\path\\to\\file.txt",
  "destination": "C:\\path\\to\\new\\file.txt"
}

Expected Behavior

The move_file tool should return a string message when file movement succeeds.

Example:
"Successfully moved file from source to destination"

Actual Behavior

The move_file tool returns an array, causing MCP protocol validation to fail.

Environment

OS: Windows 10 (build 26100)
Node.js: 22.21.1 (Built-in Claude Desktop)
Package: @modelcontextprotocol/server-filesystem@latest
MCP Client: Claude Desktop 1.0.1307

Log Excerpts

Request:
2025-11-30T09:57:51.155Z [filesystem] [info] Message from client: 
{
  "method": "tools/call",
  "params": {
    "name": "move_file",
    "arguments": {
      "source": "C:\\Users\\skiku\\Documents\\MCP_Learning\\chapter01\\test1.txt",
      "destination": "C:\\Users\\skiku\\Documents\\MCP_Learning\\chapter01\\テストファイル\\test1.txt"
    }
  },
  "jsonrpc": "2.0",
  "id": 26
}

Response (Error):
2025-11-30T09:57:51.174Z [filesystem] [info] Message from server: 
{
  "jsonrpc": "2.0",
  "id": 26,
  "result": {
    "content": [{
      "type": "text",
      "text": "MCP error -32602: Output validation error: Invalid structured content for tool move_file: ..."
    }],
    "isError": true
  }
}

Additional Information

[...]

PR #3245, opened 2026-01-24
test(filesystem): add MCP SDK regression coverage for directory_tree

lopopolo-openai:
## Summary

This PR **adds regression test coverage only** for a `directory_tree` structured-content validation bug that appears to have already been fixed upstream.

There are **no runtime behavior changes** in this PR.

## Context

The `-32602` error reported in:

- https://github.com/modelcontextprotocol/servers/issues/3106
- https://github.com/modelcontextprotocol/servers/issues/3093

occurs when a tool declares `outputSchema: { content: string }` but returns `structuredContent.content` as an **array**.

From git history, this appears to have been fixed in:

- `e6933ca` — "Fix: Changed structuredContent output to match outputSchema (#3099)" (2025-12-11)

Note: I opened this PR after seeing the open bug reports and reproducing the issue in an older build, but I did not initially confirm that `main` already contained the fix. I’m keeping this PR scoped to test coverage in case it’s still useful as a regression guard.

## What this PR does

- Adds `src/filesystem/__tests__/directory-tree.mcp-sdk.test.ts`
  - Uses `@modelcontextprotocol/sdk` + `StdioClientTransport` to call `directory_tree` end-to-end.
  - Asserts that `structuredContent.content` is a `string` and JSON-parseable.
- Makes `structured-content.test.ts` slightly more robust on macOS by resolving the temp directory with `fs.realpath()` so the allowlist path matches `/private/...` temp paths.

This is intended to help prevent reintroducing the same outputSchema/structuredContent mismatch.

## Minimal reproduction snippet (for older builds)

I was able to reproduce the `-32602` validation error by running the filesystem server over stdio and calling `directory_tree` through the MCP SDK client.

Maintainers can run the following snippet directly (Node 20+):

```ts
import { Client } from '@modelcontextprotocol/sdk/client/index.js';
import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
import fs from 'node:fs/promises';
import os from 'node:os';
import path from 'node:path';

const tmp = await fs.mkdtemp(path.join(os.tmpdir(), 'mcp-fs-tree-'));
const root = await fs.realpath(tmp);
await fs.writeFile(path.join(root, 'a.txt'), 'hello');

const transport = new StdioClientTransport({
  command: 'npx',
  args: ['-y', '@modelcontextprotocol/server-filesystem', root],
});

const client = new Client({ name: 'debug-client', version: '1.0.0' }, { capabilities: {} });
await client.connect(transport);

const result = await client.callTool({
  name: 'directory_tree',
  arguments: { path: root },
});

console.log(result.structuredContent);
await client.close();
```

On buggy builds, this call throws `McpError: -32602 ... Expected string, received array at path: ["content"]`.

## Release context

I have not traced the exact release tag for `e6933ca`, but based on timestamps it looks like the fix (2025-12-11) should be included in npm releases published **after that date** (for example `@modelcontextprotocol/server-filesystem@2025.12.18` and later).

Happy to update this PR with the exact release tag if you'd like me to track it down precisely.

## Test plan

- `npm test --workspace @modelcontextprotocol/server-filesystem`

## Follow-up suggestion (optional)

If you'd like, I can follow up with a small CI-level "functional" test job that runs a tiny MCP SDK client against the built `dist/index.js` binaries for select servers (filesystem, memory, etc.) to assert that schema validation passes for a few representative tool calls.

I didn't include CI changes in this PR since I wasn't sure how you'd want to scope cross-package functional tests.

## Authorship note

This contribution (bug identification, repro, test updates, and PR drafting) was authored with OpenAI Codex.

### DIFF
--- a/src/filesystem/README.md
+++ b/src/filesystem/README.md
@@ -209,6 +209,55 @@ Add this to your `claude_desktop_config.json`:
 
 Note: you can provide sandboxed directories to the server by mounting them to `/projects`. Adding the `ro` flag will make the directory readonly by the server.
 
+## Troubleshooting
+
+### MCP error -32602: "Invalid structured content for tool directory_tree"
+
+If you see an error like:
+
+```
+MCP error -32602: Output validation error
+Invalid structured content for tool directory_tree
+Expected string, received array at path: ["content"]
+```
+
+it usually means the server returned `structuredContent.content` as an array even
+though the tool's `outputSchema` declares `content` as a string.
+
+You can reproduce a correct end-to-end `directory_tree` call with the MCP SDK using
+the following script:
+
+```ts
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+
+const tmp = await fs.mkdtemp(path.join(os.tmpdir(), 'mcp-fs-tree-'));
+const root = await fs.realpath(tmp);
+await fs.writeFile(path.join(root, 'a.txt'), 'hello');
+
+const transport = new StdioClientTransport({
+  command: 'npx',
+  args: ['-y', '@modelcontextprotocol/server-filesystem', root],
+});
+
+const client = new Client({ name: 'debug-client', version: '1.0.0' }, { capabilities: {} });
+await client.connect(transport);
+
+const result = await client.callTool({
+  name: 'directory_tree',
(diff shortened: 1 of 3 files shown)


cliffhall:
@claude review

olaservo (approved):
Test-only, covers the `-32602` structured-content bug from #3106 / #3093 fixed in e6933ca.

Checked that `__dirname` is already used by `startup-validation.test.ts` and `structured-content.test.ts`, and that spawning `dist/index.js` works in CI because `npm ci` runs `prepare` and builds before `npm test`.

PR #3296, opened 2026-02-07
fix(filesystem): handle Windows EPERM when renaming over locked files

### LINKED ISSUE #3199: filesystem edit_file: EPERM error when renaming over locked files on Windows
**Describe the bug**
The `filesystem_edit_file` MCP tool failed on Windows operating systems with the following error:

**To Reproduce**
1. Create a text file `test.txt` with some content
2. Open the file in a text editor (e.g., VS Code, Notepad) - this adds a file read lock
3. Attempt to run the following Node.js script to overwrite the file:

```javascript
import fs from "fs/promises";
import { randomBytes } from 'crypto';

const filePath = './test.txt';
const tempPath = `${filePath}.${randomBytes(16).toString('hex')}.tmp`;

async function editFile() {
  await fs.writeFile(tempPath, 'updated content', 'utf-8');
  await fs.rename(tempPath, filePath); // Throws EPERM on Windows
}

editFile();
```

4. Observe the `EPERM` error

**Expected behavior**
The tool should successfully overwrite files on Windows, just as it does on Linux and macOS platforms.

**Logs**
Error: EPERM: operation not permitted, rename 'C:\test.txt.abc123.tmp' -> 'C:\test.txt'.

**Additional context**
On Windows, `fs.rename()` cannot overwrite existing files when the file is locked (e.g., opened in an editor), resulting in `EPERM` error. This is a Windows-specific limitation of the Node.js `fs.rename()` function.

**Solution Applied**
Replaced `fs.rename()` with `fs.cp()` which supports the `force: true` option to overwrite existing files:

```javascript
const tempPath = `${filePath}.${randomBytes(16).toString('hex')}.tmp`;
try {
  await fs.writeFile(tempPath, modifiedContent, 'utf-8');
  // Use cp with force option - works on all platforms including Windows
  await fs.cp(tempPath, filePath, { force: true });
} finally {
  try {
    await fs.unlink(tempPath);
  } catch {}
}
```

MattBenesch:
On Windows, fs.rename() fails with EPERM when the target file is locked by another process (e.g., open in an editor). This adds a fallback that uses fs.cp() + fs.unlink() when EPERM is encountered, allowing edits to succeed even when the file is open.

Fixes #3199

## Description
Adds atomicReplaceFile() helper that falls back to fs.cp() when fs.rename() fails with EPERM.

## Publishing Your Server

**Note: We are no longer accepting PRs to add servers to the README.** Instead, please publish your server to the [MCP Server Registry](https://github.com/modelcontextprotocol/registry) to make it discoverable to the MCP ecosystem.

To publish your server, follow the [quickstart guide](https://github.com/modelcontextprotocol/registry/blob/main/docs/modelcontextprotocol-io/quickstart.mdx). You can browse published servers at [https://registry.modelcontextprotocol.io/](https://registry.modelcontextprotocol.io/).

## Server Details
  - Server: filesystem
  - Changes to: tools (edit_file, write_file)

## Motivation and Context
On Windows, files open in editors (VS Code, Notepad, etc.) are locked. The current fs.rename() approach fails with EPERM when trying to atomically replace these locked files. This prevents users from editing files they have open.

## How Has This Been Tested?
Unit tests added for EPERM fallback behavior. All 46 lib.test.ts tests pass.

## Breaking Changes
None - existing behavior unchanged on Linux/macOS. Windows users will now have edits succeed where they previously failed.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -279,10 +279,65 @@ describe('Lib Functions', () => {
     describe('writeFileContent', () => {
       it('writes file content', async () => {
         mockFs.writeFile.mockResolvedValueOnce(undefined);
-        
+
         await writeFileContent('/test/file.txt', 'new content');
-        
+
         expect(mockFs.writeFile).toHaveBeenCalledWith('/test/file.txt', 'new content', { encoding: "utf-8", flag: 'wx' });
+      });
+
+      it('falls back to fs.cp when fs.rename fails with EPERM (Windows locked file)', async () => {
+        // First write fails because file exists
+        const eexistError = new Error('EEXIST') as NodeJS.ErrnoException;
+        eexistError.code = 'EEXIST';
+
+        // Rename fails with EPERM (Windows file lock)
+        const epermError = new Error('EPERM') as NodeJS.ErrnoException;
+        epermError.code = 'EPERM';
+
+        mockFs.writeFile
+          .mockRejectedValueOnce(eexistError)  // First write fails (file exists)
+          .mockResolvedValueOnce(undefined);    // Temp file write succeeds
+        mockFs.rename.mockRejectedValueOnce(epermError);
+        mockFs.cp.mockResolvedValueOnce(undefined);
+        mockFs.unlink.mockResolvedValueOnce(undefined);
+
+        await writeFileContent('/test/file.txt', 'new content');
+
+        // Should have tried rename first, then fallen back to cp + unlink
+        expect(mockFs.rename).toHaveBeenCalledWith(
+          expect.stringMatching(/\/test\/file\.txt\.[a-f0-9]+\.tmp$/),
+          '/test/file.txt'
+        );
+        expect(mockFs.cp).toHaveBeenCalledWith(
+          expect.stringMatching(/\/test\/file\.txt\.[a-f0-9]+\.tmp$/),
+          '/test/file.txt',
+          { force: true }
+        );
+        expect(mockFs.unlink).toHaveBeenCalledWith(
+          expect.stringMatching(/\/test\/file\.txt\.[a-f0-9]+\.tmp$/)
+        );
+      });
+
+      it('propagates non-EPERM errors from fs.rename', async () => {
+        // First write fails because file exists
+        const eexistError = new Error('EEXIST') as NodeJS.ErrnoException;
+        eexistError.code = 'EEXIST';
+
+        // Rename fails with EACCES (not EPERM)
+        const eaccesError = new Error('EACCES') as NodeJS.ErrnoException;
+        eaccesError.code = 'EACCES';
+
+        mockFs.writeFile
+          .mockRejectedValueOnce(eexistError)
+          .mockResolvedValueOnce(undefined);
+        mockFs.rename.mockRejectedValueOnce(eaccesError);
+        mockFs.unlink.mockResolvedValueOnce(undefined); // Cleanup of temp file
+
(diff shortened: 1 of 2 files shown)


olaservo (changes requested):
Thanks for this fix! I reproduced the bug on Windows 11 (Node v22.12.0) and confirmed `fs.rename` fails with EPERM when the target is locked, while the fallback succeeds. Build and all 145 tests pass.

I think you need to make one change to the fallback strategy to preserve the codebase's existing symlink safety guarantees, plus a small robustness fix:

---

### Security: `fs.cp` with `force: true` introduces a TOCTOU window

The existing code comments at both call sites explicitly guard against this:

> Security: Use atomic rename to prevent race conditions where symlinks could be created between validation and write. Rename operations replace the target file atomically and don't follow symlinks.

We verified in the [Node.js source](https://github.com/nodejs/node/blob/main/lib/internal/fs/cp/cp.js#L161-L164) that `fs.cp` with `{ force: true }` calls `unlink(dest)` before `copyFile`. This creates a brief window where the target path is empty and a symlink could be planted before `copyFile` runs. While practical risk is low (Windows-only path, symlinks require admin/Developer Mode, microsecond window), this is a regression from the current code's guarantees.

**Fix: try `fs.writeFile` first.** Writing directly to the locked file overwrites in-place without unlinking — no empty path, no TOCTOU window. We tested both approaches against all four Windows file-sharing flag combinations using PowerShell `[System.IO.File]::Open()`:

| Lock type | `rename` | `writeFile` | `fs.cp` |
|---|---|---|---|
| `READ\|WRITE\|DELETE` (VS Code / Node.js) | EPERM | **OK** | **OK** |
| `READ\|WRITE` (no DELETE) | EPERM | **OK** | **EBUSY** |
| `READ\|DELETE` (no WRITE) | EPERM | EBUSY | **OK** |
| `READ` only | EPERM | EBUSY | EBUSY |

Neither alone covers all cases, but cascading them (`writeFile` first, `fs.cp` as fallback) covers the most scenarios without introducing any new security risk:

```typescript
async function atomicReplaceFile(tempPath: string, targetPath: string): Promise<void> {
  try {
    await fs.rename(tempPath, targetPath);
  } catch (error) {
    if ((error as NodeJS.ErrnoException).code === 'EPERM') {
      // Try direct overwrite first — no unlink means no TOCTOU symlink window.
      // Works when the locker grants FILE_SHARE_WRITE (most common case).
      try {
        const content = await fs.readFile(tempPath);
        await fs.writeFile(targetPath, content);
      } catch {
        // Fall back to fs.cp for the uncommon case where the locker grants
        // FILE_SHARE_DELETE but not WRITE. fs.cp calls unlink(dest) before
        // copyFile, which has a brief TOCTOU window, but this is acceptable:
        // Windows-only, symlinks need admin, and the window is microseconds.
        await fs.cp(tempPath, targetPath, { force: true });
      }
      // Best-effort temp cleanup — don't fail a successful write over this
      try {
        await fs.unlink(tempPath);
      } catch {
        // Target was already written successfully
      }
    } else {
      throw error;
    }
  }
}
```

**Why this ordering is safe:** In the common scenarios (VS Code, most editors), `writeFile` succeeds with zero TOCTOU window. The `fs.cp` fallback only runs in the uncommon `FILE_SHARE_DELETE`-without-`WRITE` case. So the cascading approach is strictly equal or better than `fs.cp` alone in every scenario.

---

(This review was assisted by Claude and  edited by me)

cliffhall:
@claude review

LuuOW (commented):

PR #3332, opened 2026-02-14
fix: include outgoing relations in open_nodes

Limitless2023:
## Summary

Fix for issue #3137 - open_nodes returns empty relations array even when entities have outgoing relations.

## Changes

- Modified openNodes to return outgoing relations (where the entity is the from endpoint) in addition to relations between opened nodes
- Updated tool description to reflect new behavior
- Updated tests to match new expected behavior

## Why

Previously, open_nodes only returned relations where both endpoints were in the requested node list. This made it impossible to traverse the graph from specific nodes without reading the entire graph.

Now, when you open a node, you also get its outgoing relations, enabling graph traversal.

## Test

All tests pass (42/42)

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -336,16 +336,19 @@ describe('KnowledgeGraphManager', () => {
       expect(result.entities.map(e => e.name)).toContain('Bob');
     });
 
-    it('should include relations between opened nodes', async () => {
+    it('should include relations between opened nodes and outgoing relations', async () => {
       const result = await manager.openNodes(['Alice', 'Bob']);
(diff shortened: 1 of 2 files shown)


BossChaos:
## Code Review

**PR:** [fix: include outgoing relations in open_nodes](https://github.com/modelcontextprotocol/servers/pull/3332) by @Limitless2023

- ✅ Bug/security fix

*Code review*

LuuOW (commented):

olaservo:
This was fixed on `main` in `ca7ea22` (fix(memory): return relations connected to requested nodes in openNodes/searchNodes), which includes relations where either endpoint is in the requested set, so it covers the outgoing case here as well as incoming. #3137 was closed on that basis. Closing as superseded.

PR #3205, opened 2026-01-12
Fix windows absolute path parsing bug

Shellishack:
## Description
Fix Windows absolute path parsing bug described in https://github.com/modelcontextprotocol/servers/issues/3174

## Publishing Your Server

**Note: We are no longer accepting PRs to add servers to the README.** Instead, please publish your server to the [MCP Server Registry](https://github.com/modelcontextprotocol/registry) to make it discoverable to the MCP ecosystem.

To publish your server, follow the [quickstart guide](https://github.com/modelcontextprotocol/registry/blob/main/docs/modelcontextprotocol-io/quickstart.mdx). You can browse published servers at [https://registry.modelcontextprotocol.io/](https://registry.modelcontextprotocol.io/).

## Server Details

- Server: filesystem
- Changes to: roots

## Motivation and Context

Fix Windows path parsing bug where any path isn't recognized as valid path except ones starting with `./`

## How Has This Been Tested?

Manully tested on Windows and Linux.

## Breaking Changes

No

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [x] My changes follows MCP security best practices
- [ ] I have updated the server's README accordingly
- [ ] I have tested this with an LLM client
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have documented all environment variables and configuration options

PR #3353, opened 2026-02-19
fix(filesystem): fix quote style and add fileURLToPath regression test

olaservo:
## Summary
Follow-up to #3205 with two small improvements:

- **Fix import quote style**: Changed `import { fileURLToPath } from "url"` to use single quotes, matching the rest of the file
- **Add regression test**: New test using `pathToFileURL()` to verify properly-encoded `file://` URIs (the format VS Code sends via MCP roots) are handled correctly on all platforms

## Test plan
- [x] `npx vitest run src/filesystem/__tests__/roots-utils.test.ts` — all 4 tests pass (3 existing + 1 new)

🦉 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/filesystem/__tests__/roots-utils.test.ts
+++ b/src/filesystem/__tests__/roots-utils.test.ts
@@ -3,6 +3,7 @@ import { getValidRootDirectories } from '../roots-utils.js';
 import { mkdtempSync, rmSync, mkdirSync, writeFileSync, realpathSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
+import { pathToFileURL } from 'url';
 import type { Root } from '@modelcontextprotocol/sdk/types.js';
 
 describe('getValidRootDirectories', () => {
@@ -43,6 +44,21 @@ describe('getValidRootDirectories', () => {
       expect(result).toContain(testDir2);
       expect(result).toContain(testDir3);
       expect(result).toHaveLength(3);
+    });
+
+    it('should handle standard file URIs from pathToFileURL', async () => {
+      // pathToFileURL produces properly-encoded URIs (e.g. file:///C:/... on Windows)
+      // This is the format VS Code and other editors send via MCP roots
+      const roots: Root[] = [
+        { uri: pathToFileURL(testDir1).href, name: 'Standard File URI' },
+        { uri: pathToFileURL(testDir2).href, name: 'Standard File URI 2' },
+      ];
+
+      const result = await getValidRootDirectories(roots);
+
+      expect(result).toContain(testDir1);
+      expect(result).toContain(testDir2);
+      expect(result).toHaveLength(2);
     });
 
     it('should normalize complex paths', async () => {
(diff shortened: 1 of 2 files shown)


src/filesystem/__tests__/roots-utils.test.ts
```diff
@@ -45,6 +46,21 @@ describe('getValidRootDirectories', () => {
...
+        { uri: pathToFileURL(testDir1).href, name: 'Standard File URI' },
+        { uri: pathToFileURL(testDir2).href, name: 'Standard File URI 2' },
+      ];
+
+      const result = await getValidRootDirectories(roots);
+
+      expect(result).toContain(testDir1);
+      expect(result).toContain(testDir2);
+      expect(result).toHaveLength(2);
+    });
```
koic, line 62: It appears that the tests pass on non-Windows platforms regardless of the changes in #3205. Since CI also does not seem to run tests on Windows, how about converting this into a regression test that can be reproduced on macOS and other platforms as shown below?

```suggestion
    it('should handle standard file URIs from pathToFileURL', async () => {
      // pathToFileURL encodes special characters (e.g. spaces become %20, # becomes %23).
      // fileURLToPath() decodes them correctly; slice(7) does not.
      // e.g. pathToFileURL('/tmp/my dir').href => 'file:///tmp/my%20dir'
      //   fileURLToPath('file:///tmp/my%20dir') => '/tmp/my dir' (correct)
      //   'file:///tmp/my%20dir'.slice(7)       => '/tmp/my%20dir' (wrong)
      const dirWithSpaces = join(testDir1, 'dir with spaces');
      const dirWithHash = join(testDir1, 'dir#hash');
      mkdirSync(dirWithSpaces);
      mkdirSync(dirWithHash);

      const roots: Root[] = [
        { uri: pathToFileURL(dirWithSpaces).href },
        { uri: pathToFileURL(dirWithHash).href },
      ];

      const result = await getValidRootDirectories(roots);

      expect(result).toContain(dirWithSpaces);
      expect(result).toContain(dirWithHash);
      expect(result).toHaveLength(2);
    });
```

cliffhall:
@claude review

LuuOW (commented):

ChrisKrawczyk:
Hit this from a brand-new MCP host (Obsidian plugin) using `pathToFileURL(<vault path>).href` against `@modelcontextprotocol/server-filesystem@2026.1.14`. The `slice(7)` regression in `roots-utils.ts` makes every Windows vault path with a space (OneDrive folders, `Program Files`, etc.) fail with:

```
Skipping invalid path or inaccessible: file:///C:/Users/.../OneDrive%20-%20Microsoft/Vaults/copilot-test
No valid root directories provided by client
```

Confirmed root cause: `slice(7)` keeps the leading `/` before the drive letter and never decodes `%20`, so `path.resolve` produces `C:\C:\...%20...` which `fs.realpath` rejects with `ENOENT`. The fix in this PR (`fileURLToPath`) is correct — verified locally that all common Node-emitted file URIs round-trip cleanly.

+1 for merge — would unblock a class of Windows users whose vault sits anywhere with a space in the path (very common: OneDrive, Documents, etc.).

PR #3361, opened 2026-02-20
fix(filesystem): add symlink startup test and clarify catch block comment

olaservo:
## Summary

Follow-up to #3254 (symlinked allowed directories fix), implementing the two suggestions from the review:

- **Integration test**: Add test in `startup-validation.test.ts` that spawns the server with a symlinked directory argument, verifying the startup code correctly produces both path forms. This closes the gap between the unit test in `path-validation.test.ts` (which manually constructs the allowedDirectories array) and the actual startup behavior.
- **Code comment**: Clarify the catch block comment in `index.ts` to document the design tradeoff: when a directory doesn't exist at startup, only the unresolved form is stored, so full symlink support requires the directory to exist at startup.

## Test plan

- [x] All 146 filesystem tests pass (7 test files)
- [x] New symlink startup test passes on Windows
- [x] Test gracefully skips on systems without symlink permissions (EPERM)
- [x] Build succeeds with no TypeScript errors

🦉 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/filesystem/__tests__/startup-validation.test.ts
+++ b/src/filesystem/__tests__/startup-validation.test.ts
@@ -84,6 +84,32 @@ describe('Startup Directory Validation', () => {
     expect(result.stderr).toContain('Error: None of the specified directories are accessible');
   });
 
+  // Test for macOS /tmp -> /private/tmp symlink issue (GitHub issue #3253)
+  // Verifies that the startup code stores both the symlink path and its
+  // resolved target in allowedDirectories, so requests through either form work.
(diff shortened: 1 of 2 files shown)


cliffhall:
@claude review

a-akimov:
@olaservo Do you still want to update the comment as per Claude's recommendation? I am fine with merging it as it is as well.

a-akimov (approved):

PR #3600, opened 2026-03-16
fix(gitlab): handle undefined response in search_repositories

### LINKED ISSUE #3454: Fix "McpError: Cannot read properties of undefined (reading 'map')" - in server-gitlab
**Describe the bug**
When using @modelcontextprotocol/server-gitlab against Enterprise self-hosted GitLab(v18.4) with google adk, the `search_repositories` tool crashes with "McpError: Cannot read properties of undefined (reading 'map')".

**To Reproduce**
1. StdioServerParameters(command="npx", env={"gitlab_access_token": access_token, "gitlab_API_URL": URL})
2. StdioConnectionParams(), McpToolset 
3. LlmAgent from google.adk and implementing [git_mcp] to the tools parameters.
4. Running the runner leading to the "McpError: Cannot read properties of undefined (reading 'map')".
5. The given prompt is "Have a look at my gitlab and see which repository I have access to, Kind reminder, when you sue the 'search_repositories' tool, you must provide the 'search' argument"

**Expected behavior**
It should precisely execute which repo I am having and no error.

**Logs**
"McpError: Cannot read properties of undefined (reading 'map')"

**Additional context**

nielskaspers:
## Summary

Fixes #3454

The `search_repositories` tool in `server-gitlab` crashes with `McpError: Cannot read properties of undefined (reading 'map')` when used against Enterprise self-hosted GitLab (v18.4).

**Root cause**: The `/projects` API endpoint on some Enterprise GitLab instances may return a non-array response body. When this `undefined` value is passed to Zod's `z.array()` validator, it internally calls `.map()` on `undefined`, causing the crash.

**Fix**: Add nullish coalescing (`?? []`) to default `projects` to an empty array when the API response is `undefined` or `null`.

```diff
-    items: projects
+    items: projects ?? []
```

**Note**: The `src/gitlab/` directory was previously archived. This PR restores it to apply the fix. The actual code change is a single line in `src/gitlab/index.ts`.

### DIFF
--- a/src/gitlab/Dockerfile
+++ b/src/gitlab/Dockerfile
@@ -1,0 +1,24 @@
+FROM node:22.12-alpine AS builder
+
+COPY src/gitlab /app
+COPY tsconfig.json /tsconfig.json
+
+WORKDIR /app
+
+RUN --mount=type=cache,target=/root/.npm npm install
+
+RUN --mount=type=cache,target=/root/.npm-production npm ci --ignore-scripts --omit-dev
+
+FROM node:22.12-alpine AS release
+
+WORKDIR /app
+
+COPY --from=builder /app/dist /app/dist
+COPY --from=builder /app/package.json /app/package.json
+COPY --from=builder /app/package-lock.json /app/package-lock.json
+
+ENV NODE_ENV=production
+
+RUN npm ci --ignore-scripts --omit-dev
+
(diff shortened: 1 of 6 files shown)


nielskaspers:
Closing this PR — I missed that the gitlab server was archived in the main branch. The issue should be directed to the standalone gitlab server repo instead.

PR #3611, opened 2026-03-17
fix(gitlab): guard against undefined response in search_repositories

### LINKED ISSUE #3454: Fix "McpError: Cannot read properties of undefined (reading 'map')" - in server-gitlab

joaquinhuigomez:
## Summary

Fixes `Cannot read properties of undefined (reading 'map')` crash when calling `search_repositories` against Enterprise / self-hosted GitLab instances (e.g. v18.4)

- The GitLab `GET /projects?search=` API returns a bare array on gitlab.com but may return an object wrapper on self-hosted instances. The code passed the raw response directly to Zod schema validation, which calls `.map()` on `undefined`.
- Adds null-safety guard that normalises the response before validation: uses the array directly if present, otherwise extracts `.items` or `.projects` from the wrapper object, falling back to `[]`.

**Note:** The `src/gitlab/` source was previously moved to `modelcontextprotocol/servers-archived` (which is read-only and does not accept PRs). This PR restores the directory with the fix applied. The only functional change is in `searchProjects()` in `src/gitlab/index.ts` — all other files are identical to the archived version. The same fix has also been pushed to [my fork of servers-archived](https://github.com/joaquinhuigomez/servers-archived/tree/fix/gitlab-search-repositories-crash) in case maintainers prefer to apply it there.

Fixes #3454

## Test plan

- [ ] Verify `search_repositories` works against gitlab.com (bare array response)
- [ ] Verify `search_repositories` works against self-hosted GitLab Enterprise (object wrapper response)
- [ ] Verify empty search results return `{ count: 0, items: [] }` instead of crashing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -699,6 +699,10 @@
       "resolved": "src/filesystem",
       "link": true
     },
+    "node_modules/@modelcontextprotocol/server-gitlab": {
+      "resolved": "src/gitlab",
+      "link": true
+    },
     "node_modules/@modelcontextprotocol/server-memory": {
       "resolved": "src/memory",
       "link": true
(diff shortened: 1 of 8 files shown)


joaquinhuigomez:
Hi — just checking if this is still on the review queue. Happy to address any feedback.

joaquinhuigomez:
Closing this out.

PR #3614, opened 2026-03-17
Revalidate git roots on each request

### LINKED ISSUE #3613: Enforce current roots on every git tool call
## Problem
The git server refreshes client roots when listing repositories, but tool execution validates `repo_path` only against the optional command-line repository restriction. If roots change at runtime, a repo path that used to be reachable can still be used in later tool calls.

## Why now
The git server is a reference implementation for safe repository access. Root changes need to revoke access deterministically instead of relying on an undefined reconnect/restart boundary.

## Version under test
- Commit: `b60eca1b3bfcd12e2afe7000882db363a1b89f6e` (`origin/main` on 2026-03-17)
- Runtime: Python `3.14.0`, macOS `darwin/arm64`

## Minimal repro
1. Start the git MCP server with roots support enabled and no fixed `--repository` argument.
2. Expose repo `A` through roots and successfully call a git tool against it.
3. Update the client roots so repo `A` is no longer allowed.
4. Reuse the old `repo_path` in another git tool call.

## Expected behavior
The second call should be rejected based on the current roots.

## Actual behavior
`call_tool()` only validates against the command-line repository restriction, so root-based access is not re-enforced during tool execution.

## Code paths
- `src/git/src/mcp_server_git/server.py`
- `src/git/tests/test_server.py`
- `src/filesystem/roots-utils.ts`
- `src/git/README.md`

## Validation target
Changing roots at runtime should revoke access to repositories that are no longer within the allowed set without requiring a restart.

davidahmann:
Closes #3613

## Problem
The git server resolves the repository allowlist once and can keep accepting paths from roots that were later removed at runtime.

## What changed
- rebuild the allowed repository list from the current runtime roots on every tool call
- validate the requested repo path against that live allowlist before dispatching the tool
- add regression tests for exact matches, removed roots, and empty-root handling

## Validation
- `uv run --python 3.11 pytest tests/test_server.py` in `src/git`

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -255,6 +255,26 @@ def validate_repo_path(repo_path: Path, allowed_repository: Path | None) -> None
         )
 
 
+def validate_repo_path_against_allowed_repositories(
+    repo_path: Path, allowed_repositories: Sequence[Path]
+) -> None:
+    """Validate that repo_path is within at least one allowed repository path."""
+    if not allowed_repositories:
+        raise ValueError("Repository path is not within the current allowed roots")
+
+    for allowed_repository in allowed_repositories:
+        try:
+            validate_repo_path(repo_path, allowed_repository)
+            return
+        except ValueError:
+            continue
+
+    allowed_list = ", ".join(str(path) for path in allowed_repositories)
+    raise ValueError(
+        f"Repository path '{repo_path}' is outside the allowed repositories: {allowed_list}"
+    )
+
+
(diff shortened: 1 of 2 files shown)


davidahmann:
This closes the stale-root gap in the git server, so repo paths removed from the live roots set stop being accepted on later tool calls.

The implementation is minimal: each request rebuilds the current allowlist from the runtime roots and validates the target repo path against that live set before dispatch.

Validation:
- `uv run --python 3.11 pytest tests/test_server.py` (pass)

Inspired by research context: CAISI publishes independent, reproducible AI agent governance research: https://caisi.dev

cliffhall:
@claude review

LuuOW (commented):

PR #3616, opened 2026-03-17
fix(filesystem): merge MCP roots with CLI-provided directories instead of replacing them

### LINKED ISSUE #3602: server-filesystem: MCP roots protocol overwrites CLI-provided allowed directories
## Description

When the MCP client supports the roots protocol, the filesystem server's `oninitialized` handler **replaces** all command-line allowed directories with only the roots provided by the client. This means any additional directories passed via CLI args are silently discarded.

## Steps to Reproduce

1. Configure the filesystem MCP server with multiple allowed directories via CLI args:
   ```
   npx @modelcontextprotocol/server-filesystem /home/user /mnt/Storage /mnt/Games /mnt/Emulators
   ```
2. Connect from a client that supports the MCP roots protocol (e.g., Claude Code) where the working directory is `/home/user`
3. Call `list_allowed_directories`

## Expected Behavior

All four directories should be listed as allowed:
```
/home/user
/mnt/Storage
/mnt/Games
/mnt/Emulators
```

## Actual Behavior

Only the client-provided root is listed:
```
/home/user
```

The other three directories are silently dropped and become inaccessible.

## Root Cause

In `dist/index.js`, the `oninitialized` handler (around line 564) unconditionally replaces `allowedDirectories` with the client roots when the client supports the roots protocol:

```javascript
server.server.oninitialized = async () => {
    const clientCapabilities = server.server.getClientCapabilities();
    if (clientCapabilities?.roots) {
        const response = await server.server.listRoots();
        if (response && 'roots' in response) {
            await updateAllowedDirectoriesFromRoots(response.roots);
        }
    }
};
```

And `updateAllowedDirectoriesFromRoots` does a full replacement:

```javascript
allowedDirectories = [...validatedRootDirs];
```

## Suggested Fix

CLI-provided directories should be preserved and merged with client roots, not replaced. For example:

```javascript
async function updateAllowedDirectoriesFromRoots(requestedRoots) {
    const validatedRootDirs = await getValidRootDirectories(requestedRoots);
    if (validatedRootDirs.length > 0) {
        // Merge with existing CLI-provided directories instead of replacing
        const merged = new Set([...allowedDirectories, ...validatedRootDirs]);
        allowedDirectories = [...merged];
        setAllowedDirectories(allowedDirectories);
    }
}
```

The same fix should apply to the `RootsListChangedNotificationSchema` handler.

## Environment

- `@modelcontextprotocol/server-filesystem` version: 2026.1.14
- Client: Claude Code (supports MCP roots protocol)
- OS: Linux (openSUSE Tumbleweed)
- All directories exist and are mounted ext4 partitions

flobo3:
Fixes #3602

## Problem
When an MCP client supports the roots protocol, the filesystem server's `oninitialized` handler unconditionally replaces all command-line allowed directories with only the roots provided by the client. This causes any additional directories passed via CLI arguments to be silently discarded and become inaccessible.

## Solution
Modified `updateAllowedDirectoriesFromRoots` in `src/filesystem/index.ts` to merge the client-provided roots with the existing `allowedDirectories` (which contains the CLI-provided directories) using a `Set` to ensure uniqueness, instead of completely replacing them.

This ensures that both CLI-provided directories and client-provided roots are accessible to the server.

### DIFF
--- a/src/filesystem/__tests__/roots-merge.test.ts
+++ b/src/filesystem/__tests__/roots-merge.test.ts
@@ -1,0 +1,73 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdtempSync, rmSync, realpathSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+
+describe('updateAllowedDirectoriesFromRoots merge behaviour', () => {
+  let cliDir: string;
+  let mcpDir1: string;
+  let mcpDir2: string;
+
+  beforeEach(() => {
+    cliDir = realpathSync(mkdtempSync(join(tmpdir(), 'mcp-cli-')));
+    mcpDir1 = realpathSync(mkdtempSync(join(tmpdir(), 'mcp-roots1-')));
+    mcpDir2 = realpathSync(mkdtempSync(join(tmpdir(), 'mcp-roots2-')));
+  });
+
+  afterEach(() => {
+    rmSync(cliDir, { recursive: true, force: true });
+    rmSync(mcpDir1, { recursive: true, force: true });
+    rmSync(mcpDir2, { recursive: true, force: true });
+  });
+
+  it('should preserve CLI dirs after init with roots', () => {
+    // Simulates the snapshot: cliAllowedDirectories = [cliDir]
+    const cliAllowedDirectories = [cliDir];
+    const validatedRootDirs = [mcpDir1];
+
+    const merged = new Set([...cliAllowedDirectories, ...validatedRootDirs]);
+    const allowedDirectories = [...merged];
+
+    expect(allowedDirectories).toContain(cliDir);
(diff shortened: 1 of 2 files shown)


cliffhall:
@claude review

flobo3:
Thanks for the thorough review! All three points addressed in 65c4a49:

1. **Accumulation bug** — CLI directories are now saved to an immutable `cliAllowedDirectories` snapshot after startup filtering. Each `updateAllowedDirectoriesFromRoots` call merges against this snapshot, so removed MCP roots are properly replaced rather than accumulated.

2. **Stale comment** — Updated the `RootsListChangedNotificationSchema` handler comment to reflect the new merge semantics.

3. **Tests** — Added `roots-merge.test.ts` covering CLI dir preservation after init, root replacement without accumulation across `roots/list_changed`, and multiple update cycles.

LuuOW (commented):

PR #3651, opened 2026-03-20
fix(filesystem): sort directory_tree entries in stable lexicographic order

### LINKED ISSUE #3539: Make filesystem directory-tree output ordering deterministic
## Problem
Directory tree responses can vary by host filesystem enumeration order, causing unstable output for the same directory across runs.

## Why now
The filesystem server is a reference implementation; deterministic output is required for reproducible comparisons and contract tests.

## Expected behavior
For the same directory contents and exclude patterns, `directory_tree` should return a stable ordering across repeated runs and platforms.

## Claim-to-codepath map
- `src/filesystem/index.ts`
- `src/filesystem/lib.ts`
- `src/filesystem/__tests__/directory-tree.test.ts`
- `src/filesystem/__tests__/structured-content.test.ts`

## Validation requirements
- Enforce deterministic sorting in recursive tree traversal.
- Add test assertions for stable sibling ordering and nested ordering.
- Run filesystem package tests.

## Evidence packet
- Commit under test: `$(git -C /Users/davidahmann/Projects/agent-ecosystem/mcp-servers rev-parse HEAD)`
- Runtime: Node.js + vitest on macOS
- Minimal repro: call `directory_tree` multiple times over same fixture and compare ordering of siblings.

jasonsuhari:
Closes #3539

## Problem

`fs.readdir` returns entries in filesystem enumeration order, which is non-deterministic across platforms and runs. The `directory_tree` tool passed these entries directly into the result array without sorting, so identical directory contents could produce different output depending on host OS and filesystem.

## Fix

Sort entries by name using `localeCompare` after each `readdir` call in `buildTree`, applied at every level of the recursive traversal:

```ts
const entries = (await fs.readdir(validPath, { withFileTypes: true }))
  .sort((a, b) => a.name.localeCompare(b.name));
```

No behaviour changes beyond ordering — no new parameters, no schema changes.

## Tests

Applied the same fix to `buildTreeForTesting` in `directory-tree.test.ts` and added three new tests under a `buildTree ordering` describe block:

- **Stable sibling order** — asserts each name ≤ next name (lexicographically) at the root level
- **Stable nested sibling order** — same assertion on children of a subdirectory
- **Idempotency** — `JSON.stringify` of two consecutive calls produces identical output

All 149 filesystem tests pass.

### DIFF
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -14,6 +14,7 @@ from mcp.types import (
     PromptMessage,
     TextContent,
     Tool,
+    ToolAnnotations,
     INVALID_PARAMS,
     INTERNAL_ERROR,
 )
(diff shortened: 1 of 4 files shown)


cliffhall:
@claude review

LuuOW (commented):

PR #3286, opened 2026-02-03
feat(memory): add file locking to support multi-instance

xz-dev:
## Description

This PR implements proper cross-process file locking for the memory server using `proper-lockfile` to ensure atomic read-modify-write operations across multiple MCP server instances.

**Key changes:**
- Add `proper-lockfile` dependency for cross-process file locking
- Wrap all write operations (`createEntities`, `deleteEntities`, `createRelations`, `deleteRelations`, `addObservations`, `deleteObservations`) in `withLock()` to ensure atomic read-modify-write
- Use `write-file-atomic` for more robust atomic file writes
- Add comprehensive lock tests validating consistency under heavy concurrent writes

### Why PR #3060's in-memory lock solution is insufficient

PR #3060 attempts to fix Issue #2579's race condition using an **in-memory Promise-based lock**. While this works for concurrent operations within a single process, it fundamentally misses the real-world deployment model:

**MCP servers using stdio transport are spawned as separate processes per client.**

```
┌─────────────────┐   ┌─────────────────┐   ┌─────────────────┐
│  Claude Code    │   │   Claude Code   │   │   Claude Code   │
└────────┬────────┘   └────────┬────────┘   └────────┬────────┘
         │                     │                     │
         ▼                     ▼                     ▼
┌─────────────────┐   ┌─────────────────┐   ┌─────────────────┐
│mcp-server-memory│   │mcp-server-memory│   │mcp-server-memory│
│   (Process A)   │   │   (Process B)   │   │   (Process C)   │
│ [In-memory lock]│   │ [In-memory lock]│   │ [In-memory lock]│
└────────┬────────┘   └────────┬────────┘   └────────┬────────┘
         │                     │                     │
         └──────────┬──────────┴──────────┬──────────┘
                    ▼                     ▼
              ┌─────────────────────────────────┐
              │         memory.json             │
              │   (STILL VULNERABLE TO RACES!)  │
              └─────────────────────────────────┘
```

PR #3060 acknowledges this limitation:
> "⚠️ Important: This is a Temporary Measure for Single Editor Usage"
> "If you use multiple LLM clients or editors concurrently with the same memory file, you may still experience race conditions."

This multi-client scenario is the **default use case** for many users, not an edge case.

## Server Details
- Server: memory
- Changes to: tools (all write operations now wrapped with cross-process file lock)

## Motivation and Context

The memory server experiences file corruption when multiple MCP server instances (spawned by different AI clients) attempt concurrent writes to the same `memory.json` file. 

[...]

PR #3060, opened 2025-11-25
Added in-memory locks for the process, fix for Bug #2579 temporary measure.

davidw-philips:
Adds in-memory locks to prevent race conditions in mcp-server-memory that caused file corruption and "Unexpected non-whitespace character after JSON" errors

## Description

Solves the following issue: Bug Report and Resolution: mcp-server-memory Failures due to Race Condition and Environment Misconfiguration #2579

The memory server had no protection against concurrent file writes. When multiple operations tried to read, modify, and write to the `memory.jsonl` file simultaneously, they would corrupt the file causing JSON parsing errors. This fix implements a Promise-based in-memory lock manager that serializes all file operations, ensuring atomic read-modify-write cycles without data loss or corruption.

## Server Details

- Server: memory
- Changes to: Core file operation logic and locking mechanism

## Motivation and Context

The memory server was experiencing critical failures when handling concurrent tool invocations:
1. **Race Condition**: Multiple concurrent read-modify-write operations on `memory.jsonl`
2. **File Corruption**: Interleaved writes resulted in truncated JSON lines
3. **Data Loss**: Some writes would completely overwrite others
4. **JSON Parse Errors**: "Unexpected non-whitespace character after JSON" when reading corrupted files

This was particularly problematic when Claude made multiple memory operations in quick succession or when the server handled rapid concurrent requests. The fix ensures all operations are properly serialized while maintaining high performance through JavaScript's asynchronous Promise-based queue system.

## How Has This Been Tested?

Comprehensive testing with vitest including:
- **Race Condition Prevention Tests** (5 tests): Concurrent entity/relation creation, high-concurrency scenarios, mixed read/write operations
- **Lock Verification Tests** (6 tests): Serialization verification, JSONL format integrity, atomic operation verification
- **Existing Tests**: All 30 original knowledge graph tests continue to pass
- **Coverage**: 82.69% statement coverage, 97.46% branch coverage, 94.73% function coverage

Total: **50/50 tests passing** with no regressions

## Breaking Changes

No breaking changes. The fix is transparent to users:
- Same API and tool signatures
- Same file format (JSONL)
- No configuration changes required
- Backward compatible with existing `memory.jsonl` files

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

[...]

Issue #2579, opened 2025-08-19
Bug Report and Resolution: `mcp-server-memory` Failures due to Race Condition and Environment Misconfiguration


Manamama-Gemini-Cloud-AI-01:
### Summary

This report details the investigation and resolution of a critical bug that rendered the `mcp-server-memory` toolset non-functional, causing "Unexpected non-whitespace character after JSON" errors. The root cause was a combination of a race condition in the server's file-writing logic and an environment misconfiguration that prevented the execution of the corrected code. The issue was ultimately resolved after the user manually deleted the corrupted `memory.json` file, allowing the fixed server environment to take over.

### Root Cause Analysis

The investigation revealed a multi-faceted root cause:

1.  **Code Bug:** The server's code lacked a file lock on the `memory.json` file. This created a **race condition** where concurrent writes would corrupt the file. This was demonstrated by the provided `test_race_condition.py` script.
2.  **Data Corruption:** The corrupted `memory.json` file was the direct cause of the JSON parsing errors. The server would attempt to read the malformed file and fail.
3.  **Environment Issue:** The primary blocker was an environment misconfiguration. The system was using `npx` to run an old, buggy version of the server from the internet, instead of the locally compiled and fixed version that included the race condition fix.

### Evidence of the Bug (Race Condition)

The following Python script simulates the original buggy logic. It starts two processes that attempt to write to the same JSON file concurrently without a lock. This can result in a corrupted file, demonstrating the race condition.

<details>
<summary>Bug Simulation Code</summary>

```python
import json
import os
from multiprocessing import Process, Manager

# The file to be shared and potentially corrupted
TEST_FILE = 'test_memory.json'

# This function simulates the buggy load/save logic from the TypeScript code
def buggy_operation(process_id, new_entry, shared_dict):
    """Simulates a single process reading, updating, and writing the file."""
    try:
        # 1. Load the graph (read from file)
        try:
            with open(TEST_FILE, 'r') as f:
                data = f.read()
                # In a race condition, this data could be incomplete
                graph = json.loads(data) if data else {'entries': []}
        except (FileNotFoundError, json.JSONDecodeError):
            graph = {'entries': []}

        # 2. Add the new entry
        graph['entries'].append(new_entry)
        
        # 3. Save the graph (write back to file)
        with open(TEST_FILE, 'w') as f:
            # This is the critical section where the race condition happens.
            # One process can overwrite the other's changes.
            json.dump(graph, f)
        
[...]

PR #3653, opened 2026-03-20
fix(memory): add trailing newline to JSONL output

nickveenhof:
## Summary

`saveGraph` writes `lines.join("\n")` without a trailing newline. The [JSONL spec](https://jsonlines.org/) requires every record to be terminated by `\n`. This makes the output non-conformant and fragile.

## The problem

`saveGraph` produces a file where the last line has no terminating `\n`:

```
{"type":"entity","name":"Alice","entityType":"person","observations":["test"]}
{"type":"relation","from":"Alice","to":"Bob","relationType":"knows"}  <-- no \n
```

A JSONL file should look like this:

```
{"type":"entity","name":"Alice","entityType":"person","observations":["test"]}\n
{"type":"relation","from":"Alice","to":"Bob","relationType":"knows"}\n
```

The missing terminator means any tool or process that appends to the file (log rotation, crash recovery, file concatenation) produces invalid JSONL:

```
{"type":"relation","from":"Alice","to":"Bob","relationType":"knows"}{"type":"entity","name":"External",...}
```

`JSON.parse()` then throws on that line:

```
Unexpected non-whitespace character after JSON at position 88 (line 1 column 89)
```

## How I hit this

I run the memory server across multiple concurrent sessions pointing at the same `memory.jsonl` file. After several days I found a line with two JSON objects concatenated. Every tool call returned the parse error above until I manually edited the file.

The root cause is likely the race condition described in #2579: multiple MCP server instances calling `saveGraph` concurrently on the same file. #3286 addresses that with cross-process file locking and atomic writes, which is the proper fix.

This PR does not solve the race condition. It makes the output JSONL-spec-conformant, which is correct regardless of whether #3286 lands.

## Fix

```typescript
// Before

// After
await fs.writeFile(this.memoryFilePath, lines.join("\n") + "\n");
```

`loadGraph` already filters empty lines (`filter(line => line.trim() !== "")`), so the trailing `\n` has no side effects on read.

## Tests

Three new tests in `knowledge-graph.test.ts`:

1. **Trailing newline check**: verifies the file ends with `\n`.
2. **Per-line validity**: verifies every non-empty line parses as valid JSON individually.
3. **External append simulation**: writes entities via `saveGraph`, appends a line via `fs.appendFile`, then reads back via a fresh `KnowledgeGraphManager` instance. This test **fails without the fix** and **passes with it**.

All 48 tests pass. No existing tests affected.

## Related

- #2579: the race condition that causes JSONL corruption with concurrent instances
- #3060: in-memory lock (closed, insufficient for multi-process)
- #3286: cross-process file locking with `proper-lockfile` (open, the full fix)


cliffhall:
@claude review

LuuOW (commented):

PR #3738, opened 2026-03-28
fix(filesystem): improve edit_file error with nearest-match diagnostics

### LINKED ISSUE #2034: edit_file tool fails when exact text match not found due to whitespace differences
Describe the bug
The edit_file tool throws "Could not find exact match for edit" error when the oldText parameter doesn't exactly match the file content, typically due to whitespace, line ending, or formatting differences between expected and actual text.
To Reproduce

Use edit_file with oldText parameter containing text that differs in whitespace from actual file content
Tool attempts to find exact match in file
Error occurs: "Could not find exact match for edit"
Edit operation fails completely

Expected behavior
The edit_file tool should either handle minor whitespace differences or provide clear guidance on exact text matching requirements. Alternative: tool should show snippet of actual file content around failed match location.
Logs
Error: Could not find exact match for edit:
[oldText content that failed to match]
Additional context
This occurs frequently when working with code files where whitespace, indentation, or line endings may vary. The tool requires perfect character-by-character matching including all whitespace. Recommendation: always use read_file before edit_file to ensure exact text matching.

Christian-Sidak:
## Summary

When `edit_file` fails to find a match (neither exact nor whitespace-flexible), the error message currently just dumps the `oldText` with no context about what the file actually contains. This makes it hard for both agents and humans to diagnose *why* the match failed.

This PR improves the error to include:
- **File path** in the error message
- **Closest near-match location** with 1-indexed line numbers
- **Match percentage** (how many lines matched after trimming)
- **Line-by-line comparison** showing which lines matched and which diverged, with both the actual and expected content

Example error after this change:
```
Could not find match for edit in /path/to/file.ts
Closest match (75% of lines) at lines 42-45:
   line 42: "  const foo = bar;"
 ! line 43: "  baz();"
     expected: "  baz(x);"
   line 44: "  return result;"
   line 45: "}"

Edit oldText:
  const foo = bar;
  baz(x);
  return result;
}
```

No changes to the actual matching/replacement logic -- this is purely diagnostic improvement.

Fixes #2034

## Test plan
- [x] All 146 existing tests pass
- [x] Updated test name and assertion to match new error format
- [x] Verified error includes file path and diagnostic context on mismatch

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -494,13 +494,13 @@ describe('Lib Functions', () => {
         );
       });
 
-      it('throws error for non-matching edits', async () => {
+      it('throws error for non-matching edits with diagnostic info', async () => {
         const edits = [
           { oldText: 'nonexistent line', newText: 'replacement' }
         ];
-        
+
         await expect(applyFileEdits('/test/file.txt', edits, false))
-          .rejects.toThrow('Could not find exact match for edit');
+          .rejects.toThrow('Could not find match for edit');
       });
 
       it('handles complex multi-line edits with indentation', async () => {
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -248,7 +248,43 @@ export async function applyFileEdits(
     }
 
     if (!matchFound) {
-      throw new Error(`Could not find exact match for edit:\n${edit.oldText}`);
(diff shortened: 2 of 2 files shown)


Christian-Sidak:
Friendly bump -- would love a review when someone has a chance. Happy to address any feedback.

Christian-Sidak:
Closing due to inactivity. Happy to reopen if there's interest.

cliffhall:
@claude review

PR #3740, opened 2026-03-28
feat(time): default to configured timezone when omitted

### LINKED ISSUE #2853: mcp-server-time: Default Time Zone
I feel like forcing specification of a timezone is wasteful, and `--local-timezone` should be the default when unspecified via tool request.

When `mcp-server-time` is provided `--local-timezone`, the LLM should not have to specify the time.
1. Wasteful, error-prone with previous behavior
2. The LLM could learn the TZ simply by looking up the time with this behavior

I am wasting too much context enforcing LLMs to specify the timezone.  It doesn't make sense to them, and it doesn't make sense to me either.

```json
{
"error": "HTTP error 422: {\"detail\":[{\"type\":\"missing\",\"loc\":[\"body\",\"timezone\"],\"msg\":\"Field required\",\"input\":{}}]}"
}
```

```
"error": "500: {'message': 'Error processing mcp-server-time query: Missing required argument: timezone'}"
```
```
{
  "detail": [
    {
      "type": "missing",
      "loc": [
        "body",
        "timezone"
      ],
      "msg": "Field required",
      "input": {}
    }
  ]
}
```

Is this behavior already available and I am missing something?

Christian-Sidak:
## Summary

When `--local-timezone` is configured, the time server currently only uses it as a hint in the tool description, hoping the LLM will pass it explicitly. If the LLM omits the `timezone` argument, the call fails with "Missing required argument."

This PR makes the configured timezone an actual default:

- **`get_current_time`**: `timezone` is now optional. Defaults to `local_tz` when not provided.
- **`convert_time`**: `source_timezone` and `target_timezone` are now optional, both defaulting to `local_tz`. Only `time` remains required.
- Tool descriptions updated to say "Defaults to X if not provided" instead of "Use X as local timezone if no timezone provided by the user."

This means a user can say "what time is it?" and the LLM can call `get_current_time` with no arguments, getting the local time. Previously this would error.

Fixes #2853

## Test plan
- [x] All 38 existing tests pass (no regressions)
- [x] 3 new tests for default timezone behavior:
  - `get_current_time` with no timezone uses local_tz
  - `convert_time` with no source_timezone uses local_tz
  - `convert_time` with no target_timezone uses local_tz
- [x] 41/41 tests pass total

### DIFF
--- a/src/time/src/mcp_server_time/server.py
+++ b/src/time/src/mcp_server_time/server.py
@@ -137,10 +137,10 @@ async def serve(local_timezone: str | None = None) -> None:
                     "properties": {
                         "timezone": {
                             "type": "string",
-                            "description": f"IANA timezone name (e.g., 'America/New_York', 'Europe/London'). Use '{local_tz}' as local timezone if no timezone provided by the user.",
+                            "description": f"IANA timezone name (e.g., 'America/New_York', 'Europe/London'). Defaults to '{local_tz}' if not provided.",
                         }
                     },
-                    "required": ["timezone"],
+                    "required": [],
                 },
                 annotations=ToolAnnotations(
                     readOnlyHint=True,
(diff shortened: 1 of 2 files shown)


Christian-Sidak:
Friendly bump -- any chance this could get a look? Happy to make adjustments.

Christian-Sidak:
Closing due to inactivity. Happy to reopen if there's interest.

cliffhall:
@claude review

PR #3791, opened 2026-04-02
fix: preserve UNC path prefix in isPathWithinAllowedDirectories

### LINKED ISSUE #3756: isPathWithinAllowedDirectories fails for UNC paths on Windows (network drives)

Root cause:

Fix:
} else {
}

To reproduce:

hashwnath:
## Summary

Fixes #3756

On Windows, `isPathWithinAllowedDirectories` fails for UNC paths (e.g. `\\server\share\project`) because `path.resolve(path.normalize(...))` can corrupt the UNC prefix — stripping a leading backslash and turning `\\server\share` into `\server\share`, which `path.resolve` then interprets as drive-relative (e.g. `C:\server\share`). This causes all UNC path access to fail with "Access denied - path outside allowed directories".

## Fix

Extract path normalization into a `normalizePath()` helper that detects UNC paths and normalizes them without `path.resolve`:

- UNC paths (`\\server\share\...`) are always absolute, so `path.normalize` alone is sufficient
- After normalization, verify the `\\` prefix is preserved; restore it if stripped
- Non-UNC paths continue using `path.resolve(path.normalize(...))` as before

## Test plan

- All 53 existing path-validation tests pass
- On Windows with a UNC allowed directory (`\\192.168.x.x\share`), files within the share are now accessible
- Non-UNC paths (drive letters, Unix paths) behave identically to before

### DIFF
--- a/src/filesystem/path-validation.ts
+++ b/src/filesystem/path-validation.ts
@@ -1,8 +1,37 @@
 import path from 'path';
 
 /**
+ * Checks if a path is a Windows UNC path (e.g. \\server\share).
+ */
+function isUNCPath(p: string): boolean {
+  return p.startsWith('\\\\') && !p.startsWith('\\\\?\\');
+}
+
+/**
+ * Normalizes a path, preserving UNC path prefixes on Windows.
+ *
+ * On Windows, path.resolve(path.normalize(uncPath)) can corrupt UNC paths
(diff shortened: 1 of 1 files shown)


cliffhall:
@claude review

LuuOW (commented):

PR #3804, opened 2026-04-03
docs(memory): add WSL2 troubleshooting for npx stdout JSON parse errors

ssam18:
Running the memory server via `npx` in WSL2 causes JSON parse errors at a fixed byte offset because npx writes package resolution output to stdout, which corrupts the MCP stdio JSON-RPC stream. Added a troubleshooting section with two workarounds:

1. global install 
2. direct `node` invocation

So users can get unblocked without needing to dig through issues..


cliffhall:
We aren't accepting unsolicited out-of-scope guides/docs. This doesn't seem like a problem specific to this server but to WSL and stdio servers in general.

BossChaos:
## Code Review

**PR:** [docs(memory): add WSL2 troubleshooting for npx stdout JSON parse errors](https://github.com/modelcontextprotocol/servers/pull/3804) by @ssam18

- ✅ Feature implementation

*Code review*

LuuOW (commented):

PR #1210, opened 2025-04-02
Add Wikidata MCP

zzaebok:
## Description

Adds Wikidata MCP server to README.
https://github.com/zzaebok/mcp-wikidata

## Server Details

- Server: 
- Changes to: README.md

## How Has This Been Tested?

Tested it with MCP client and check the LLM uses the provided tools appropriately.
You can see the client code here: https://github.com/zzaebok/mcp-wikidata/blob/master/src/client.py

Output (some fields are omitted for clarity):

```python
{
  "messages": [
      HumanMessage(
          content="Can you recommend me a movie directed by Bong Joonho?",
      ),
      AIMessage(
          tool_calls=[
              {
                  "name": "search_entity",
                  "args": {"query": "Bong Joon-ho"},
              }
          ],
      ),
      ToolMessage(
          content="Q495980",
          name="search_entity",
      ),
      AIMessage(
          tool_calls=[
              {
                  "name": "get_properties",
                  "args": {"entity_id": "Q495980"},
              }
          ],
      ),
      ToolMessage(
          content='["P345", "P244", "P214", "P227", ...]',
          name="get_properties",
      ),
      AIMessage(
          tool_calls=[
              {
                  "name": "search_property",
                  "args": {"query": "director"},
              }
          ],
      ),
      ToolMessage(
          content="P57",
          name="search_property",
      ),
      AIMessage(
          tool_calls=[
              {
                  "name": "execute_sparql",
                  "args": {
                      "sparql_query": 'SELECT ?film ?filmLabel WHERE {\n  ?film wdt:P57 wd:Q495980.\n  SERVICE wikibase:label { bd:serviceParam wikibase:language "en". }\n} LIMIT 1'
                  },
              }
          ],
      ),
      ToolMessage(
          content='[{"film": {"type": "uri", "value": "http://www.wikidata.org/entity/Q483761"}, "filmLabel": {"xml:lang": "en", "type": "literal", "value": "Mother"}}]',
          name="execute_sparql",
      ),
      AIMessage(
          content='I recommend the movie "Mother," which was directed by Bong Joon-ho.',
      ),
  ]
}
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
[...]

PR #1174, opened 2025-04-01
Add Alibaba Cloud AnalyticDB for MySQL MCP Server to README

lipandongDr:
## Description
Add Alibaba Cloud AnalyticDB for MySQL MCP Server  to README

## Server Details
- Server:  https://github.com/aliyun/alibabacloud-adb-mysql-mcp-server
- Changes to:  add readme

## Motivation and Context
Provide an official MCP Server from Alibaba Cloud AnalyticDB for MySQL so that the users could seamlessly connect AI Agent to  AnalyticDB for MySQL database.

## How Has This Been Tested?
- It has been tested in MCP inspector and some MCP clients(such as: Claude Desktop). 
- The expected results could be provided for querying database or table metadata，querying and analyzing data in Alibaba Cloud AnalyticDB for MySQL database.

## Breaking Changes
No

## Types of changes
- [x] Documentation update

## Checklist
- [x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [x] My changes follows MCP security best practices
- [x] I have updated the server's README accordingly
- [x] I have tested this with an LLM client
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have documented all environment variables and configuration options

PR #3816, opened 2026-04-04
fix(everything): use relatedTask option to enable elicitation over HTTP (#3228)

### LINKED ISSUE #3228: Enable full HTTP elicitation support after SDK streaming API is available
## Summary

The `simulate-research-query` tool in the Everything Server currently implements graceful degradation for elicitation on HTTP transport. When running over HTTP, elicitation attempts fail and the task uses a default interpretation instead of prompting the user.

This is because the SDK's `sendRequest` doesn't work for elicitation over HTTP - the connection is closed after the initial `CreateTaskResult` response.

## Background

The SDK's streaming elicitation API (`elicitInputStream`) enables elicitation over HTTP by creating a new SSE stream. Once this is available in a released SDK version, we can upgrade to full HTTP elicitation support.

## Changes needed

1. Update `@modelcontextprotocol/sdk` dependency to version with streaming elicitation
2. Replace try-catch `sendRequest` pattern with streaming API in `simulate-research-query.ts`
3. Follow spec-compliant pattern:
   - Set `input_required` status before elicitation (spec SHOULD)
   - Use streaming elicitation (works over HTTP)
   - Transition back to `working` after receiving input (spec SHOULD)

## Current behavior

| Transport | Elicitation | Task Completion |
|-----------|-------------|-----------------|
| STDIO | ✅ Works | ✅ Works |
| HTTP | ❌ Graceful degradation | ✅ Works (uses default) |

## Target behavior

| Transport | Elicitation | Task Completion |
|-----------|-------------|-----------------|
| STDIO | ✅ Works | ✅ Works |
| HTTP | ✅ Works (via streaming) | ✅ Works |

galagaevdc:
## Summary

Fixes #3228 — elicitation in `simulate-research-query` now works on all transports (STDIO, SSE, Streamable HTTP).

The original issue assumed `elicitInputStream` (a new SDK API) was required for HTTP support. It turns out the existing SDK already has the mechanism: passing `{ relatedTask: { taskId } }` to `sendRequest`.

## Root Cause

When `sendRequest` is called from the background `runResearchProcess` without `relatedTask`, the SDK sends the message directly through the transport using the `relatedRequestId` of the original `tools/call` request. On HTTP, that response stream is already closed (the `CreateTaskResult` was already returned), so the elicitation silently fails.

## Fix

Pass `{ relatedTask: { taskId } }` as the third argument to `sendRequest` when sending `elicitation/create`. When `relatedTask` is set, the SDK:

1. Enqueues the request in `_taskMessageQueue` instead of sending via transport directly
2. When the client calls `tasks/result` (as the spec says it SHOULD upon seeing `input_required`), the SDK dequeues and delivers the message through the `tasks/result` SSE stream
3. The client responds via POST → routed to the response handler → `sendRequest` resolves

This matches the spec's sequence diagram for the `input_required` flow: https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#input-required-status

## Changes

- `simulate-research-query.ts`: add `{ relatedTask: { taskId } }` to the `sendRequest` call; update comments and report text to remove the outdated "fails on HTTP" messaging
- `tools.test.ts`: add two tests verifying that `sendRequest` receives `relatedTask` for ambiguous queries, and is not called for non-ambiguous queries

## Transport support after fix

| Transport | Elicitation | Task Completion |
|-----------|-------------|-----------------|
| STDIO | ✅ Works | ✅ Works |
| SSE | ✅ Works | ✅ Works |
| Streamable HTTP | ✅ Works | ✅ Works |

## Note
There is a bug in MCP inspector, so it will be impossible to test using it. I've already prepared fix for MCP inspector, but it's not merged yet https://github.com/modelcontextprotocol/inspector/pull/1174

### DIFF
--- a/src/everything/__tests__/tools.test.ts
+++ b/src/everything/__tests__/tools.test.ts
@@ -15,6 +15,7 @@ import { registerTriggerSamplingRequestTool } from '../tools/trigger-sampling-re
 import { registerTriggerElicitationRequestTool } from '../tools/trigger-elicitation-request.js';
 import { registerGetRootsListTool } from '../tools/get-roots-list.js';
 import { registerGZipFileAsResourceTool } from '../tools/gzip-file-as-resource.js';
+import { registerSimulateResearchQueryTool } from '../tools/simulate-research-query.js';
 
 // Helper to capture registered tool handlers
 function createMockServer() {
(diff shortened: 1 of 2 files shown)


cliffhall:
@claude review

galagaevdc:
Hi @cliffhall, I’ve reviewed Claude’s comments. The only issue he suggested fixing is:
```
vi.useRealTimers() not in finally (lines 793, 818): If vi.runAllTimersAsync() throws, timers remain fake and will leak into subsequent tests. Should be:

try {
  await vi.runAllTimersAsync();
} finally {
  vi.useRealTimers();
}
```
However, this pattern is commonly used in other tests, so I don’t think it needs to be changed:
```
await vi.runAllTimersAsync();
vi.useRealTimers();
```
Please let me know if you have a different opinion.

cliffhall (commented):
## Review — `relatedTask` for elicitation over HTTP

I pulled this PR down, rebased it onto current `main`, and verified it end-to-end. **The approach is correct and the fix is the right one.** Details and a few non-blocking observations below.

### Verification

- **Rebased onto current `main`** (the branch was ~25 commits behind, including the zod v4 / SDK `^1.29.0` upgrade). It rebased **cleanly, no conflicts**.
- `npm run build` clean; `vitest` **61 passing**, including the two new tests here.
- **End-to-end over Streamable HTTP** (SDK 1.29 client): task goes `working → input_required`, the `elicitation/create` is delivered, the client's `accept` resolves, and the final report shows the real *"After receiving clarification..."* path — **not** the "unavailable" fallback.
- **End-to-end over stdio**: same result, so adding `relatedTask` does **not** regress the transport that already worked.
- Confirmed in **Inspector V2**, which handles `input_required` correctly (as does Inspector V1 via the author's #1174).

### The mechanism is right

The PR description nails it: the issue (#3228) assumed the new streaming `elicitInputStream` API (SDK #1210) was required, but `{ relatedTask: { taskId } }` on the existing `sendRequest` is sufficient and simpler. For reference, both are now available in SDK 1.29 (#1210 merged 2026-02-11), but `relatedTask` is the lighter-weight, spec-aligned choice here — it enqueues the request and delivers it through the `tasks/result` stream on the client's poll, matching the spec's `input_required` sequence. No reason to reach for `elicitInputStream` for this.

The `input_required` → (elicit) → `working` status transitions are spec-compliant, and the comment/report-text updates accurately describe the new behavior.

### Non-blocking: capability check is now too loose for form-mode

This is **pre-existing** (not introduced by this PR), but it becomes more relevant now that the elicitation path actually runs on HTTP:

```ts
const clientSupportsElicitation = clientCapabilities.elicitation !== undefined;
...
ambiguous: validatedArgs.ambiguous && clientSupportsElicitation,
```

Under SDK ≥1.29's `getSupportedElicitationModes`, a client that declares `elicitation: { url: {} }` (URL-mode only) but **not** `form` is treated as *not* supporting form mode — so this tool's form-mode `elicitation/create` is rejected with `-32602 Client does not support form-mode elicitation requests`, and the `try/catch` degrades to the default interpretation. It won't crash, but such a client silently won't get the disambiguation prompt. Consider gating on form-mode support specifically (the SDK exports `getSupportedElicitationModes(capabilities.elicitation)`) rather than "any elicitation capability." Reasonable as a follow-up rather than a blocker.

### Nits (all optional)

- **Tests assert the contract, not the delivery.** The two new tests mock `sendRequest` and assert `relatedTask` is passed (good — that's the actual code change) and that non-ambiguous calls don't elicit. They don't exercise real cross-transport delivery, which is the point of the fix — that's inherently the SDK's job, and I covered it with the e2e runs above. Might be worth a line in the PR noting manual HTTP verification.
- **Stronger assertion available.** The first test could also assert the accepted clarification propagates into the stored result (e.g. `storeTaskResult` receives a report containing the chosen interpretation), locking in the accept-handling path, not just the request shape.
- **Fallback string.** The catch hardcodes `"technical (default - elicitation unavailable)"`; minor, and consistent with the existing decline/cancel strings.

### Verdict

LGTM on substance — correct mechanism, verified on stdio + Streamable HTTP, spec-compliant flow. The only thing it strictly needs is a **rebase onto current `main`** (clean, as noted). The form-mode capability gating is worth a follow-up but isn't a blocker for the demo's happy path.

cliffhall (approved):
LGTM! 👍

Tested locally with a V1 inspector running [PR #1174](https://github.com/modelcontextprotocol/inspector/pull/1174) and a V2 Inspector which handles input_required properly.

cliffhall:
Thanks for this, @galagaevdc

PR #3852, opened 2026-04-07
fix: all mcp protocol endpoints (/sse, /message, /mc... in sse.ts

orbisai0security:
## Summary
Fix critical severity security issue in `src/everything/transports/sse.ts`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-004 |
| **Severity** | CRITICAL |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-004` |
| **File** | `src/everything/transports/sse.ts:26` |

**Description**: All MCP protocol endpoints (/sse, /message, /mcp) lack authentication middleware, allowing any network attacker to invoke MCP tools without proving their identity. While OAuth mechanisms exist in the codebase, there is no evidence of enforcement on these critical endpoints that handle filesystem operations, knowledge graph modifications, and fetch requests.

## Changes
- `src/everything/transports/sse.ts`

## Verification
- [x] Build passes
- [x] Scanner re-scan confirms fix
- [x] LLM code review passed

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/src/everything/transports/sse.ts
+++ b/src/everything/transports/sse.ts
@@ -16,6 +16,33 @@ app.use(
   })
 );
 
+// Authentication middleware: enforces Bearer token when MCP_AUTH_TOKEN is set
+const authMiddleware = (
+  req: express.Request,
+  res: express.Response,
+  next: express.NextFunction
+): void => {
+  const authToken = process.env.MCP_AUTH_TOKEN;
+  if (!authToken) {
+    // No auth token configured — skip enforcement (development mode)
+    next();
+    return;
+  }
+  const authHeader = req.headers.authorization;
+  if (!authHeader || !authHeader.startsWith("Bearer ")) {
+    res
+      .status(401)
+      .json({ error: "Unauthorized: Missing or invalid Authorization header" });
+    return;
+  }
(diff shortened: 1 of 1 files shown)


olaservo:
Thanks, but closing this. The everything server is a reference/demo implementation — authentication on transports is intentionally left to the deployer per the [MCP authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization), which delegates auth to the transport layer. Adding hardcoded middleware here would obscure the reference behavior we want users to learn from. Automated scanner findings against reference servers don't generally indicate real vulnerabilities; they should be evaluated in the context of the server's purpose. If you believe there's a real spec-level concern, please open an issue with details first so it can be discussed.

orbisai0security:
Totally fair, thanks for the context. I agree that the everything server is a reference/demo, and auth should be handled by the deployer at the transport layer per the spec, so hardcoding middleware here would be misleading. I’ll treat the scanner hit as informational and, if it’s useful, open an issue proposing a prominent docs note/example for how to front these endpoints with auth in real deployments.

cliffhall:
@claude review

Issue #3878, opened 2026-04-08
mcp-server-fetch drops SSR content from streaming/progressive rendering sites


morozow:
## Description

The `fetch` tool silently drops the majority of page content from websites that use progressive server-side rendering (SSR) with a streaming architecture. This affects a growing number of modern web applications that deliver content in two phases:

1. An immediate lightweight loading shell (visible to the user instantly)
2. SSR-rendered content injected into the DOM with `visibility: hidden` or off-screen positioning, waiting for client-side hydration to make it visible

The `extract_content_from_html()` function passes the full HTML through `readabilipy` with Mozilla Readability (`use_readability=True`). Readability's content extraction algorithm treats elements with `visibility: hidden`, `display: none`, or `position: absolute; top: -9999px` as non-content and strips them entirely. This means the actual page content — navigation, articles, documentation, interactive elements — is removed, and only the loading shell text survives into the final markdown output.

## Steps to Reproduce

1. Configure `mcp-server-fetch` as an MCP server
2. Fetch any page from `https://runtimeweb.com` (e.g. `https://runtimeweb.com/concepts`)
3. Observe the returned content

**Expected result:** Full page content — specification text, headings, tables, code blocks, navigation

**Actual result:** A single line: `"Unified Serverless Framework for Full-Stack TypeScript Applications"` — which is the loading shell subtitle, not the page content

## Root Cause Analysis

The site uses event-driven streaming to Lambda-backed SSR. The HTTP response is delivered as `Transfer-Encoding: chunked`:

| Chunk | Size | Content | Timing |
|-------|------|---------|--------|
| 1 | 13.5 KB | Loading shell (`visibility: visible`) | +0ms |
| 2–4 | 70 KB | SSR content from Lambda (`visibility: hidden`, `position: absolute`, `top: -9999px`) | +25ms |
| terminator | 0 | End of response | +71ms |

The HTTP client (`httpx`) correctly receives all chunks — the full 83 KB HTML. The problem occurs afterward, in `extract_content_from_html()`:

```python
def extract_content_from_html(html: str) -> str:
    ret = readabilipy.simple_json.simple_json_from_html_string(
        html, use_readability=True
    )
    if not ret["content"]:
        return "<error>Page failed to be simplified from HTML</error>"
    content = markdownify.markdownify(
        ret["content"],
        heading_style=markdownify.ATX,
    )
    return content
```

[...]

PR #3879, opened 2026-04-08
fix(fetch): add fallback extraction for readability-stripped content

morozow:
## Description

Add fallback content extraction in `mcp-server-fetch` when Mozilla Readability strips too much content from the page. This fixes silent content loss on sites using progressive SSR with hidden pre-hydration markup.

## Server Details

- Server: `fetch`
- Changes to: `extract_content_from_html()` in `src/fetch/src/mcp_server_fetch/server.py`

## Motivation and Context

Sites using progressive server-side rendering (streaming + Lambda SSR) deliver content in two phases:

1. An immediate visible loading shell (small HTML)
2. SSR-rendered content in a hidden container (`visibility:hidden; position:absolute; top:-9999px`) that becomes visible after React hydration

Mozilla Readability treats hidden elements as non-content and strips them. The `fetch` tool receives the full HTML (e.g. 83 KB or 245 KB) but returns only the loading shell text — typically a single line — with no error or warning.

**Before fix:**
```
fetch("https://runtimeweb.com")
→ "Unified Serverless Framework for Full-Stack TypeScript Applications"   (1 line from 245 KB)

fetch("https://stdiobus.com/spec")
→ "Transport-level routing for MCP/ACP protocols"   (1 line from 83 KB)

```

**After fix:**
Both return full page content — headings, paragraphs, tables, code blocks, navigation.

This pattern is used by Next.js streaming, Remix deferred loaders, and custom SSR architectures. The number of affected sites will grow as progressive SSR adoption increases.

## How Has This Been Tested?

- Unit tests covering: normal HTML (no regression), hidden SSR content (fallback activates), empty readability output (fallback activates), large hidden content with small visible shell (real-world pattern)
- All existing tests pass
- Manually verified against `https://stdiobus.com` and `https://runtimeweb.com`

**Verification with curl:**
```bash
# Full HTML arrives — not a network issue
curl -s https://stdiobus.com/ | wc -c        # ~83965 bytes
curl -s https://runtimeweb.com/ | wc -c      # ~245608 bytes

# SSR content is inside hidden wrapper
curl -s https://stdiobus.com/ | grep -c "visibility:hidden"   # 1
```

## Breaking Changes

None. The fix only activates when Readability extracts less than 1% of the HTML size as text. Normal sites where Readability works correctly are unaffected.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix adds a three-stage extraction pipeline:

1. **Readability** (existing) — works for most sites
2. **readabilipy without Readability JS** (new fallback) — less aggressive, doesn't filter by visibility
3. **Raw markdownify** (new last resort) — converts full HTML directly

Stage 2 and 3 only activate when stage 1 produces text shorter than 1% of the input HTML (`max(1, len(html) // 100)`). No new dependencies added.

Root cause: `readabilipy.simple_json.simple_json_from_html_string(html, use_readability=True)` invokes Mozilla Readability which evaluates CSS visibility and discards elements with `visibility:hidden`, `position:absolute; top:-9999px`, `opacity:0`. These styles are standard in progressive SSR for pre-hydration content delivery.

### DIFF
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -27,6 +27,10 @@ DEFAULT_USER_AGENT_MANUAL = "ModelContextProtocol/1.0 (User-Specified; +https://
 def extract_content_from_html(html: str) -> str:
     """Extract and convert HTML content to Markdown format.
 
+    Uses readability for content extraction with fallback mechanisms for cases
+    where readability strips too much content (e.g. sites using hidden divs
+    for SSR hydration).
+
     Args:
         html: Raw HTML content to process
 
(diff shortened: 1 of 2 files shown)


olaservo:
Thanks for filing the issue (#3878) and the first PR to fix it — your analysis of the Readability visibility-stripping problem was spot on and helped frame the solution.

We're going with #3922 which uses the same 3-stage approach but has a more complete test suite (6 fully-mocked tests) and preserves the existing `<error>` behavior for empty pages.

Closing in favor of #3922, but credit to you for identifying the root cause.

---

PR #3880, opened 2026-04-09
fix(fetch): detect encoding for non-UTF-8 pages using charset-normalizer

olegsa:
## Summary

- Add automatic character encoding detection to `mcp-server-fetch` using `charset-normalizer` for pages that don't declare charset in the HTTP `Content-Type` header
- Introduces `get_response_text()` helper that checks `response.charset_encoding` first, then falls back to statistical byte analysis via `charset-normalizer`
- Fixes garbled text when fetching pages served in non-UTF-8 encodings (e.g. windows-1251, windows-1255, windows-1256, euc-kr) without a charset declaration in the HTTP header

## Motivation

Many websites (especially non-English ones) serve content in legacy encodings like windows-1255 (Hebrew), windows-1251 (Cyrillic), windows-1256 (Arabic), or euc-kr (Korean) without declaring the charset in the HTTP `Content-Type` header. The current code uses `response.text` which defaults to UTF-8, producing garbled/mojibake output for these pages.

## Changes

- **`server.py`**: Added `get_response_text()` that uses `charset-normalizer` for encoding detection when HTTP headers lack charset info. Replaced `response.text` with `get_response_text(response)` in `fetch_url()`. Also moved httpx from local imports to a top-level import.
- **`pyproject.toml`**: Added `charset-normalizer>=3.0.0` as an explicit dependency (already a transitive dep via `requests`).
- **`tests/test_server.py`**: Added `TestGetResponseText` class with 5 tests covering UTF-8 passthrough, Ukrainian (windows-1251), Hebrew (windows-1255), Arabic (windows-1256), and Korean (euc-kr) encoding detection.

## Test plan

- [x] All existing tests pass
- [x] New encoding detection tests pass for Ukrainian, Hebrew, Arabic, Korean
- [x] UTF-8 pages with charset in HTTP header still use the standard path
- [x] Non-HTML content (JSON, etc.) is unaffected

Made with [Cursor](https://cursor.com)


olaservo (commented):
Nice fix — the `get_response_text()` helper is well-structured with a conservative fallback chain (charset header → charset-normalizer detection → default). Good test coverage across multiple encodings.

One nit: `test_korean_euc_kr` actually encodes the text as UTF-8, not euc-kr. Consider renaming it to `test_utf8_without_charset_header`, or changing it to actually encode as euc-kr to match the test name.

Otherwise LGTM.

---

olegsa:
Thanks for the review, @olaservo! Good catch on the euc-kr test — you're right, it was silently encoding as UTF-8. Fixed in 37da471: the test now actually calls `text.encode("euc-kr")` and uses a longer sample so `charset-normalizer`'s statistical detection has enough bytes to work with. This gives us genuine coverage of the detection path for Korean rather than a duplicate of `test_utf8_passthrough`. All 5 `TestGetResponseText` tests still pass locally.

cliffhall:
@claude review

olegsa:
@cliffhall @olaservo 
are we ok to merge it ?

LuuOW (commented):

PR #3882, opened 2026-04-09
fix(filesystem): add optional depth parameter to list_directory

### LINKED ISSUE #2703: [bug] list_directory params is not allowed to have the additional property depth
**Describe the bug**
Model failed to generate a tool call

Failed to parse arguments for tool "list_directory": params is not allowed to have the additional property "depth"

<|start|>assistant<|channel|>commentary to=functions.list_directory <|constrain|>json<|message|>{"path":"/Users/alekseiivanov/dev/github/application/app/composites/contracts/ai_document_analyser","depth":1}

**To Reproduce**
1.  Open LM Studio on on Mac
2. Load lmstudio-community/gpt-oss-120b-MLX-8bit model
3. add filesystem mcp and ask it to list direcotry
4. it will call a tool with a `depth` argument which breaks the flow.

**Expected behavior**
filesystem mcp successfully lists directory as it does with qwen3 models.

**Logs**
```
Model failed to generate a tool call

Failed to parse arguments for tool "list_directory": params is not allowed to have the additional property "depth"

<|start|>assistant<|channel|>commentary to=functions.list_directory <|constrain|>json<|message|>{"path":"./", "depth":2}
```
**Additional context**

Maybe I'm misusing the tool. But it might be useful to add an optional parameter here
https://github.com/modelcontextprotocol/servers/blob/8ba0ff5e4c74b1b0034f30f4ad75632183df67f4/src/filesystem/index.ts#L117

lui62233:
## Summary

Add optional `depth` parameter to the `list_directory` tool's Zod schema, inputSchema, and handler to support recursive directory listing.

## Problem

Some LLMs (e.g., `lmstudio-community/gpt-oss-120b-MLX-8bit`) generate a `depth` argument when calling `list_directory`, but the schema rejected it as "params is not allowed to have the additional property 'depth'".

## Solution

1. Added `depth: z.number().optional()` to `ListDirectoryArgsSchema`
2. Added `depth` to the `inputSchema` in the tool registration  
3. Updated handler to perform recursive listing when `depth > 0`

Closes #2703

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -132,6 +132,7 @@ const CreateDirectoryArgsSchema = z.object({
 
 const ListDirectoryArgsSchema = z.object({
   path: z.string(),
+  depth: z.number().optional().describe("Recursive depth for directory listing (1 = one level deep)"),
 });
 
 const ListDirectoryWithSizesArgsSchema = z.object({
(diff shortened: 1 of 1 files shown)


olaservo (changes requested):
Thanks for the idea — a `depth` parameter for `list_directory` could be useful. However, the current implementation has some issues that need to be addressed:

1. **Runtime crash**: `listDirRecursive` returns an array of formatted strings, but the code after the call still tries to access `.isDirectory()` and `.name` on each entry (which are Dirent methods, not string methods). This will throw at runtime.

2. **Infinite recursion when depth is omitted**: When `depth` is `undefined` (the default), the condition `maxDepth === undefined || currentDepth < maxDepth` is always true, causing the function to recurse into every subdirectory indefinitely. The tool should behave exactly as before when `depth` is not provided (list only the immediate directory).

3. **Missing input schema update**: The Zod schema is updated but the inline `inputSchema` in `registerTool` is not, so the `depth` parameter won't be advertised to clients.

4. **No tests or CI**: No tests are included, and CI doesn't appear to be running on this branch.

5. **No depth cap**: Consider adding a maximum depth limit (e.g., 10) to prevent a runaway traversal on large filesystems.

---

cliffhall:
@claude review

LuuOW (commented):

hxaxd:
@lui62233 @olaservo, would you be open to me taking this over in a fresh PR and addressing the requested changes?

PR #3892, opened 2026-04-10
fix(release): include all file changes in version bump detection

### LINKED ISSUE #3870: Release script skips version bump for packages with only lockfile changes
**Describe the bug**

`has_changes()` in `scripts/release.py` only considers `.py` and `.ts` files when deciding which packages to version-bump at release time:

```python
relevant_files = [f for f in changed_files if f.suffix in [".py", ".ts"]]
return len(relevant_files) >= 1
```

If a sub-package only has lockfile changes between tags (e.g. `uv.lock` from dependabot), the function returns `False` and the package's `pyproject.toml` version is never bumped to the CalVer tag. It keeps whatever stale version is on `main`.

**To Reproduce**

Between tags `2026.1.14` and `2026.1.26`, the only change in `src/git/` was `uv.lock`:

```bash
git diff --name-only 2026.1.14 2026.1.26 -- src/git/
# src/git/uv.lock
```

The release script skipped `src/git/`, so `pyproject.toml` stayed at `0.6.2`:

```bash
git show 2026.1.26:src/git/pyproject.toml | grep '^version'
# version = "0.6.2"

git show 2026.1.14:src/git/pyproject.toml | grep '^version'
# version = "2026.1.14"
```

**Expected behavior**

`src/git/pyproject.toml` should read `version = "2026.1.26"` on the `2026.1.26` tag, since the package directory had changes included in that release.

**Logs**

N/A this is in the release automation, not a runtime issue.

**Additional context**

Downstream consumers (SBOM generators, CVE scanners) key off the version in `pyproject.toml`. When it says `0.6.2`, advisories with CalVer "fixed in" thresholds never match, so scanners flag the package as vulnerable even though the source is identical to a patched release.

JosephDoUrden:
## Summary

- Remove the `.py`/`.ts` suffix filter from `has_changes()` in `scripts/release.py` so that **all** file changes within a package directory trigger a version bump
- Previously, lockfile-only changes (e.g. `uv.lock` from Dependabot) were silently skipped, leaving stale versions in `pyproject.toml`

Fixes #3870

## Problem

`has_changes()` filtered changed files to only `.py` and `.ts` suffixes:

```python
relevant_files = [f for f in changed_files if f.suffix in [".py", ".ts"]]
```

This caused packages with only lockfile changes between tags to retain stale versions. Downstream consumers (SBOM generators, CVE scanners) rely on the version in `pyproject.toml` to match the CalVer release tag.

## Change

Any change within a package directory now counts as a relevant change, removing the suffix filter entirely.

## AI Disclosure

AI assistance (Claude) was used for issue research. The implementation was written and reviewed by the author.

### DIFF
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -112,9 +112,8 @@ def has_changes(path: Path, git_hash: GitHash) -> bool:
             text=True,
         )
 
-        changed_files = [Path(f) for f in output.stdout.splitlines()]
-        relevant_files = [f for f in changed_files if f.suffix in [".py", ".ts"]]
-        return len(relevant_files) >= 1
+        changed_files = output.stdout.splitlines()
+        return len(changed_files) >= 1
     except subprocess.CalledProcessError:
         return False
 


cliffhall:
@claude review

LuuOW (commented):

JosephDoUrden:
Closing this — #4604 replaces the `release.py` heuristic with changesets, so patching the suffix list isn't the right layer any more.

Leaving #3870 open though: the case it describes is `pyproject.toml` drifting from the release tag when only a lockfile changes, which SBOM and CVE tooling reads. Worth a check against the new pipeline once it lands.

PR #3893, opened 2026-04-10
fix(memory): resolve Vitest false positive for expected rejection

### LINKED ISSUE #3073: Vitest False Positive in PR Added in-memory locks for the process, fix for Bug #2579 temporary measure. #3060
Describe the bug

Vitest reports an “Unhandled Errors” warning for a test that intentionally throws an error ("should throw error for non-existent entity").
The test itself passes correctly using .rejects.toThrow(), but Vitest incorrectly surfaces the expected rejection as an unhandled error.

This causes the TypeScript / Test memory CI job to fail even though there is no bug in the memory server implementation.

To Reproduce

Run the memory server tests using:

pnpm test:memory

Look for the test:

"should throw error for non-existent entity"

Observe that the test passes but Vitest outputs an “Unhandled Errors” warning.

In GitHub Actions, this warning results in a failing test job.

Expected behavior

Vitest should correctly treat .rejects.toThrow() as a handled and expected error, and should not report the intentionally thrown error as an unhandled rejection.

CI should pass when the test succeeds, without false-positive warnings.

Logs

Example output:

Unhandled Errors:
  ❯ Error: Entity does not exist
    at ...

The test reports PASS

The warning appears only in the "Unhandled Errors" summary

No runtime errors or implementation failures occur

Additional context

This behavior is a Vitest testing quirk, not a bug in the memory server code.

All locking, file serialization, and race condition tests pass successfully.

The new lock implementation is functioning as expected.

The false warning appears because Vitest sometimes logs expected async errors as “unhandled” even when wrapped in .rejects.toThrow().

Possible mitigations (optional):

Rewrite the test using assert.rejects(...)

Wrap error tests to avoid Vitest’s unhandled rejection reporting

Add scoped process.on("unhandledRejection", () => {}); for this file (not recommended except for CI cleanliness)

If you want, I can also rewrite the test file for a clean CI run.

JosephDoUrden:
## Summary

- Fix Vitest reporting a false-positive "Unhandled Errors" warning for the "should throw error for non-existent entity" test in the memory server
- Wrap the `addObservations` call in an async closure before asserting with `.rejects.toThrow()`, so Vitest's global unhandled rejection handler does not capture the expected error before the assertion intercepts it

Fixes #3073

## Root cause

`addObservations` is an async function that throws synchronously inside a `.map()` callback. When passed directly to `expect().rejects.toThrow()`, the resulting Promise rejection is seen by Vitest's global rejection handler before the `.rejects` wrapper can intercept it. Wrapping in `async () => { await ... }` ensures the rejection is properly scoped.

## Changes

- `src/memory/__tests__/knowledge-graph.test.ts`: Wrap the assertion in an async closure

## Test plan

- [x] All 45 memory server tests pass (`npx vitest run`)
- [x] No "Unhandled Errors" warning in test output

## AI Disclosure

AI assistance (Claude) was used for issue research. The implementation was written and reviewed by the author.

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -146,11 +146,11 @@ describe('KnowledgeGraphManager', () => {
     });
 
     it('should throw error for non-existent entity', async () => {
-      await expect(
-        manager.addObservations([
+      await expect(async () => {
+        await manager.addObservations([
           { entityName: 'NonExistent', contents: ['some observation'] },
-        ])
-      ).rejects.toThrow('Entity with name NonExistent not found');
+        ]);
+      }).rejects.toThrow('Entity with name NonExistent not found');
     });
   });
 


olaservo (approved):
Correct fix for the well-known Vitest/Jest async rejection race. Wrapping in an async closure ensures the rejection is scoped to the `expect` chain. Clean and minimal.

---

cliffhall:
@claude review

LuuOW (commented):

JosephDoUrden:
This has been approved and green since April — mergeable, CI passing, one-file change. Is anything outstanding on it, or is it just waiting for a merge?

PR #3894, opened 2026-04-10
fix(fetch): fall back to raw HTML conversion when Readability returns empty content

### LINKED ISSUE #3878: mcp-server-fetch drops SSR content from streaming/progressive rendering sites

peteroyce:
Fixes #3878

Pages rendered by SSR frameworks (Next.js, Remix) often produce empty or near-empty Readability output because their primary content lives in `<script>` hydration payloads or elements that Readability strips out. This causes the fetch server to return an unhelpful `<error>Page failed to be simplified from HTML</error>` response even when the raw HTML contains plenty of readable content.

**Fix:** After running Readability, check the length of the extracted text. If it falls below a configurable threshold (`_READABILITY_MIN_CONTENT_LENGTH = 200` chars), fall back to converting the raw HTML directly with `markdownify` after stripping noise tags (`<script>`, `<style>`, `<nav>`, `<footer>`, `<header>`, `<noscript>`).

**Changes:**
- `server.py`: Added `_html_to_markdown_fallback()` helper and updated `extract_content_from_html()` to use it when Readability returns insufficient content
- `test_server.py`: Added three test cases covering the SSR fallback path, Readability success path (regression guard), and script/style tag stripping

Normal pages with well-structured HTML continue to use Readability as before.

### DIFF
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -1,3 +1,4 @@
+import re
 from typing import Annotated, Tuple
 from urllib.parse import urlparse, urlunparse
 
@@ -23,9 +24,44 @@ from pydantic import BaseModel, Field, AnyUrl
 DEFAULT_USER_AGENT_AUTONOMOUS = "ModelContextProtocol/1.0 (Autonomous; +https://github.com/modelcontextprotocol/servers)"
 DEFAULT_USER_AGENT_MANUAL = "ModelContextProtocol/1.0 (User-Specified; +https://github.com/modelcontextprotocol/servers)"
 
+# Minimum character threshold for Readability output to be considered useful.
+# Pages rendered by SSR frameworks (Next.js, Remix) often produce very short or
+# empty Readability output because their content lives in <script> hydration
+# payloads or elements marked display:none.  When the extracted text falls below
+# this threshold we fall back to a direct HTML→Markdown conversion of the raw
+# page so that callers still receive meaningful content.
+_READABILITY_MIN_CONTENT_LENGTH = 200
+
+# Tags whose content adds no value when doing the raw-HTML fallback conversion.
+_STRIP_TAGS_RE = re.compile(
+    r"<(script|style|nav|footer|header|noscript)[\s>].*?</\1>",
+    re.IGNORECASE | re.DOTALL,
+)
+
+
+def _html_to_markdown_fallback(html: str) -> str:
+    """Convert raw HTML to Markdown after stripping noise tags.
+
+    Used as a fallback when Readability fails to extract meaningful content.
+
+    Args:
+        html: Raw HTML content.
+
+    Returns:
+        Plain Markdown text derived directly from the HTML structure.
+    """
+    cleaned = _STRIP_TAGS_RE.sub("", html)
+    return markdownify.markdownify(cleaned, heading_style=markdownify.ATX)
+
 
 def extract_content_from_html(html: str) -> str:
     """Extract and convert HTML content to Markdown format.
+
+    First attempts extraction via Mozilla Readability (via readabilipy).  If
+    Readability returns empty or very short content — which happens on SSR
+    pages built with Next.js or Remix that rely on client-side hydration — the
+    function falls back to a direct HTML→Markdown conversion of the raw page
+    so that meaningful content is always returned.
 
     Args:
         html: Raw HTML content to process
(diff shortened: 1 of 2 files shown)


olaservo:
Thanks for the contribution! We're going with #3922 for this fix — it uses a 3-stage pipeline (including readabilipy without Readability as a middle ground) and a proportional threshold that scales with page size.

Your noise-stripping idea (removing `<script>`, `<style>`, `<nav>`, etc. before fallback conversion) has merit and could be a useful follow-up enhancement.

Closing in favor of #3922.

---

PR #3905, opened 2026-04-11
fix: treat lockfile-only package changes as releasable

### LINKED ISSUE #3870: Release script skips version bump for packages with only lockfile changes

haosenwang1018:
## Summary
- make `scripts/release.py` treat any tracked file change inside a package directory as a releasable change
- stop filtering release bumps to only `.py` and `.ts` files, which skipped lockfile-only package updates
- add a regression test covering a package with only `uv.lock` changes

## Why
Issue #3870 shows that a package can change between tags via `uv.lock` alone, but the current release script skips bumping its version because `has_changes()` only counts `.py` and `.ts` files.

That leaves `pyproject.toml` on an older version even though the tagged release includes a changed package directory.

## Testing
- `python -m pytest tests/test_release.py -q`

Fixes #3870


olaservo (commented):
The core fix looks good — lockfile-only changes should absolutely be treated as releasable, and the test is thorough (creates a real temp git repo).

One small ask: the README.md comma fix (line 144) is unrelated scope. Could you split that into a separate commit or drop it from this PR? Keeps the git history clean.

Otherwise this is ready to go.

---

BossChaos:
## Code Review

**PR:** [fix: treat lockfile-only package changes as releasable](https://github.com/modelcontextprotocol/servers/pull/3905) by @haosenwang1018

- ✅ Bug/security fix

*Code review*

LuuOW (commented):

PR #3907, opened 2026-04-11
ci: pin TypeScript workflow actions to immutable SHAs

grtninja:
## Summary
- pin the floating actions in `.github/workflows/typescript.yml` to immutable commit SHAs
- keep the patch scoped to the existing TypeScript workflow only
- preserve the current job matrix, publish flow, and permissions

## Why
This workflow handles package publishing and uses `NODE_AUTH_TOKEN`, so pinning mutable action tags reduces GitHub Actions supply-chain risk without changing behavior.

## Validation
- `git diff --check`
- `python -c "import pathlib, yaml; p=pathlib.Path(r'.github/workflows/typescript.yml'); yaml.safe_load(p.read_text(encoding='utf-8')); print('yaml_ok', p)"`

## Notes
- no workflow logic changes
- no job/permission changes
- no overlap with the repo's current open workflow PRs on other files

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -14,7 +14,7 @@ jobs:
     outputs:
       packages: ${{ steps.find-packages.outputs.packages }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       - name: Find JS packages
         id: find-packages
         working-directory: src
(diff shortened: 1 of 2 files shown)


olaservo (approved):
Clean supply-chain security hardening. Both SHAs verified against their respective v6 tags. No workflow logic changes.

---

cliffhall:
@claude review

LuuOW (commented):

PR #3920, opened 2026-04-12
fix(filesystem): preserve UNC path prefix during normalization

### LINKED ISSUE #3756: isPathWithinAllowedDirectories fails for UNC paths on Windows (network drives)

Yanhu007:
Fixes #3756

On Windows, `path.normalize` can strip one leading backslash from UNC paths (`\\server\share` becomes `\server\share`). `path.resolve` then interprets this as drive-relative (e.g. `C:\server\share`), causing `isPathWithinAllowedDirectories` to always return `false` for network share paths.

## Changes

- Add `safeNormalize()` helper that detects UNC paths before normalization and restores the `\\` prefix if it gets stripped
- Replace both `path.resolve(path.normalize(...))` calls with `safeNormalize()`
- Add UNC path test cases (within/outside/exact match)

### DIFF
--- a/src/filesystem/__tests__/path-validation.test.ts
+++ b/src/filesystem/__tests__/path-validation.test.ts
@@ -997,4 +997,24 @@ describe('Path Validation', () => {
       expect(forbiddenContent).toBe('ORIGINAL CONTENT'); // Unchanged
     });
   });
+
+  describe('UNC paths (Windows network shares)', () => {
+    // UNC path tests use string-only checks (no filesystem access needed)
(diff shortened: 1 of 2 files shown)


olaservo:
Thanks for working on this! We're going with #3921 which takes a slightly different approach (skipping `path.resolve` entirely for UNC paths rather than repairing after the fact), and its CI is passing.

That said, your test cases are great — we've asked the #3921 author to add UNC-specific tests based on yours. Appreciate the contribution!

Closing in favor of #3921.

---

PR #3921, opened 2026-04-12
Fix isPathWithinAllowedDirectories for UNC paths on Windows

### LINKED ISSUE #3756: isPathWithinAllowedDirectories fails for UNC paths on Windows (network drives)

Christian-Sidak:
## Summary

Fixes #3756

On Windows, `isPathWithinAllowedDirectories` fails for UNC paths (e.g. `\\192.168.1.1\share\`) because `path.normalize` strips a leading backslash from UNC paths, and then `path.resolve` interprets the result as a drive-relative path (e.g. `C:\server\share`). This causes the allowed-directory check to always fail for network shares.

This PR introduces a `normalizePossiblyUNCPath` helper that detects UNC paths and preserves the `\\` prefix through normalization, rather than passing them through `path.resolve` which corrupts them. Non-UNC paths continue to use the existing `path.resolve(path.normalize(...))` logic.

## Changes

- Added `isUNCPath()` helper to detect UNC paths by their `\\` prefix
- Added `normalizePossiblyUNCPath()` that preserves the UNC prefix through `path.normalize`, restoring the stripped backslash when needed
- Applied `normalizePossiblyUNCPath()` to both the input path and allowed directory normalization in `isPathWithinAllowedDirectories`

## Test plan

- All 53 existing tests pass (including the existing UNC path test case)
- The existing UNC test at line 432 of `path-validation.test.ts` covers the core scenario: matching `\\server\share\project` against allowed directories, subdirectory access, and cross-server rejection
- To reproduce the original bug: configure `\\192.168.x.x\share\` as an allowed directory on Windows and attempt file access -- previously always denied, now correctly allowed

### DIFF
--- a/src/filesystem/__tests__/path-validation.test.ts
+++ b/src/filesystem/__tests__/path-validation.test.ts
@@ -439,6 +439,28 @@ describe('Path Validation', () => {
         expect(isPathWithinAllowedDirectories('\\\\other\\share\\project', allowed)).toBe(false);
       }
     });
+
+    it('allows UNC path within allowed UNC directory', () => {
+      if (path.sep === '\\') {
+        const allowed = ['\\\\server\\share\\project'];
+        expect(isPathWithinAllowedDirectories('\\\\server\\share\\project\\subdir\\file.txt', allowed)).toBe(true);
+      }
+    });
+
+    it('blocks UNC path outside allowed UNC directory', () => {
+      if (path.sep === '\\') {
+        const allowed = ['\\\\server\\share\\project'];
+        expect(isPathWithinAllowedDirectories('\\\\server\\share\\other', allowed)).toBe(false);
+        expect(isPathWithinAllowedDirectories('\\\\other\\share\\project', allowed)).toBe(false);
+      }
+    });
+
+    it('allows exact UNC path match', () => {
+      if (path.sep === '\\') {
+        const allowed = ['\\\\server\\share\\project'];
+        expect(isPathWithinAllowedDirectories('\\\\server\\share\\project', allowed)).toBe(true);
+      }
+    });
   });
 
   describe('Symlink Tests', () => {
--- a/src/filesystem/path-validation.ts
+++ b/src/filesystem/path-validation.ts
@@ -1,8 +1,35 @@
 import path from 'path';
 
 /**
+ * Checks if a path is a UNC path (e.g. \\server\share).
+ */
+function isUNCPath(p: string): boolean {
+  return p.startsWith('\\\\');
+}
+
+/**
+ * Normalizes a path that may be a UNC path on Windows.
+ *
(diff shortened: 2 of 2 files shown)


olaservo (commented):
Good approach — skipping `path.resolve` entirely for UNC paths is the right call since that's what causes the corruption. The `normalizePossiblyUNCPath` helper is clean and well-documented. Non-UNC paths are unaffected.

One request before we can merge: could you add test cases for UNC paths? PR #3920 has some good ones you could reference. They should be gated behind `path.sep === '\'` (like the existing UNC test at line 432) since UNC paths are meaningless on Linux. Something like:
- UNC path within allowed UNC directory → true
- UNC path outside allowed UNC directory → false
- Exact UNC path match → true

---

Christian-Sidak:
Added the three UNC path test cases (within allowed directory, outside allowed directory, exact match) all gated behind `path.sep === '\\'`. CI is re-running now.

quocnam1:
FYI I opened #3981 which addresses the same root issue plus three other corruption sites (validatePath, bootstrap init, trailing-separator normalization) that also cause UNC failures on Windows. Happy to close this one if #3981 is preferred, or to rebase on top of it if you'd rather land yours first.

Christian-Sidak:
Thanks for the heads up, @quocnam1! Happy to have this landed alongside or after #3981 -- whatever the maintainers prefer. If #3981's broader scope is more useful, I'm fine closing this one. Either way works for me.

cliffhall:
@claude review

Christian-Sidak:
Removed the redundant inline comment flagged by @cliffhall and the Claude bot review -- the JSDoc on the function already covers it.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

LuuOW (commented):

Christian-Sidak:
Status update: #3981 (broader-scope fix covering additional corruption sites) is still open and unmerged, so this narrower, more targeted fix stands on its own for now. CI is green, and the requested changes have been addressed -- UNC test cases added (within/outside/exact-match, gated on `path.sep === '\\'`) and the redundant inline comment removed. Ready for review.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Thanks @quocnam1 -- since your #3981 covers more corruption sites (validatePath, bootstrap init, trailing-separator normalization), I think it is the stronger overall fix. Happy to close this one if maintainers prefer the consolidated approach. Up to you, @olaservo -- whichever is easier to land.

PR #3922, opened 2026-04-12
fix(fetch): fall back when Readability strips hidden SSR content

### LINKED ISSUE #3878: mcp-server-fetch drops SSR content from streaming/progressive rendering sites

Christian-Sidak:
## Summary

- Adds a three-stage fallback to `extract_content_from_html()` so that pages using progressive SSR (hidden pre-hydration markup) are not silently reduced to a single line of loading-shell text
- Stage 1: Readability (existing behavior, unchanged for normal sites)
- Stage 2: readabilipy without Readability JS (less aggressive, does not filter by CSS visibility)
- Stage 3: Raw markdownify conversion (last resort)
- Fallback only activates when Readability output is shorter than 1% of the input HTML

## Motivation

Sites using progressive server-side rendering (Next.js streaming, Remix deferred, custom Lambda SSR) deliver content in two phases: a small visible loading shell, then the real content in a hidden container (`visibility:hidden; position:absolute; top:-9999px`) that becomes visible after client-side hydration. Mozilla Readability treats hidden elements as non-content and strips them entirely, causing `mcp-server-fetch` to return only the loading shell text with no indication that content was lost.

For example, fetching `https://runtimeweb.com` returns just `"Unified Serverless Framework for Full-Stack TypeScript Applications"` instead of the full page content.

## Changes

- `src/fetch/src/mcp_server_fetch/server.py`: Modified `extract_content_from_html()` to try three extraction stages, falling back only when the previous stage produces disproportionately little text
- `src/fetch/tests/test_server.py`: Added 6 unit tests covering all fallback paths, threshold behavior, and no-regression for normal pages

## Breaking Changes

None. The fix only activates when Readability extracts less than 1% of the HTML size as text. Normal sites where Readability works correctly are completely unaffected.

## Test plan

- [x] All 6 new fallback tests pass
- [x] Existing tests unaffected (they test Readability directly via Node.js, independent of this change)
- [x] No new dependencies added

Fixes #3878

### DIFF
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -27,22 +27,59 @@ DEFAULT_USER_AGENT_MANUAL = "ModelContextProtocol/1.0 (User-Specified; +https://
 def extract_content_from_html(html: str) -> str:
     """Extract and convert HTML content to Markdown format.
 
+    Uses Mozilla Readability via readabilipy as the primary extraction method.
+    Falls back to readabilipy without Readability (less aggressive filtering)
+    or direct markdownify conversion when Readability strips too much content,
+    which commonly happens with progressive SSR sites that deliver content in
+    hidden containers awaiting client-side hydration.
+
     Args:
         html: Raw HTML content to process
 
     Returns:
         Simplified markdown version of the content
     """
+    # Minimum expected content length as a fraction of input HTML.
+    # If extracted text is shorter than this, Readability likely stripped
+    # meaningful content (e.g. hidden SSR markup).
+    min_expected_length = max(1, len(html) // 100)
+
+    # Stage 1: Try Readability (best quality for standard pages)
     ret = readabilipy.simple_json.simple_json_from_html_string(
         html, use_readability=True
     )
-    if not ret["content"]:
-        return "<error>Page failed to be simplified from HTML</error>"
+    content_html = ret.get("content", "")
+    if content_html:
+        content = markdownify.markdownify(
+            content_html,
+            heading_style=markdownify.ATX,
+        )
+        if len(content.strip()) >= min_expected_length:
+            return content
+
+    # Stage 2: Try readabilipy without Readability JS (less aggressive,
+    # does not filter by CSS visibility)
+    ret = readabilipy.simple_json.simple_json_from_html_string(
+        html, use_readability=False
+    )
+    content_html = ret.get("content", "")
+    if content_html:
+        content = markdownify.markdownify(
+            content_html,
+            heading_style=markdownify.ATX,
+        )
+        if len(content.strip()) >= min_expected_length:
+            return content
+
+    # Stage 3: Convert full HTML directly with markdownify (last resort)
     content = markdownify.markdownify(
-        ret["content"],
+        html,
         heading_style=markdownify.ATX,
     )
-    return content
+    if content.strip():
+        return content
+
+    return "<error>Page failed to be simplified from HTML</error>"
 
 
 def get_robots_txt_url(url: str) -> str:
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -3,6 +3,7 @@ from urllib.parse import urlparse, urlunparse
 
 import markdownify
 import readabilipy.simple_json
+from bs4 import BeautifulSoup
 from mcp.shared.exceptions import McpError
 from mcp.server import Server
 from mcp.server.stdio import stdio_server
@@ -71,9 +72,14 @@ def extract_content_from_html(html: str) -> str:
         if len(content.strip()) >= min_expected_length:
             return content
 
-    # Stage 3: Convert full HTML directly with markdownify (last resort)
+    # Stage 3: Convert full HTML directly with markdownify (last resort).
+    # Strip <script> and <style> first — markdownify renders them verbatim as
+    # plain text, which injects large blobs of JS/CSS noise into the output.
+    soup = BeautifulSoup(html, "html.parser")
+    for tag in soup(["script", "style"]):
+        tag.decompose()
     content = markdownify.markdownify(
-        html,
+        str(soup),
         heading_style=markdownify.ATX,
     )
     if content.strip():
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -30,7 +30,7 @@ def extract_content_from_html(html: str) -> str:
 
     Uses Mozilla Readability via readabilipy as the primary extraction method.
     Falls back to readabilipy without Readability (less aggressive filtering)
-    or direct markdownify conversion when Readability strips too much content,
+    or direct markdownify conversion when Readability returns empty content,
     which commonly happens with progressive SSR sites that deliver content in
     hidden containers awaiting client-side hydration.
 
@@ -40,11 +40,6 @@ def extract_content_from_html(html: str) -> str:
     Returns:
         Simplified markdown version of the content
     """
-    # Minimum expected content length as a fraction of input HTML.
-    # If extracted text is shorter than this, Readability likely stripped
-    # meaningful content (e.g. hidden SSR markup).
-    min_expected_length = max(1, len(html) // 100)
-
     # Stage 1: Try Readability (best quality for standard pages)
     ret = readabilipy.simple_json.simple_json_from_html_string(
         html, use_readability=True
@@ -55,7 +50,7 @@ def extract_content_from_html(html: str) -> str:
             content_html,
             heading_style=markdownify.ATX,
         )
-        if len(content.strip()) >= min_expected_length:
+        if content.strip():
             return content
 
     # Stage 2: Try readabilipy without Readability JS (less aggressive,
@@ -69,7 +64,7 @@ def extract_content_from_html(html: str) -> str:
             content_html,
             heading_style=markdownify.ATX,
         )
-        if len(content.strip()) >= min_expected_length:
+        if content.strip():
             return content
 
     # Stage 3: Convert full HTML directly with markdownify (last resort).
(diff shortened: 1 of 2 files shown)


olaservo (dismissed):
This is the strongest of the three Readability fallback PRs (#3879, #3894, #3922):

- 3-stage pipeline (Readability → readabilipy without Readability → raw markdownify) gives a good quality gradient
- Proportional 1% threshold scales with page size, unlike a fixed constant
- Preserves the `<error>` return for truly empty pages (no behavior change)
- 6 fully-mocked deterministic tests with good edge case coverage
- Smallest diff to production code (only 4 lines removed)

We'll close the other two PRs with credit to @morozow for filing the original issue (#3878).

---

morozow:
@Christian-Sidak @olaservo I analyzed both implementations https://github.com/modelcontextprotocol/servers/pull/3922, https://github.com/modelcontextprotocol/servers/pull/3879 in the context of the MCP fetch server and the expected contract of a transport-level content extractor.
https://github.com/modelcontextprotocol/servers/pull/3922 (current) variant introduces opinionated validation via length thresholds, `.strip()`, and a hard `<error>` state. This breaks neutrality: it can discard valid outputs (e.g., whitespace-only content), produces non-data responses, and makes the result dependent on heuristics tied to HTML size. In MCP terms, this violates separation of concerns – the fetch layer should not decide what is "good enough" content.

https://github.com/modelcontextprotocol/servers/pull/3879 variant behaves as a proper extraction primitive: it attempts readability, falls back when needed, and always returns the extracted content without enforcing interpretation or artificial failure states. This keeps the pipeline predictable for agents and preserves full fidelity of the source, which is critical for downstream processing.

https://github.com/modelcontextprotocol/servers/pull/3879 variant is the correct approach for MCP fetch, as it maintains a clean transport contract and avoids embedding policy/validation logic into the extraction layer.

PR needs either refactoring or revert to the implementation from https://github.com/modelcontextprotocol/servers/pull/3879, https://github.com/modelcontextprotocol/servers/pull/3947 to be eligible for merge and properly resolve the issue described in https://github.com/modelcontextprotocol/servers/issues/3878.

morozow:
@olaservo Updated tests for full edge-case coverage and reopened as https://github.com/modelcontextprotocol/servers/pull/3947

Christian-Sidak:
Good points, @morozow -- I agree the length threshold and `<error>` wrapping add opinionation that doesn't belong in the transport layer. Happy to strip those heuristics and just return whatever readabilipy/markdownify produces without any filtering. Note that #3947 was closed -- is there a preferred path forward, or would a simplified version of this PR work?

cliffhall:
@claude review

Christian-Sidak:
Addressed the two actionable items from the Claude code review:

1. **Stage 3 script/style stripping** — added BeautifulSoup pre-processing to strip `<script>` and `<style>` tags before passing to markdownify. This avoids large blobs of raw JS/CSS appearing in the Markdown output when Stage 3 activates.

2. Left the 1% threshold as-is for now — the PR is already approved and the threshold drives the fallback behavior; open to adjusting if maintainers prefer a simpler empty-check approach.

Christian-Sidak:
Removed the 1% content-length threshold as discussed -- Stages 1 and 2 now fall through only when Readability returns empty content (`content.strip()`), not based on a proportional length heuristic. Updated the test accordingly.

This should address @morozow's concern about transport-layer neutrality and the Claude bot review's suggestion to simplify to an empty/None check.

Christian-Sidak:
Fixed the failing fetch test — after removing the 1% threshold, the fallback now triggers on empty/None content rather than proportional size. Updated `test_readability_strips_content_falls_back_to_no_readability` to pass empty string from Stage 1 (which correctly exercises the empty-check path). All other tests still pass.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

LuuOW (commented):

Christian-Sidak:
Status update: the two competing PRs (#3879, #3894) plus the follow-on #3947 have all been closed, so this remains the only open fix for #3878. CI is green, and the earlier feedback has been addressed -- the 1% length threshold was removed (fallback now triggers only on empty/None content) and Stage 3 now strips `<script>`/`<style>` before markdownify per the code review. Ready for another look.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

PR #3938, opened 2026-04-13
fix(filesystem): respect CLI directories when client supports MCP roots

### LINKED ISSUE #3929: MCP roots protocol overrides CLI-provided allowed directories
## Summary

When a client supports the MCP roots protocol, `server-filesystem` unconditionally replaces CLI-provided allowed directories with the client's roots on initialization. This makes it impossible to scope the server to a directory outside the client's project root.

## Reproduction

1. Configure a filesystem MCP server with an explicit directory argument:
   ```json
   {
     "my-project": {
       "command": "npx",
       "args": ["-y", "@modelcontextprotocol/server-filesystem", "/path/to/external-project"]
     }
   }
   ```

2. Use it from a client that supports MCP roots (e.g., Claude Code with project root at `/path/to/my-repo`).

3. Call `list_allowed_directories` — it returns `/path/to/my-repo` instead of `/path/to/external-project`.

## Root cause

In `dist/index.js`, the `oninitialized` handler fetches roots from the client and calls `updateAllowedDirectoriesFromRoots`, which replaces the `allowedDirectories` array entirely — even when directories were explicitly provided via CLI args:

```js
server.server.oninitialized = async () => {
    const clientCapabilities = server.server.getClientCapabilities();
    if (clientCapabilities?.roots) {
        const response = await server.server.listRoots();
        if (response && 'roots' in response) {
            await updateAllowedDirectoriesFromRoots(response.roots);
            // ^ this replaces CLI-provided directories
        }
    }
    // ...
};
```

## Expected behavior

CLI arguments should take precedence over MCP roots. If the user explicitly passes directories, those should be respected. MCP roots should only be used as a fallback when no CLI directories are provided.

## Suggested fix

Only fetch and apply roots when no CLI args were provided:

```js
if (clientCapabilities?.roots && allowedDirectories.length === 0) {
```

An alternative would be to merge both sources (CLI args + roots), but that changes the security model since CLI args are meant to be the explicit allowlist.

## Version

`2026.1.14`

ssam18:
When a client advertises roots support, the server was unconditionally replacing CLI-provided directories with whatever the client returned, making it impossible to scope the server to a directory outside the project root.

The fix tracks whether directories were passed via CLI args at startup and skips the roots fetch in both the `oninitialized` handler and the `roots/list_changed` notification handler when they were. MCP roots now only kick in as a fallback when no CLI args are given. Fixes #3929

### DIFF
--- a/src/filesystem/__tests__/startup-validation.test.ts
+++ b/src/filesystem/__tests__/startup-validation.test.ts
@@ -36,6 +36,60 @@ async function spawnServer(args: string[], timeoutMs = 2000): Promise<{ exitCode
   });
 }
 
+/**
+ * Spawns the filesystem server and performs a minimal MCP handshake advertising roots
+ * capability, then sends a roots/list_changed notification with an unrelated directory.
+ * Returns collected stderr for assertion.
+ */
+async function spawnServerWithRootsHandshake(
+  args: string[],
+  unrelatedRootUri: string,
+  timeoutMs = 2500,
+): Promise<{ stderr: string }> {
+  return new Promise((resolve) => {
+    const proc = spawn('node', [SERVER_PATH, ...args], {
+      stdio: ['pipe', 'pipe', 'pipe'],
+    });
+
+    let stderr = '';
+    proc.stderr?.on('data', (data) => {
+      stderr += data.toString();
+    });
+    proc.stdout?.on('data', () => { /* drain JSON-RPC responses */ });
+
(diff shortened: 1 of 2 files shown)


cliffhall:
@claude review

alialobidm (commented):
``

alialobidm:
@jchris

LuuOW (commented):

PR #3948, opened 2026-04-14
fix(filesystem): tolerate Windows realpath ENOENT

krot4u:
## Summary
- Fix Windows roots parsing/validation when s.realpath() returns ENOENT for existing paths (e.g. SUBST/mapped drives).
- In alidatePath(), fall back to stat() checks when 
ealpath() fails with ENOENT to avoid false  Parent directory does not exist.
- Add/adjust tests to avoid dependence on Windows 8.3 short-path differences and to cover the new fallback behavior.

## Motivation
Cursor/VS Code clients may provide roots like ile:///d%3A/.... On some Windows setups, s.realpath() can fail with ENOENT for paths that exist, causing filesystem MCP to drop roots and reject valid operations.

## Test plan
- 
pm test in src/filesystem


cliffhall:
@claude review

krot4u:
Pushed fixes for review notes:

- Narrowed roots realpath fallback to only suppress ENOENT (rethrow other errors)
- Cleaned up validatePath tests to mock stat() fallback (removed dead realpath mocks)
- Added a mocked unit test covering the SUBST/mapped-drive scenario (realpath ENOENT + stat succeeds)

Tests: npm test -w @modelcontextprotocol/server-filesystem

LuuOW (commented):

PR #3965, opened 2026-04-16
fix(filesystem): normalize path separators in search_files results

### LINKED ISSUE #3517: [Bug]: Filesystem server glob pattern matching inconsistent across platforms
## 环境信息
- 操作系统：Ubuntu 22.04 / macOS Sonoma / Windows 11
- 编程语言：TypeScript / Node.js 20.x
- MCP版本：1.x
- 运行环境：Node.js

## 问题描述
文件系统服务器的glob模式匹配在不同平台上结果不一致。

## 复现步骤
1. 在Windows和Linux上使用相同glob模式
2. 比较匹配结果
3. 观察差异

## 预期行为
相同模式应跨平台返回一致的结果。

## 实际行为
Windows和Linux返回不同的匹配结果。

## 报错日志
N/A - 匹配结果不一致问题。

## 最小复现代码
// Pattern: **/*.ts
// Windows: returns /src/index.ts
// Linux: returns ./src/index.ts (different path format)

## 初步分析
glob库在不同平台使用不同的路径分隔符处理逻辑，导致结果差异。

## 建议解决方案
1. 在执行glob前标准化路径分隔符
2. 使用path模块统一处理路径
3. 添加平台特定的测试用例
4. 在文档中说明跨平台行为差异

---
Reported by Financier-Nuri from Unum AI

ianliuy:
## What's broken?

The filesystem server's `search_files` tool and `directory_tree` exclude patterns return inconsistent results across platforms. On Windows, glob patterns like `**/*.ts` fail to match because paths use backslashes internally.

## Who is affected?

Anyone using the filesystem MCP server on Windows. Linux/macOS users are unaffected since `path.sep` is already `/`.

## When does it trigger?

Every time `search_files` or `directory_tree` (with exclude patterns) is called on Windows. The glob patterns silently fail to match, returning incorrect results.

## Where is the bug?

- `src/filesystem/lib.ts` — `searchFilesWithValidation()`, line computing `relativePath`
- `src/filesystem/index.ts` — `buildTree()`, line computing `relativePath`

Both use `path.relative()` which returns backslash-separated paths on Windows (e.g., `src\index.ts`), then pass them directly to `minimatch` which expects forward slashes.

## Why does it happen?

`path.relative()` uses the OS-native separator (`\` on Windows, `/` on Linux). `minimatch` is a glob library that operates on forward-slash paths. When backslash paths are passed to minimatch on Windows, patterns like `**/*.ts` don't match `src\index.ts`.

## How did we fix it?

Normalize `relativePath` to forward slashes immediately after computing it:

```typescript
// Before
const relativePath = path.relative(rootPath, fullPath);

// After
const relativePath = path.relative(rootPath, fullPath).split(path.sep).join('/');
```

This is applied in both locations. The fix is minimal (2 lines changed) and only affects the path format used for glob matching — the returned full paths from `search_files` remain OS-native.

## How do we know it works?

All 57 related tests pass (lib.test.ts, directory-tree.test.ts, structured-content.test.ts including the `search_files` test). The `.split(path.sep).join('/')` pattern is a no-op on Linux/macOS (where `path.sep` is already `/`), so there's zero risk of regression on those platforms.

Fixes #3517

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -554,7 +554,8 @@ server.registerTool(
       const result: TreeEntry[] = [];
 
       for (const entry of entries) {
-        const relativePath = path.relative(rootPath, path.join(currentPath, entry.name));
+        // Normalize to forward slashes for consistent cross-platform glob matching
+        const relativePath = path.relative(rootPath, path.join(currentPath, entry.name)).split(path.sep).join('/');
         const shouldExclude = excludePatterns.some(pattern => {
           if (pattern.includes('*')) {
             return minimatch(relativePath, pattern, { dot: true });
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -389,7 +389,8 @@ export async function searchFilesWithValidation(
       try {
         await validatePath(fullPath);
 
-        const relativePath = path.relative(rootPath, fullPath);
+        // Normalize to forward slashes for consistent cross-platform glob matching
+        const relativePath = path.relative(rootPath, fullPath).split(path.sep).join('/');
         const shouldExclude = excludePatterns.some(excludePattern =>
           minimatch(relativePath, excludePattern, { dot: true })
         );


cliffhall:
@claude review

ianliuy:
Hi @cliffhall, just a polite follow-up on this Windows path normalization fix.

All checks are green, and the Claude review you triggered concluded this is correct and safe to merge. Would you be willing to take a look for formal approval when you have a chance?

Happy to add the Windows-path regression test Claude suggested if you would prefer that before merge.

LuuOW (commented):

PR #3213, opened 2026-01-14
remove type from loading

SashankMeka1:
## Description
`loadGraph()` loads in type because `saveGraph` saves it as described in [3210](https://github.com/modelcontextprotocol/servers/issues/3210)

## Server Details

- Server: memory
- Changes to: index.ts, knowledge-graph.tst.ts

## Motivation and Context
Fixed a validation error where search_nodes (and other tools) returned entities/relations with a type field that wasn't in the schema.
## How Has This Been Tested?
Added 3 unit tests verifying entities/relations loaded from file

## Breaking Changes
None

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [x] My changes follows MCP security best practices
- [x] I have updated the server's README accordingly
- [x] I have tested this with an LLM client
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have documented all environment variables and configuration options

PR #3966, opened 2026-04-16
fix(memory): remove Zod-based outputSchema to fix serialization failures

### LINKED ISSUE #3622: outputSchema uses Zod objects instead of JSON Schema, violating spec and causing serialization failures
Issue
The memory server's tool definitions use Zod schema objects directly in outputSchema, which violates the MCP specification and causes serialization failures when tools/list endpoint attempts to convert schemas to JSON.

Root Cause
Per the [MCP Tools specification](https://modelcontextprotocol.io/specification/2025-11-25/server/tools), **outputSchema must be a plain JSON Schema object:**

"outputSchema": {
  "type": "object",
  "properties": { /* ... */ },
  "required": [ /* ... */ ]
}

However, the current implementation **passes Zod schema objects directly:**

outputSchema: z.object({ /* ... */ })

When the tools/list endpoint serializes these schemas to JSON-RPC responses, Zod objects fail to serialize, resulting in _zod property errors.

Impact
Spec Violation: outputSchema should be JSON Schema, not language-specific objects
SDK Compatibility: Works with Claude Desktop's bundled SDK (which may have a workaround), but fails with the npm SDK version
General Open Source: Projects using the npm SDK directly will encounter this bug

Zod Isn't Needed for outputSchema
inputSchema needs runtime validation (Zod is appropriate)
**outputSchema is only for documentation/type hints — no validation needed
Using plain JSON Schema for output is cleaner and spec-compliant**

Proposed Solutions

Option 1: Mixed approach (inputSchema uses Zod, outputSchema uses JSON Schema)
Keep Zod for input validation (better DX, runtime safety)
Use plain JSON Schema for output documentation
Minimal change, maintains benefits of Zod for validation

Option 2: Transparent conversion (Zod everywhere, auto-convert on serialization)
Use Zod throughout for consistency
Convert to JSON Schema automatically before sending in tools/list response
Requires zod-to-json-schema dependency

Would appreciate feedback on the preferred approach before submitting a PR.

ianliuy:
## What's broken?

The memory server's tool definitions pass Zod schema objects directly to `outputSchema`, which violates the [MCP spec](https://modelcontextprotocol.io/specification/2025-11-25/server/tools) requiring plain JSON Schema objects. This causes `_zod` serialization errors when the `tools/list` endpoint serializes schemas to JSON-RPC responses.

## Who is affected?

Projects using the npm `@modelcontextprotocol/sdk` directly (rather than Claude Desktop's bundled SDK) encounter serialization failures when calling `tools/list` or invoking tools with `outputSchema` defined.

## Root cause

`outputSchema` fields use Zod raw shapes like:
```ts
outputSchema: {
  entities: z.array(EntitySchema),
  relations: z.array(RelationSchema)
}
```
The MCP spec requires these to be plain JSON Schema objects.

## Why not just replace with plain JSON Schema?

I investigated replacing Zod schemas with hand-written JSON Schema objects (the approach suggested in the issue). However, the current SDK (v1.26.0–1.29.0) **cannot accept plain JSON Schema for `outputSchema`**:

1. **`tools/list` silently drops it** — `normalizeObjectSchema()` returns `undefined` for non-Zod objects, so `outputSchema` never appears in the response
2. **Tool calls crash** — `validateToolOutput()` calls `safeParseAsync(undefined, ...)` which throws `Cannot read properties of undefined (reading '_zod')`

This is a TypeScript SDK limitation, not a spec issue. Until the SDK adds native JSON Schema passthrough for `outputSchema`, the safest fix is to remove it entirely.

## Fix

Remove `outputSchema` and `structuredContent` from all 9 tool registrations. Tools continue to return data via the `content` array (JSON text), preserving full functionality. The only change is that `tools/list` no longer advertises output schemas for these tools.

## Testing

- `npm run build` passes (TypeScript compilation clean)
- Verified at runtime that tools still return correct JSON data via `content`
- Verified that removing `outputSchema` skips output validation (no crashes)

Fixes #3622

### DIFF
--- a/src/memory/index.ts
+++ b/src/memory/index.ts
@@ -266,16 +266,12 @@ server.registerTool(
     description: "Create multiple new entities in the knowledge graph",
     inputSchema: {
       entities: z.array(EntitySchema)
-    },
-    outputSchema: {
-      entities: z.array(EntitySchema)
     }
   },
   async ({ entities }) => {
     const result = await knowledgeGraphManager.createEntities(entities);
     return {
-      content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }],
-      structuredContent: { entities: result }
+      content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }]
     };
   }
 );
@@ -288,16 +284,12 @@ server.registerTool(
     description: "Create multiple new relations between entities in the knowledge graph. Relations should be in active voice",
     inputSchema: {
       relations: z.array(RelationSchema)
-    },
-    outputSchema: {
-      relations: z.array(RelationSchema)
     }
   },
   async ({ relations }) => {
     const result = await knowledgeGraphManager.createRelations(relations);
     return {
-      content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }],
-      structuredContent: { relations: result }
+      content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }]
     };
   }
 );
@@ -313,19 +305,12 @@ server.registerTool(
         entityName: z.string().describe("The name of the entity to add the observations to"),
         contents: z.array(z.string()).describe("An array of observation contents to add")
       }))
-    },
-    outputSchema: {
-      results: z.array(z.object({
-        entityName: z.string(),
-        addedObservations: z.array(z.string())
-      }))
     }
   },
   async ({ observations }) => {
     const result = await knowledgeGraphManager.addObservations(observations);
     return {
-      content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }],
-      structuredContent: { results: result }
+      content: [{ type: "text" as const, text: JSON.stringify(result, null, 2) }]
     };
   }
 );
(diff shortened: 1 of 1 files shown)


cliffhall:
@claude review

ianliuy:
Hi @cliffhall, following up on this one as well.

All checks are green. I also opened #4076 to track restoring `outputSchema` / `structuredContent` for the memory server once the TypeScript SDK supports plain JSON Schema passthrough for `outputSchema`.

Would you be willing to take a look for formal approval when you have a chance?

olaservo:
Thanks for the investigation. Revisiting this against current `main` (SDK `1.29.0` + zod v4), it appears this is no longer an issue:

- `tools/list` now serializes the Zod `outputSchema` to a valid JSON Schema object with no `_zod` serialization error.
- `read_graph` / `search_nodes` / `open_nodes` return `structuredContent` that passes output-schema validation (the `-32602` "additional properties" failure was a separate `type`-leak bug, fixed independently in #3213).

I verified this end-to-end with an SDK client against `main` so I'm going to close this rather than merge.

PR #3981, opened 2026-04-18
Fix UNC path validation: cover all corruption sites on Windows

### LINKED ISSUE #3756: isPathWithinAllowedDirectories fails for UNC paths on Windows (network drives)

quocnam1:
## Summary

Fixes #3756 — Windows UNC path validation fails in multiple sites of the filesystem server, not just `isPathWithinAllowedDirectories`.

This PR supersedes #3921, which patches only `isPathWithinAllowedDirectories`. Three other sites in the filesystem server also corrupt UNC paths, causing every UNC file operation to fail with "Access denied" on Windows even after #3921 is applied. This PR fixes all four sites together and adds tests for the regressions.

## Root cause

On Windows, `path.normalize('\\\\server\\share')` strips a leading backslash (`\\server\share` → `\server\share`), and `path.resolve` then interprets the single-backslash result as drive-relative, producing `C:\server\share`. Any code path that pushes a UNC through `path.resolve` corrupts it before validation.

## The four corruption sites

| # | File | Site | Status upstream |
|---|---|---|---|
| 1 | `path-validation.ts` | `isPathWithinAllowedDirectories` | Fixed by #3921 |
| 2 | `path-validation.ts` | `normalizePossiblyUNCPath` trailing separator | Not covered |
| 3 | `lib.ts` | `validatePath` pre-check `path.resolve` | Not covered |
| 4 | `index.ts` | bootstrap normalization of allowed directories | Not covered |

### Bug #2 — trailing separator

`path.normalize` preserves trailing separators on UNC paths on Windows, unlike `path.resolve` for drive paths. After #3921 strips the leading-backslash issue, `normalizedDir + path.sep` in the startsWith check produces a double separator that never matches a real file path. Any allowed UNC directory configured with a trailing backslash (the common case: `\\server\share\`) fails every subdirectory match.

Fix: strip the trailing separator inside `normalizePossiblyUNCPath`, harmonizing with `path.resolve` behavior on non-UNC paths.

### Bug #3 — `validatePath`

`validatePath` calls `path.resolve(expandedPath)` on the absolute path before the allowed-directory check. On a UNC input, this corrupts the path to `C:\server\share\...` before the patched check ever sees it. Every UNC write, stat, edit, or rename operation fails with "Access denied" regardless of #3921.

Fix: skip `path.resolve` for UNC paths and let `normalizePath` (which is already UNC-aware in `path-utils.ts`) handle them.

### Bug #4 — bootstrap

Same `path.resolve` corruption runs at server startup when normalizing every allowed directory before `setAllowedDirectories`. This means every UNC allowed directory stored in the server's state is already corrupted, independently of runtime request handling.

Fix: same guard as #3 applied in the bootstrap loop.

## Tests

- **`path-validation.test.ts`** — 4 new tests for the trailing-separator case (bug #2): file inside, subdirectories, exact match, sibling prefix rejection. All gated behind `path.sep === '\\'`.
- **`lib.test.ts`** — 6 new tests for `validatePath` end-to-end with UNC (bug #3), using the existing `vi.mock('fs/promises')` pattern and `mockFs.realpath` returning the input unchanged. Covers within/exact/nested/outside-directory/different-server/trailing-separator. All gated behind `process.platform === 'win32'`.
- **Bug #4 (bootstrap)** — no direct unit test. A proper integration test would require a live SMB share in CI, which GitHub Actions doesn't provide. The bootstrap fix applies the exact same `expanded.startsWith('\\\\')` guard as `validatePath`, so the logic is covered by the `lib.ts` tests and by manual validation on a real Windows + SMB setup.

All 159 existing tests continue to pass. Non-UNC paths retain existing behavior on all platforms (same `path.resolve(path.normalize(...))` chain).

## Manual validation

Tested on Windows 11 + Node v25 against an SMB share on a Linux server, with the binary built from this branch loaded via `node` in `claude_desktop_config.json`:

- `list_directory`, `read_text_file`, `write_file`, `get_file_info`, `edit_file` on `\\server\share\...` all succeed.
- Non-UNC paths (`C:\Users\...`) behavior unchanged.
- Allowed directories configured with and without trailing backslashes both work.

Before this PR (even with #3921 applied): every UNC write and stat fails with "Access denied", only `list_directory` succeeds by bypassing `validatePath`.

## Checklist

- [x] Non-UNC path behavior unchanged (verified by existing tests passing on Linux/macOS/Windows)
- [x] All fixes gated on `\\` prefix — zero impact on Unix paths
- [x] Tests gated behind `path.sep === '\\'` or `process.platform === 'win32'` (no-op on Unix CI)
- [x] No new dependencies
- [x] Commit message follows existing style

Happy to split this into multiple commits (one per site + tests) if reviewers prefer that format.

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -228,6 +228,64 @@ describe('Lib Functions', () => {
           process.cwd = originalCwd;
         }
       });
+
+      // Tests for UNC path resolution (bug #3 in the companion PR):
+      // validatePath calls path.resolve() on the absolute path before the
+      // allowed-directory check. On Windows, path.resolve corrupts UNC paths
+      // (\\server\share becomes C:\server\share via path.normalize stripping
+      // a leading backslash). Without the UNC guard, every UNC access throws
+      // "Access denied" even when the UNC directory is explicitly allowed.
+      describe('UNC path handling on Windows', () => {
+        beforeEach(() => {
+          if (process.platform !== 'win32') return;
+          setAllowedDirectories(['\\\\server\\share\\project']);
+          // Simulate fs.realpath returning the UNC path as-is
+          mockFs.realpath.mockImplementation(async (p: any) => p.toString());
+        });
+
+        it('accepts UNC file path within allowed UNC directory', async () => {
+          if (process.platform !== 'win32') return;
+          const testPath = '\\\\server\\share\\project\\file.txt';
+          const result = await validatePath(testPath);
+          expect(result).toBe(testPath);
+        });
+
+        it('accepts UNC directory path matching allowed UNC exactly', async () => {
+          if (process.platform !== 'win32') return;
+          const testPath = '\\\\server\\share\\project';
+          const result = await validatePath(testPath);
+          expect(result).toBe(testPath);
+        });
+
+        it('accepts nested UNC subdirectory within allowed UNC', async () => {
+          if (process.platform !== 'win32') return;
+          const testPath = '\\\\server\\share\\project\\src\\sub\\deep.ts';
+          const result = await validatePath(testPath);
+          expect(result).toBe(testPath);
+        });
+
+        it('rejects UNC path outside allowed UNC directory', async () => {
+          if (process.platform !== 'win32') return;
+          const testPath = '\\\\server\\share\\other\\file.txt';
+          await expect(validatePath(testPath))
+            .rejects.toThrow('Access denied - path outside allowed directories');
+        });
+
+        it('rejects UNC path on a different server', async () => {
+          if (process.platform !== 'win32') return;
+          const testPath = '\\\\other-server\\share\\project\\file.txt';
+          await expect(validatePath(testPath))
+            .rejects.toThrow('Access denied - path outside allowed directories');
+        });
+
+        it('accepts UNC file path when allowed UNC has trailing separator', async () => {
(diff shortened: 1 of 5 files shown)


BossChaos:
## Code Review

**PR:** [Fix UNC path validation: cover all corruption sites on Windows](https://github.com/modelcontextprotocol/servers/pull/3981) by @quocnam1

- ✅ Bug/security fix

*Code review*

BossChaos:
## Code Review

**PR:** [Fix UNC path validation: cover all corruption sites on Windows](https://github.com/modelcontextprotocol/servers/pull/3981) by @quocnam1

- ✅ Bug/security fix

*Code review*

yyzquwu:
Tested this on Windows.

Environment:
- Windows 11
- Node `v24.9.0`
- npm `11.6.0`
- PR head `de13fd24781943c8414d3330b9c0dba78b87307f`

Results:

```text
npm run build --workspace @modelcontextprotocol/server-filesystem
# passed

npm test --workspace @modelcontextprotocol/server-filesystem -- --run
# 7 test files passed
# 159 tests passed
```

The Windows-only UNC cases in `path-validation.test.ts` and `lib.test.ts` ran as part of the filesystem suite. This looks like the more complete fix for #3756 compared with #3921 because it covers `validatePath` and startup allowed-directory normalization too.

LuuOW (commented):

PR #3982, opened 2026-04-18
fix(fetch): accept null for optional parameters

### LINKED ISSUE #2035: fetch fails without optional settings

rohitg00:
Closes #2035.

Clients like LibreChat send explicit `null` for optional Pydantic parameters. The Fetch model typed them as `int`/`bool`, so Pydantic rejected `null` at type coercion before field defaults could apply.

## Fix

Adds a `mode="before"` field_validator on `max_length`, `start_index`, and `raw` that maps `None` to the field's declared default. Fields keep their `int`/`bool` types, validators (`gt=0`, `lt=1_000_000`, `ge=0`) continue to enforce on resolved values, and falsy-coalescing is avoided so valid `start_index=0` and explicit `raw=false` pass through unchanged.

## Tests

9 new tests covering:
- Null handling per field and all-null together
- Omitted fields use defaults
- Explicit values preserved (including `start_index=0`, `raw=false`)
- Validators still reject invalid values

Full suite: 29 passed, ruff clean, pyright 0 errors.

## AI disclosure

Human-led, drafted with Claude Code assistance. I chose the validator pattern after reviewing closed PR #2631 (which used `or` coalescing that silently discards valid falsy values like `start_index=0`). The fix was tested locally before opening the PR.


BossChaos:
## Code Review

**PR:** [fix(fetch): accept null for optional parameters](https://github.com/modelcontextprotocol/servers/pull/3982) by @rohitg00

- ✅ Bug/security fix

*Code review*

LuuOW (commented):

PR #3997, opened 2026-04-21
fix(git): auto-detect repo root from --repository

### LINKED ISSUE #3029: mcp-server-git: Make `--repository .` auto-detect git root like standard git commands
## Problem

When using `mcp-server-git` with `--repository .`, the server uses the literal current working directory instead of finding the git repository root. This differs from standard git behavior and creates portability issues:

1. **Breaks when invoked from subdirectories**: If the MCP client is launched from a subdirectory of the repository, `--repository .` points to that subdirectory instead of the repository root, causing failures.

2. **Forces absolute paths**: Teams must use absolute paths in configuration, requiring each developer to maintain local overrides since everyone has different clone paths.

## Current Behavior

```json
{
  "mcpServers": {
    "git": {
      "command": "uvx",
      "args": [
        "mcp-server-git",
        "--repository",
        "."
      ]
    }
  }
}
```

This only works when the MCP client is launched exactly from the repository root.

## Proposed Solution: Make `.` Auto-detect Repository Root (Preferred)

When `--repository .` is specified, `mcp-server-git` should automatically search upward from the current directory to find the git repository root, just like standard git commands do.

**Implementation:** Use `git rev-parse --show-toplevel` to find the repository root when the specified path is `.` or any path within a git repository.

**Benefits:**
- **Matches standard git behavior**: Every git command (git status, git log, etc.) works from anywhere inside the repository
- **Zero configuration changes needed**: Existing configurations with `--repository .` would just work better
- **Team-friendly**: Shared configuration works for all developers regardless of:
  - Which subdirectory they're working in
  - Where they cloned the repository
  - Which OS they're using

### Option 2: Environment Variable Expansion
Allow `--repository "${GIT_REPO_ROOT}"` but this still requires configuration and environment setup.

### Option 3: New Flag
Add `--auto-detect-root` flag, but this breaks existing configurations and is less intuitive than making `.` work naturally.

## Use Case

We maintain a large IaC repository (~1000 files) with multiple subdirectories (`terraform/`, `applications/`, etc.). Team members work in specific subdirectories, and the current behavior forces us to either:
- Always launch the MCP client from root (inconvenient)
- Maintain per-developer absolute paths in local config files (maintenance burden)

Standard git commands work perfectly from any subdirectory—`mcp-server-git` should too.

## Comparison with Standard Git Behavior

```bash
# These work from ANY subdirectory within a git repo:
cd /path/to/repo/terraform/environments/production
git status              # ✅ Works - finds repo root automatically
git log                 # ✅ Works - finds repo root automatically

# But mcp-server-git with --repository . does NOT:
mcp-server-git --repository .  # ❌ Uses CWD literally, doesn't find repo root
```

Would be happy to contribute a PR for this!

FazeelUsmani:
Lets mcp-server-git --repository <path> accept any path inside a Git working tree instead of only the repo root. At startup the server walks up parents to find .git/, like git rev-parse--show-toplevel.                                                                                                                                                                        
                  
  Fixes the case in #3029 where --repository . only works when the client is launched from the repo root. Now it works from any subdirectory, so a checked-in mcp.json works for everyone without per-developer absolute paths.

**Closes #3029.**                                                                                                                                                                            
                  
  Real git walks up the tree from any subdir. The server should too, otherwise --repository. It is basically unusable in a shared config.                                                    
   
  On the thread, @jonathanhefner and @domdomegg both said not to special-case, so this applies to any path passed to --repository.                                                       
                  
  **How Has This Been Tested?**                                                                                                                                                                
                  
  - Unit tests for resolve_repo_root: root path, nested subdir, non-repo path, and Path(".") via monkeypatch.chdir.                                                                        
  - Two integration tests that drive serve() with a stubbed stdio_server to confirm the resolver is actually wired into startup and the rewrite gets logged.
  - Local smoke test: ran mcp-server-git --repository . from a few nested subdirs and confirmed it picks up the right repo.                                                                
                                                                                                                                                                                           
  **Breaking Changes**                                                                                                                                                                         
                                                                                                                                                                                           
  No impact for configs that already point at the repo root.                                                                                                                               
   
  One behavior change: --repository /repo/non_repo_subdir used to fail with "not a valid Git repository". Now it resolves upward to /repo and starts. If the resolved path differs from the input, it logs it at INFO so it's visible. The README documents this and notes that the flag alone doesn't enforce subtree scoping.
                                                                                                                                                                                           
  The error message also changed from "is not a valid Git repository" to "is not inside a Git repository".                                                                                     
   
  **Types of changes**                                                                                                                                                                         
                  
  - Bug fix (non-breaking change which fixes an issue)                                                                                                                                     
  - New feature (non-breaking change which adds functionality)
  - Breaking change (fix or feature that would cause existing functionality to change)                                                                                                     
  - Documentation update

  **Checklist**                                                                                                                                                                                
   
  - I have read the MCP Protocol Documentation                                                                                                                                             
  - My changes follow MCP security best practices
  - I have updated the server's README accordingly
  - I have tested this with an LLM client
  - My code follows the repository's style guidelines                                                                                                                                      
  - New and existing tests pass locally
  - I have added appropriate error handling                                                                                                                                                
  - I have documented all environment variables and configuration options                                                                                                                  
   
  **Additional context**                                                                                                                                                                       
                  
  Kept the fix scoped to the --repository CLI flag. Applying the same parent-walking to per-tool repo_path would break git_add semantics: repo.git.add(".") runs at the repo root, so passing repo_path=/repo/subdir with files=["."] would stage the whole repo instead of just the subdir. Leaving repo_path strict avoids that. The README says so explicitly.

### DIFF
--- a/src/git/README.md
+++ b/src/git/README.md
@@ -119,6 +119,12 @@ python -m mcp_server_git
 
 ## Configuration
 
+`--repository` accepts any path inside a Git working tree. On startup the server walks up to the enclosing `.git/`, the same way `git rev-parse --show-toplevel` does. So `--repository .` works when a shared config is run from any subdirectory of the repo.
+
+Note that this also means `--repository /repo/subdir` resolves to `/repo` and allows tool calls anywhere under it. If the resolved path differs from the input, startup logs show the rewrite. If you want to restrict operations to a subtree, this flag alone won't do it; use a separate repo or an external sandbox.
+
+Per-tool `repo_path` arguments are not resolved this way and must point at the repo root.
+
 ### Usage with Claude Desktop
 
 Add this to your `claude_desktop_config.json`:
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -234,6 +234,14 @@ def git_show(repo: git.Repo, revision: str) -> str:
             output.append(d.diff)
     return "".join(output)
 
+def resolve_repo_root(path: Path) -> Path:
+    """Return the enclosing repo root for ``path``, like ``git rev-parse --show-toplevel``."""
+    repo = git.Repo(path, search_parent_directories=True)
+    if repo.working_dir:
+        return Path(repo.working_dir)
+    return Path(path).resolve()
+
+
 def validate_repo_path(repo_path: Path, allowed_repository: Path | None) -> None:
     """Validate that repo_path is within the allowed repository path."""
     if allowed_repository is None:
@@ -295,11 +303,14 @@ async def serve(repository: Path | None) -> None:
 
     if repository is not None:
         try:
-            git.Repo(repository)
-            logger.info(f"Using repository at {repository}")
+            resolved = resolve_repo_root(repository)
         except git.InvalidGitRepositoryError:
-            logger.error(f"{repository} is not a valid Git repository")
+            logger.error(f"{repository} is not inside a Git repository")
             return
+        if resolved != repository:
+            logger.info(f"Resolved --repository {repository} to repo root {resolved}")
+        repository = resolved
+        logger.info(f"Using repository at {repository}")
 
     server = Server("mcp-git")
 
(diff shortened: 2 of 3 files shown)


LuuOW (commented):

PR #4001, opened 2026-04-22
fix(tools): restrict get-env to accept key parameter only

### LINKED ISSUE #3986: server-everything: get-env tool returns full process.env without any filtering
I was doing a security audit of the official MCP servers and stumbled into something that gave me pause.

The `get-env` tool in `server-everything` returns the entire `process.env` object -- unfiltered, unsanitized, with no parameter required. Call it with empty args and you get ~6KB of every environment variable on the host machine.

### What I saw

The tool's handler is literally:

```ts
server.registerTool(name, config, async (args): Promise<CallToolResult> => {
  return {
    content: [{ type: "text", text: JSON.stringify(process.env, null, 2) }],
  };
});
```

`inputSchema` is `{}`, so zero arguments triggers it. No annotations, no filtering, no opt-in.

### What this leaks

On a typical Windows dev box this dumps:
- `USERNAME`, `USERDOMAIN`, `COMPUTERNAME` -- host and user identity
- `PATH`, `APPDATA`, `LOCALAPPDATA` -- full directory structure
- `HOMEPATH`, `USERPROFILE` -- home directory path
- Any API keys the user happens to have in env: `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GITHUB_TOKEN`, etc.

The "etc" is the problem. You can't know what's in someone's env.

### Why this matters for an official example

`server-everything` is the demo server. People copy it to learn the protocol. If `get-env` stays as-is, that pattern propagates. I've already seen community servers that copied the "return raw env" pattern because "the official one does it."

Also: the tool has no `annotations` object, so an MCP client has zero signal that this is a high-sensitivity read operation.

### Suggested fixes

**Option 1 (safest):** Require a specific key parameter.

```ts
inputSchema: {
  type: 'object',
  properties: {
    key: { type: 'string', description: 'Specific environment variable name' }
  },
  required: ['key']
}
```

**Option 2 (if you want to keep the bulk dump):** Filter known-sensitive patterns.

```ts
const SENSITIVE = /TOKEN|KEY|SECRET|PASSWORD|CREDENTIAL|AUTH/i;
const filtered = Object.fromEntries(
  Object.entries(process.env).filter(([k]) => !SENSITIVE.test(k))
);
```

**Option 3 (annotation-only, weakest but still useful):** Add `readOnlyHint: true` and a big warning in the description that this returns the full environment.

My vote is Option 1. "Get one env var by name" is still a useful demo of a tool that reads from the host, without being a foot-gun.

---

Happy to open a PR if the maintainers agree on a direction.

Jim1874:
## Summary
Security fix: `get-env` tool was exposing entire `process.env` to any caller.

Changes:
- `inputSchema` now requires a `key` parameter
- Returns only `process.env[key]` instead of full env
- Returns 'not set' for missing keys

Closes #3986

### DIFF
--- a/.changeset/fix-get-env-key-parameter.md
+++ b/.changeset/fix-get-env-key-parameter.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/server-everything": patch
+---
+
+Fix security issue in get-env tool: require a specific key parameter instead of returning the entire process.env object unfiltered. The tool now accepts a required `key` argument and returns only the value of that specific environment variable, preventing accidental exposure of sensitive data such as API keys, tokens, and credentials.
--- a/src/everything/__tests__/tools.test.ts
+++ b/src/everything/__tests__/tools.test.ts
@@ -151,31 +151,33 @@ describe('Tools', () => {
   });
 
   describe('get-env', () => {
-    it('should return all environment variables as JSON', async () => {
+    it('should return the value of a specific environment variable', async () => {
       const { mockServer, handlers } = createMockServer();
       registerGetEnvTool(mockServer);
 
       const handler = handlers.get('get-env')!;
       process.env.TEST_VAR_EVERYTHING = 'test_value';
-      const result = await handler({});
+      const result = await handler({ key: 'TEST_VAR_EVERYTHING' });
 
       expect(result.content).toHaveLength(1);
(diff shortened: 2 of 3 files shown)


LuuOW (commented):

PR #4009, opened 2026-04-22
docs(everything): warn that get-env may expose secrets

beejak:
Extend the `get-env` tool description so operators know environment values can include API keys and tokens.


LuuOW (commented):

Issue #1970, opened 2025-05-30
🐛 MCP Filesystem: macOS Screenshot Files Fail with Unicode Characters (ENOENT Error)


jagansanikommu:
# 🐛 Bug Report: MCP Filesystem Operations Fail with macOS Screenshot Files

## **Issue Summary**
MCP filesystem operations (move, read, access) fail on macOS system-generated screenshot files despite files being visible in directory listings. This affects file organization workflows and creates inconsistent behavior between file listing and file manipulation operations.

## **Environment**
- **OS**: macOS (multiple versions affected)  
- **MCP Server**: @modelcontextprotocol/server-filesystem
- **Node.js**: (applicable versions)
- **Files Affected**: System-generated screenshot files specifically

### **Expected Behavior**
- Files visible in `list_directory` operations should be accessible for `move_file`, `read_file`, and other filesystem operations
- All files with standard characters (including spaces) should be movable

### **Actual Behavior**
- Screenshot files appear in directory listings but throw `ENOENT: no such file or directory` errors during move/access operations
- Regular files with spaces work perfectly when created manually
- Only affects macOS system-generated screenshots

## **Reproduction Steps**

1. Take screenshots on macOS (creates files like `Screenshot 2025-04-23 at 2.40.40 PM.png`)
2. Use MCP `list_directory` tool - files appear normally
3. Attempt to use `move_file` or `read_file` on these screenshots
4. Operations fail with ENOENT errors
5. Create a test file with similar naming manually - it works fine

## **Root Cause Analysis**

macOS screenshot files contain **non-standard Unicode characters**:
- Non-breaking spaces (U+00A0) instead of regular spaces (U+0020)  
- Special Unicode characters in timestamps and punctuation
- Extended file attributes that interfere with standard Node.js file operations

### **Evidence**
```bash
# This works (manually created file)
touch "Screenshot 2025-04-23 at 2.40.40 PM copy.png"
# MCP can move this file successfully

# This fails (macOS system-generated)  
mv "Screenshot 2025-04-23 at 2.40.40 PM.png" ./Images/
# ENOENT error in MCP operations
```

## **Impact**
- **High**: Affects common file organization workflows
- **User Experience**: Creates confusion when "visible" files can't be accessed
- **Productivity**: Forces manual workarounds for basic file management
- **Cross-platform**: Specifically impacts macOS users with no clear error messaging

### **For Users:**
1. Manual drag-and-drop in Finder
2. Rename files to simpler names first
[...]

PR #4016, opened 2026-04-22
fix(filesystem): normalize paths to NFC for Unicode consistency

Will-hxw:
## Summary
Apply `.normalize("NFC")` to path strings in `isPathWithinAllowedDirectories()` so that paths stored in different Unicode normalizations (NFD on macOS APFS, NFC typical) are compared correctly.

## Problem
macOS APFS stores filenames using NFD (decomposed) Unicode normalization, while most clients supply NFC (composed) paths. This causes `isPathWithinAllowedDirectories()` to return `false` for valid paths when the normalization forms differ — a directory named `01_ナレッジベース` (containing a dakuten/combining character) would be inaccessible even though it appears in the allowed list.

The same issue affects macOS system-generated screenshot files whose names contain non-breaking spaces (U+00A0) in different forms.

## Fix
After `path.resolve(path.normalize(...))`, apply `.normalize("NFC")` to both the input path and each allowed directory path before comparison. This converts all paths to the same Unicode normalization form regardless of how they're stored on disk.

## Validation
- Build passes (`npm run build`)
- 144/146 tests pass; 2 pre-existing failures unrelated to this change (root dir cross-drive behavior on Windows)
- Test suite includes a test case `handles unicode characters in paths` which now passes with the fix

## Related
- Issue #1970: MCP Filesystem macOS Screenshot Files Fail with Unicode Characters
- Issue comment confirming root cause and fix pattern: https://github.com/modelcontextprotocol/servers/issues/1970#issuecomment-4277418644

### DIFF
--- a/src/filesystem/__tests__/path-validation.test.ts
+++ b/src/filesystem/__tests__/path-validation.test.ts
@@ -199,9 +199,9 @@ describe('Path Validation', () => {
       expect(isPathWithinAllowedDirectories('/home/user/café', allowed)).toBe(true);
       expect(isPathWithinAllowedDirectories('/home/user/café/file', allowed)).toBe(true);
 
-      // Different unicode representation won't match (not normalized)
+      // Different unicode representations now match after NFC normalization
       const decomposed = '/home/user/cafe\u0301'; // e + combining accent
-      expect(isPathWithinAllowedDirectories(decomposed, allowed)).toBe(false);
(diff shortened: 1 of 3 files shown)


Will-hxw:
The test assertion at line 204 expects decomposed paths not to match, but the fix intentionally normalizes both paths to NFC so they match. The test needs updating: either change toBe(false) to toBe(true) or split into a separate test case verifying that genuinely different paths still do not match.

PR #4023, opened 2026-04-23
feat(filesystem): add maxDepth parameter to directory_tree

factspark23-hash:
## What

Add an optional `maxDepth` parameter to the `directory_tree` tool that limits recursive traversal depth.

## Why

When exploring large monorepos, deep `node_modules` trees, or complex project structures, the full recursive tree can be:
- **Slow** — thousands of entries take seconds to enumerate
- **Overwhelming** — produces massive JSON output that wastes context window
- **Unnecessary** — often you only need 2-3 levels deep

## How

- `maxDepth=0` — shows only root directory entries (no subdirectories expanded)
- `maxDepth=1` — root + immediate children
- `maxDepth=N` — N levels deep
- **Omit for unlimited depth** (backward compatible, no behavior change)

Directories at the max depth boundary are shown with empty `children: []` arrays, preserving tree structure.

## Changes

- `src/filesystem/index.ts`:
  - Added `maxDepth` to `DirectoryTreeArgsSchema` (optional, int, min 0)
  - Updated `buildTree()` to track and limit depth
  - Updated tool description to document the new parameter

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -141,7 +141,8 @@ const ListDirectoryWithSizesArgsSchema = z.object({
 
 const DirectoryTreeArgsSchema = z.object({
   path: z.string(),
-  excludePatterns: z.array(z.string()).optional().default([])
+  excludePatterns: z.array(z.string()).optional().default([]),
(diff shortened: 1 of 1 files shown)


LuuOW (commented):

PR #4024, opened 2026-04-23
feat(filesystem): add compare_files tool for file diff comparison

factspark23-hash:
## What

Add a new `compare_files` tool that compares two files and returns a unified diff.

## Why

Currently there is no way to compare two files through MCP. Users often need to:
- Verify an edit was applied correctly
- Understand differences between file versions
- Review changes before committing

Without this tool, agents must read both files entirely and manually diff them, which is inefficient and error-prone.

## How

```
compare_files({
  file1: "path/to/original.ts",
  file2: "path/to/modified.ts",
  contextLines: 3  // optional, default 3
})
```

Returns unified diff output, or reports files as identical.

## Changes

- `src/filesystem/index.ts`:
  - Added `CompareFilesArgsSchema` with `file1`, `file2`, `contextLines` params
  - Registered `compare_files` tool using existing `createUnifiedDiff` from lib.ts
  - Both files must be within allowed directories

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -159,6 +159,12 @@ const GetFileInfoArgsSchema = z.object({
   path: z.string(),
 });
 
+const CompareFilesArgsSchema = z.object({
+  file1: z.string().describe('Path to the first file'),
+  file2: z.string().describe('Path to the second file'),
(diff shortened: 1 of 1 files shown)


LuuOW (commented):

PR #4025, opened 2026-04-23
feat(filesystem): add head/tail support to read_multiple_files

factspark23-hash:
## What

Add optional `head` and `tail` parameters to the `read_multiple_files` tool, allowing line-limited reads across multiple files at once.

## Why

The single-file `read_text_file` already supports `head` and `tail`, but `read_multiple_files` does not. When working with multiple large files (logs, configs, data files), agents often only need the first/last N lines of each.

Currently the only option is to call `read_text_file` with head/tail N times, which is slower and more verbose.

## How

```
read_multiple_files({
  paths: ["log1.txt", "log2.txt", "log3.txt"],
  tail: 20  // last 20 lines of each file
})
```

- `head` — returns only the first N lines of each file
- `tail` — returns only the last N lines of each file
- Cannot use both simultaneously (throws error, same as `read_text_file`)
- Omitting both reads full files (backward compatible)

## Changes

- `src/filesystem/index.ts`:
  - Added `head` and `tail` to `ReadMultipleFilesArgsSchema`
  - Updated handler to use `headFile`/`tailFile` from lib.ts
  - Updated tool description
  - Added mutual exclusion check (same pattern as `read_text_file`)

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -108,6 +108,8 @@ const ReadMultipleFilesArgsSchema = z.object({
     .array(z.string())
     .min(1, "At least one file path must be provided")
     .describe("Array of file paths to read. Each path must be a string pointing to a valid file within allowed directories."),
+  head: z.number().int().positive().optional().describe("If provided, returns only the first N lines of each file"),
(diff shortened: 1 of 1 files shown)


LuuOW (commented):

PR #4044, opened 2026-04-25
fix(filesystem): use valid MCP content type for read_media_file

### LINKED ISSUE #4029: filesystem: read_media_file returns invalid 'blob' content type, violating MCP spec
## Bug

## Reproduction

```bash
# Start the filesystem server

# Create a text file
echo "hello" > /tmp/test.txt

```

## Expected behavior

Either:

## Environment

MukundaKatta:
Closes #4029.

`read_media_file` returned content with `type: "blob"`, which is not a valid MCP content type, so strict clients drop the response with "MCP error -32602: Invalid tools/call result". Mapped the file's MIME type to the correct MCP content (`image` for images, `audio` for audio, `resource` with embedded blob for other binaries) and updated `outputSchema` to match.

Tests added in `src/filesystem/__tests__/structured-content.test.ts`.

### DIFF
--- a/src/filesystem/__tests__/structured-content.test.ts
+++ b/src/filesystem/__tests__/structured-content.test.ts
@@ -155,4 +155,49 @@ describe('structuredContent schema compliance', () => {
       expect(Array.isArray(structuredContent.content)).toBe(false);
     });
   });
+
+  describe('read_media_file (issue #4029)', () => {
+    it('returns type: "image" for image files (valid MCP content type)', async () => {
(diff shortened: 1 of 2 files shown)


LuuOW (commented):

Seedmanc:
Can it be fixed that it reads images as base64 thus overflowing any context window, while it should instead send it as image to the vision model? Otherwise it pretty much impossible to use if you have to worry about image sizes.

PR #4090, opened 2026-05-01
fix(memory): include type field in output schemas to fix validation error

### LINKED ISSUE #3074: Schema validation error: read_graph fails with "additional properties" when reading enti
# Bug Report: Memory MCP Server - Schema Validation Error

## Summary

All read operations (`read_graph`, `search_nodes`, `open_nodes`) fail with schema validation error when reading entities from JSONL format memory file. The server includes the `type` field in entity objects when reading from file, but the output schema does not allow this field.

## Environment

- **Server**: `@modelcontextprotocol/server-memory`
- **Version**: `2025.11.25` (latest available)
- **OS**: Linux Ubuntu 24.04
- **MCP Client**: Cursor IDE

## Steps to Reproduce

1. Configure Memory MCP server with a JSONL format `memory.json` file
2. File contains entities with structure: `{"type":"entity","name":"...","entityType":"...","observations":[...]}`
3. Call any read operation:
   - `mcp_memory_read_graph`
   - `mcp_memory_search_nodes`
   - `mcp_memory_open_nodes`
4. All operations fail with the same error

## Expected Behavior

Read operations should return entities without the `type` field, as entities are already separated in the response structure (`data.entities` vs `data.relations`).

## Actual Behavior

All read operations return error:
```json
{
  "error": "MCP error -32602: Structured content does not match the tool's output schema: data.entities[0] should NOT have additional properties"
}
```

No partial data is returned - the response contains only the error field.

## Technical Details

- **File format**: JSONL (newline-delimited JSON)
- **Total lines**: 538
- **Entities**: 206
- **Relations**: 332
- **Entity schema in file**: `{type, name, entityType, observations}` (4 fields)
- **Expected entity schema in output**: `{name, entityType, observations}` (3 fields, no `type`)

The `type` field is necessary in the file to distinguish entities from relations, but should be stripped when preparing the response.

## Workaround

- Read data directly from `memory.json` file using file I/O
- Write operations (`create_entities`, `add_observations`, etc.) work correctly
- Only read operations are affected

## Additional Information

- The file structure is valid and can be read successfully using direct file access
- The issue appears to be in the server's schema validation logic
- Cache cleared (`npm cache clean --force`) but issue persists
- Issue occurs with latest server version (2025.11.25)

## Error Code

`MCP error -32602` - Invalid params / Schema validation error

aayushbaluni:
## Summary

Fixes #3074

When structured output validation is enabled, the memory server's `read_graph`, `search_nodes`, and `open_nodes` tools fail with MCP error -32602 because stored entities contain a `type` field not declared in the output schema.

## Root Cause

Entities in the JSONL knowledge graph include a `type` property (discriminator for entity vs relation lines), but the tool output schemas used strict validation that rejected unknown properties.

## Fix

Declared optional `type` literals on entity and relation Zod schemas and attach the discriminators when loading from JSONL so structured responses match both the file format and the output schema.

## Testing

- Ran memory package Vitest suite (`npm test` in `src/memory`)
- Verified TypeScript build succeeds

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -426,7 +426,7 @@ describe('KnowledgeGraphManager', () => {
       expect(JSON.parse(lines[1])).toHaveProperty('type', 'relation');
     });
 
-    it('should strip type field from entities when loading from file', async () => {
+    it('should preserve JSONL type discriminator when loading from file', async () => {
       // Create entities and relations (these get saved with type field)
       await manager.createEntities([
         { name: 'Alice', entityType: 'person', observations: ['test observation'] },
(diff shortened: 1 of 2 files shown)


olaservo:
Thanks for the PR! The underlying issue with the `type` discriminator leaking into entity output and tripping `additionalProperties: false` was fixed in #3213 (commit `8af50b9`) by stripping `type` in `loadGraph()` instead. Since the field is no longer present on returned entities, declaring it in the schema isn't needed, and we'd prefer the output schema stay strict. Closing as superseded by the merged fix.

_Drafted with assistance from Claude Code._

PR #4091, opened 2026-05-02
docs: explain how MCP servers actually launch (stdio vs SSE, Inspector)

### LINKED ISSUE #521: confusing documentation
I have an application that expects mcp to be on port 3001 over sse. the documentation tells me how to create the config.json alright, but then it goes into if I want to make a server, or a client, or this or that which I do not want to do. what I want to do is to run mcp so it's available at that port, but I'm not even sure what to run. the documentation doesn't even really go into anything useful. somehow I stumbled on the model inspector

image

it also seems to be able to talk to mcp on port 3001, but no dice. where is the binary? what do I run? where do I set the port? do I make a systemd service to start something? how does mcp run? in fact I don't even know where to put the config.json.

on the git page I see "For example, this will start the [Git](https://github.com/modelcontextprotocol/servers/blob/main/src/git) server:

# With uvx
uvx mcp-server-git" but running uvx mcp-server-git doesn't seem to do anything, there is no output, there are no ports opened.... it just sits there.

TanishkaMarrott:
## Summary

Closes #521 — addresses the recurring confusion around running MCP servers.

- Adds a **"Understanding How Servers Are Launched"** subsection inside Getting Started, between the standalone `npx`/`uvx` examples and the client config section
- Explains that stdio servers block on stdin by design — running one directly in a terminal will look like a hang, which is expected
- Clarifies that **the MCP client spawns the server**; the config file is the trigger, not a manual `uvx` command
- Adds a two-row table contrasting stdio vs SSE/HTTP transport so readers can see at a glance which mode applies to them
- Points to the MCP Inspector (`npx @modelcontextprotocol/inspector uvx mcp-server-git`) as the recommended way to test a server locally without wiring up a full client

No existing content changed. The new subsection slots in cleanly between the two existing subsections.

## Test plan

- [ ] Read the new subsection — confirm it answers the three questions in #521 (why the server appears to hang, who actually spawns the process, how to test without a full client)
- [ ] Verify the Inspector command renders correctly in GitHub markdown
- [ ] Confirm the table renders in GitHub markdown preview

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,27 @@ python -m mcp_server_git
 
 Follow [these](https://docs.astral.sh/uv/getting-started/installation/) instructions to install `uv` / `uvx` and [these](https://pip.pypa.io/en/stable/installation/) to install `pip`.
 
+### Understanding How Servers Are Launched
+
+MCP servers communicate over **stdio** (stdin/stdout) by default. If you run a server like `uvx mcp-server-git` directly in your terminal, it will appear to hang — it is waiting for a client to connect over stdin and send JSON-RPC messages. This is expected behaviour, not a bug.
+
+**You do not run MCP servers yourself.** Your MCP client (Claude Desktop, VS Code, Cursor, etc.) reads its config file and spawns the server process automatically. The client and server communicate over the spawned process's stdin/stdout. When you close the client, the server process stops.
+
+**Two transport modes:**
+
+| Transport | How the server runs | When to use |
+|---|---|---|
+| **stdio** (default) | Client spawns the server as a subprocess; communication is over stdin/stdout | Local servers — all reference servers in this repo |
+| **SSE / HTTP** | Server runs as a standalone HTTP service; client connects over a network URL | Remote or shared servers |
+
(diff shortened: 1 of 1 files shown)


alialobidm (commented):
``

TanishkaMarrott:
@cliffhall — this is a docs-only fix to README.md (21 lines added, no new servers). It closes #521 which has been open for a while. The CI failure is a permissions error on the bot's side, not an issue with the PR itself — the branch only touches README.md. Would appreciate a quick look when you have bandwidth!

MukundaKatta:
+1, this would have saved me time. I was just writing READMEs for some
stdio MCP servers and my own confidence-test was running `npx -y <package>`
in a terminal, then being confused by the silence for half a minute. The
two-row stdio vs SSE table is the right level. Tiny ask: an explicit
"if nothing prints, that's correct" line right above the standalone npx
example.

TanishkaMarrott:
Thanks @MukundaKatta — good call. I've added an explicit note right above the standalone `npx` example: 'If you run this command and see no output, that is correct — the server is waiting for a client to connect over stdin.' It also links down to the Understanding How Servers Are Launched section for full context.

PR #4104, opened 2026-05-05
fix(everything): remove dead code in resource template URI validation and subscription cleanup

1060996408:
## Description

Two small dead-code removals in `src/everything`'s resource layer. No behavioral change. 95 / 95 tests still pass.

## Server Details
- Server: `everything` (reference server)
- Changes to: resources

## Motivation and Context

Two unreachable branches in the resource handlers were doing more harm than good — once misread, they suggest invariants that don't actually hold.

### `resources/templates.ts` — `parseResourceId`

The guard

```ts
if (uri.toString().startsWith(textUriBase) && uri.toString().startsWith(blobUriBase)) {
  throw new Error(uriError);
}
```

is dead. `textUriBase` and `blobUriBase` are mutually exclusive prefixes, so the `&&` is never true. The SDK's template-based routing already guarantees the URI prefix is one of the two before the handler runs, so the surrounding "unknown URI" error path can't trigger either. Reduced to the only check that still does work — `resourceId` must be a positive integer.

### `resources/subscriptions.ts` — `sendSimulatedResourceUpdates`

```ts
if (subscribers.has(sessionId)) {
  await server.server.notification(...);
} else {
  subscribers.delete(sessionId); // subscriber has disconnected
}
```

The `else` deletes `sessionId` from `subscribers` whenever the session isn't in *one specific URI's* subscriber set. The comment claims the session has disconnected, but that doesn't follow: a session subscribed to URI A but not URI B will hit the `else` for URI B every iteration. And `Set.delete` of an absent element is a no-op, so the branch never had useful effect. Removed.

## How Has This Been Tested?

- `npm test` in `src/everything`: **95 / 95 pass**
- Spot-checked the `everything` server with an LLM client (Claude Code) — `read-resource`, subscribe / unsubscribe flows, and template-routed `text/{id}` and `blob/{id}` reads all behave as before.

## Breaking Changes

None.

## Types of changes
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] I have updated the server's README accordingly *(no user-facing change)*
- [x] I have documented all environment variables and configuration options *(none changed)*

## Additional context

- 1 commit, 2 files (`subscriptions.ts`, `templates.ts`), `+19 / −27` lines.
- No new dependencies, no API surface change.

### DIFF
--- a/src/everything/resources/subscriptions.ts
+++ b/src/everything/resources/subscriptions.ts
@@ -99,10 +99,9 @@ export const setSubscriptionHandlers = (server: McpServer) => {
 /**
  * Sends simulated resource update notifications to the subscribed client.
  *
- * This function iterates through all resource URIs stored in the subscriptions
- * and checks if the specified session ID is subscribed to them. If so, it sends
- * a notification through the provided server. If the session ID is no longer valid
- * (disconnected), it removes the session ID from the list of subscribers.
+ * Iterates the URIs in `subscriptions` and emits a
+ * `notifications/resources/updated` notification for each URI the given
+ * sessionId is subscribed to.
  *
  * @param {McpServer} server - The server instance used to send notifications.
  * @param {string | undefined} sessionId - The session ID of the client to check for subscriptions.
@@ -122,8 +121,6 @@ const sendSimulatedResourceUpdates = async (
         method: "notifications/resources/updated",
         params: { uri },
       });
-    } else {
-      subscribers.delete(sessionId); // subscriber has disconnected
     }
   }
 };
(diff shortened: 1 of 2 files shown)


1060996408:
Friendly ping — this PR has been open for ~2 weeks with all CI checks passing. It removes dead code in the everything server's resource template validation and subscription cleanup. Happy to address any feedback!

1060996408:
Hi! This PR has been open for about 2 weeks with all CI checks passing. Could a maintainer take a look when available? Happy to address any feedback. Thanks!

olaservo (approved):
Both claims verified against source.

- `subscriptions.ts`: `subscribers.delete(sessionId)` is in the `else` of `if (subscribers.has(sessionId))`, so it deletes from a Set that lacks the element. No-op.
- `templates.ts`: `textUriBase` is `${uriBase}/text` and `blobUriBase` is `${uriBase}/blob`. Neither prefixes the other, so `startsWith(a) && startsWith(b)` is never true.

PR #4107, opened 2026-05-05
fix(everything): make syncRoots idempotent on error and empty paths

1060996408:
## Description

`syncRoots` in `src/everything/server/roots.ts` claims (in JSDoc) to be idempotent and fetch roots from the client at most once per session, but the previous implementation only populated the cache on the success path. If `listRoots` threw or the client returned a falsy / unshaped response, no entry was written to the cache. Subsequent `syncRoots` calls would then re-issue `roots/list` to the client *and* re-register the `notifications/roots/list_changed` handler — once per tool invocation while the error persisted.

This change makes the cache-write semantics match the documented contract.

## Server Details
- Server: `everything` (reference server)
- Changes to: server/roots.ts (+ new __tests__/roots.test.ts)

## Motivation and Context

Two concrete issues with the old code:

1. **Repeated `listRoots` requests on transient client errors.** Every call to `get-roots-list` (or any caller of `syncRoots`) re-hit the client. With a flaky client this turns one user-initiated tool call into N silent client round-trips.
2. **Repeated `setNotificationHandler` registration.** `server.server.setNotificationHandler(RootsListChangedNotificationSchema, requestRoots)` is invoked on every retry. Whether this is a no-op (last-write-wins) or accumulates depends on SDK internals — either way it shouldn't be re-registered.

The JSDoc already promises "This function is idempotent. It should only request roots from the client once per session." The fix aligns the code with that promise rather than weakening the doc.

### What changed

In `requestRoots`:
- success path: unchanged (`roots.set(sessionId, response.roots)` + log).
- falsy / unshaped response: now caches an **empty array** so the session is marked synced.
- throw path: caches an empty array (only if no entry yet) so subsequent calls don't retry; still logs to stderr.

In `syncRoots`:
- early-return when the client lacks the `roots` capability (one less indent level for the rest).
- before `await requestRoots()`, write a placeholder `roots.set(sessionId, [])`. This guards against a re-entrant call landing while the initial fetch is still in flight (the second caller would otherwise see `!roots.has(sessionId)` and install a second notification handler).
- successful response replaces the placeholder.

Refresh continues to come from the `roots/list_changed` notification channel — that handler is registered exactly once per session.

## How Has This Been Tested?

New test file `src/everything/__tests__/roots.test.ts` (4 tests):

- caches roots and does not re-fetch on subsequent calls (success path; baseline)
- does not re-fetch when `listRoots` throws (before fix: 3 calls, after: 1)
- does not re-fetch when `listRoots` returns falsy / unshaped (before fix: 2 calls, after: 1)
- skips fetch entirely when the client lacks the `roots` capability

Full suite for `src/everything`: **99 / 99 pass** (95 existing + 4 new). `npm run build` clean.

Manual: re-ran `get-roots-list` against the local `everything` server with a working client; behavior unchanged on the success path.

## Breaking Changes

None. Internal cache semantics only; the public tool surface (`get-roots-list`) returns the same shape, including the empty-list-with-helper-text branch when the client has no roots configured.

## Types of changes
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update *(JSDoc updated alongside the behavioral fix)*

## Checklist
- [ ] I have updated the server's README accordingly *(no user-facing change)*
- [x] I have documented all environment variables and configuration options *(none changed)*

## Additional context

- 1 commit, 2 files (`server/roots.ts`, new `__tests__/roots.test.ts`), `+148 / −53`.
- No new dependencies, no protocol-level change.

### DIFF
--- a/src/everything/__tests__/roots.test.ts
+++ b/src/everything/__tests__/roots.test.ts
@@ -1,0 +1,87 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { syncRoots, roots } from '../server/roots.js';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+
+const sessionA = 'session-a';
+
+const makeMockServer = (overrides: {
+  listRoots: () => Promise<unknown>;
+  capabilities?: Record<string, unknown>;
+}) => {
+  const setNotificationHandler = vi.fn();
+  const sendLoggingMessage = vi.fn().mockResolvedValue(undefined);
+  const listRoots = vi.fn(overrides.listRoots);
+  const capabilities = overrides.capabilities ?? { roots: {} };
+
+  const mockServer = {
+    sendLoggingMessage,
+    server: {
+      getClientCapabilities: vi.fn(() => capabilities),
+      listRoots,
+      setNotificationHandler,
+    },
+  } as unknown as McpServer;
+
+  return { mockServer, listRoots, setNotificationHandler, sendLoggingMessage };
+};
+
+describe('syncRoots', () => {
+  beforeEach(() => {
+    roots.clear();
+  });
+
+  it('caches roots and does not re-fetch on subsequent calls', async () => {
+    const { mockServer, listRoots, setNotificationHandler } = makeMockServer({
+      listRoots: () => Promise.resolve({ roots: [{ uri: 'file:///a', name: 'a' }] }),
+    });
+
+    const first = await syncRoots(mockServer, sessionA);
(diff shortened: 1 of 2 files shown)


1060996408:
Friendly ping — this PR has been open for ~2 weeks with all CI checks passing. It makes syncRoots idempotent on error and handles empty paths gracefully. Happy to address any feedback!

1060996408:
Hi! This PR has been open for about 2 weeks with all CI checks passing. Could a maintainer take a look when available? Happy to address any feedback. Thanks!

PR #4115, opened 2026-05-06
fix(filesystem): preserve file permissions during write and edit operations

drudiger:
## Description

The filesystem server's atomic write pattern (write to temp file + `fs.rename()`) replaces the original file's inode, causing the new file to inherit default `0644` permissions from the temp file. This silently strips execute bits and any other non-default permissions from the original file.

This fix captures `stat.mode` before the write and restores it with `fs.chmod()` after the rename, in both `writeFileContent()` and `applyFileEdits()`.

## Server Details
- Server: filesystem
- Changes to: tools (`write_file`, `edit_file`)

## Motivation and Context

When using `edit_file` or `write_file` on an executable script (e.g., a shell script with `chmod +x`), the file loses its execute permission after the edit. This causes downstream failures when the script is invoked -- for example, a CI build calling `./pip_install.sh` fails with "Permission denied" after the MCP server edits the file.

The root cause is the atomic write security pattern: `fs.writeFile(tempPath)` creates a new file with the process's default umask (typically `0644`), then `fs.rename(tempPath, filePath)` atomically replaces the original inode. Since the rename replaces the inode entirely, the original file's permission bits are lost.

## How Has This Been Tested?

1. **Unit tests (4 new tests added, all 149 pass):**
   - `writeFileContent` preserves 755 permissions when overwriting existing file
   - `applyFileEdits` preserves 755 permissions after applying edits
   - `applyFileEdits` does NOT call chmod in dry run mode
   - Existing `applyFileEdits` tests updated with `stat`/`chmod` mocks in `beforeEach`

2. **Manual testing with MCP client:**
   - Created a file with `chmod 755`, edited it via the `edit_file` MCP tool, confirmed permissions remained `755`
   - Same test with `write_file` MCP tool, confirmed permissions preserved
   - Tested with a real Roo Code / VS Code MCP client session

## Breaking Changes

None. This is a bug fix that restores expected behavior. No configuration changes needed.

## Types of changes

## Checklist

Note: No README update is needed because this is a bug fix, not a new feature. No new environment variables or configuration options were introduced.

## Additional context

The fix adds `fs.stat()` before the atomic write and `fs.chmod()` after `fs.rename()` in two locations:
- `writeFileContent()` (line ~170 in lib.ts) -- used by the `write_file` tool
- `applyFileEdits()` (line ~272 in lib.ts) -- used by the `edit_file` tool

There is an inherent TOCTOU window between `stat` and `chmod` where another process could change the file's permissions, but this is the same class of race that exists for the content read in `applyFileEdits` and is not practically exploitable in this context.

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -308,6 +308,22 @@ describe('Lib Functions', () => {
         
         expect(mockFs.writeFile).toHaveBeenCalledWith('/test/file.txt', 'new content', { encoding: "utf-8", flag: 'wx' });
       });
+
+      it('preserves file permissions when overwriting existing file', async () => {
+        // First writeFile call with 'wx' flag fails because file exists
+        mockFs.writeFile.mockRejectedValueOnce(Object.assign(new Error('EEXIST'), { code: 'EEXIST' }));
+        // stat returns executable permissions
+        mockFs.stat.mockResolvedValueOnce({ mode: 0o100755 });
+        // Second writeFile (to temp) succeeds
+        mockFs.writeFile.mockResolvedValueOnce(undefined);
+        mockFs.rename.mockResolvedValueOnce(undefined);
+        mockFs.chmod.mockResolvedValueOnce(undefined);
+
+        await writeFileContent('/test/script.sh', 'new content');
+
+        expect(mockFs.stat).toHaveBeenCalledWith('/test/script.sh');
+        expect(mockFs.chmod).toHaveBeenCalledWith('/test/script.sh', 0o100755);
+      });
     });
 
   });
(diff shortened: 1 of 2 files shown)


olaservo (changes requested):
Right diagnosis: the temp-file + `fs.rename` pattern replaces the inode, so the file inherits the temp file's default `0644` and loses execute bits. Two changes before this can merge:

1. `fs.chmod(filePath, origStats.mode)` passes the full mode including file-type bits (`0o100755` rather than `0o755`). Linux masks to `mode & 07777` so it happens to work, but POSIX leaves those bits unspecified. Use `origStats.mode & 0o777`.

2. The `chmod` call sits inside the `try` whose `catch` unlinks the temp file and rethrows, but `rename` has already succeeded by then. A `chmod` failure (`EPERM` where the user can write the directory but does not own the file, restricted or network filesystems, Windows) makes the tool report the write as failed after the file was already modified. Give it its own try/catch that does not fail the operation.

The tests are fully mocked (`mockFs.stat.mockResolvedValueOnce({ mode: 0o100755 })`), so they assert `chmod` was called rather than that permissions survive. That is why neither of these surfaced in the suite.

olaservo:
Pushed the two changes from my review directly to this branch so it can move: mask the mode with `& 0o777`, and move `chmod` out of the rename try/catch so a chmod failure does not fail an already-completed write. Also added a test for the EPERM case. Will merge once CI is green.

PR #4128, opened 2026-05-09
docs: clarify stdio server behavior and debugging

### LINKED ISSUE #521: confusing documentation

zuhabul:
## Summary
- add a new README section explaining why running uvx mcp-server-git appears idle
- clarify stdio transport semantics (no port, no standalone daemon behavior)
- add a manual initialize command for quick liveness checks
- add MCP Inspector usage for interactive debugging

## Why
- addresses recurring confusion captured in #521
- makes first-time setup and debugging significantly clearer for contributors

Closes #521


MukundaKatta:
+1 to the manual initialize liveness check idea. The MCP Inspector pointer
is also gold for first-time setup. One small ask: if there's room, a
copy-pastable jq filter showing just `serverInfo.name + version` would help
new contributors confirm the handshake at a glance without parsing the
full JSON response.

PR #4156, opened 2026-05-14
fix(filesystem): include error reason + enumerate failed dirs at startup (#4152)

### LINKED ISSUE #4152: filesystem: server exits silently when an allowed_directories entry does not exist on disk
## Summary

The `secure-filesystem-server` package (`src/filesystem/`) terminates the Node process during startup if any path passed in `allowed_directories` is missing or inaccessible. The exit produces no stderr output, no JSON-RPC error response, and no log line indicating which path was the culprit — only the host (Claude Desktop, in my case) sees an abrupt transport close.

This is a sharp edge for end users: any folder rename, drive disconnect, or stale config entry takes the entire connector offline with no diagnostic to act on.

## Environment

- **Host:** Claude Desktop (Windows UWP, but the bug is in the server itself and is host-agnostic)
- **OS:** Windows 11
- **Filesystem connector extension ID:** `ant.dir.ant.anthropic.filesystem`
- **MCP server build:** ships with current Claude Desktop (Node-based, `secure-filesystem-server` v0.2.0, per server log line `"serverInfo":{"name":"secure-filesystem-server","version":"0.2.0"}`)

## Reproduction

1. Configure the filesystem MCP server with an `allowed_directories` list that contains at least one path which does not exist on disk. For example:

   ```json
   {
     "allowed_directories": [
       "C:\\PKM",
       "C:\\Dev",
       "C:\\DevBackup"
     ]
   }
   ```

   where `C:\PKM` and `C:\Dev` exist but `C:\DevBackup` has been renamed or deleted.

2. Launch the host so the MCP server is spawned.

3. Observe that the server starts, accepts the `initialize` handshake, and then exits within 1–2 seconds. No `tools/list` is ever served.

## Expected behaviour

One of:

- **Preferred:** the server starts successfully using the *valid* subset of `allowed_directories`, emits a structured warning (stderr or JSON-RPC `notifications/message`) naming each invalid path, and continues serving requests scoped to the surviving paths.
- **Acceptable:** the server fails fast but writes a clear stderr line identifying the offending path, e.g. `Allowed directory does not exist: C:\DevBackup — startup aborted`, and exits with a non-zero code that the host can surface.

Either makes the failure actionable.

## Actual behaviour

The server exits with no stderr output and no JSON-RPC error. From the host's perspective the connector simply disappears mid-handshake.

Verbatim from `%APPDATA%\Claude\logs\mcp-server-Filesystem.log` (timestamps trimmed):

```
[Filesystem] [info] Initializing server...
[Filesystem] [info] Using built-in Node.js for MCP server: Filesystem
[Filesystem] [info] Server started and connected successfully
[Filesystem] [info] Message from client: {"method":"initialize", ... "id":0}
[Filesystem] [info] Server transport closed
[Filesystem] [info] Server transport closed unexpectedly, this is likely due to the process exiting early.
  If you are developing this MCP server you can add output to stderr ... and it will appear in this log.
[Filesystem] [error] Server disconnected.
```

The host UI surfaces three opaque toasts: *"MCP Filesystem: Server disconnected"*, *"This isn't working right now"*, *"Could not attach to MCP server Filesystem"*. None mention which path is bad, which makes the issue hard to diagnose without reading server source.

## Suspected root cause

During startup the server resolves every entry in `allowed_directories` via something equivalent to `fs.realpathSync(path)` to canonicalise it for later path-prefix checks. When a path is missing this throws `ENOENT`. The throw is not caught at the validation site, so it propagates to top-level and Node terminates with no handler logging it.

## Suggested fix

Wrap per-path validation in try/catch and partition the input into "valid" and "invalid" buckets. Pseudocode:

```ts
const valid: string[] = [];
const invalid: { path: string; reason: string }[] = [];

for (const raw of allowedDirectories) {
  try {
    const resolved = fs.realpathSync(raw);
    const stat = fs.statSync(resolved);
    if (!stat.isDirectory()) {
      invalid.push({ path: raw, reason: 'not a directory' });
      continue;
    }
    valid.push(resolved);
  } catch (err: any) {
    invalid.push({ path: raw, reason: err?.code ?? err?.message ?? 'unknown' });
  }
}

for (const { path, reason } of invalid) {
  console.error(`[filesystem] Skipping allowed directory "${path}": ${reason}`);
}

if (valid.length === 0) {
  console.error('[filesystem] No valid allowed directories — aborting.');
  process.exit(1);
}

// continue startup with `valid`
```

This:

- Names every bad path on stderr, which the host displays in its server log.
- Keeps the connector alive for the valid subset (matches user intent in the common case where only one of several drives is offline).
- Aborts cleanly with a clear message if *every* path is bad.

An optional follow-up would be to emit a `notifications/message` MCP event after `initialize` so the host can render an inline warning in its UI, but the stderr line alone already unlocks 95% of the diagnostic value.

## Impact

This silent-failure mode wastes substantial debugging time. In my case I'd renamed a folder hours earlier and forgotten it was in the connector list; the toasts gave no indication that a path was the problem, so I spent ~30 minutes reading `mcp.log`, `main.log`, and the Filesystem server log before finally diffing the `allowed_directories` JSON against the actual filesystem.

I've also filed a corresponding feedback note with Anthropic about the Claude Desktop side of the UX (better host-side dialog when a connector fails). This issue covers only the upstream server-side change.

## Additional notes

- The same code path may affect other Node-based MCP servers in this repo that take a directory list as a startup argument — worth checking, though I haven't reproduced there.
- Happy to test a candidate fix on Windows + Claude Desktop UWP and confirm the new behaviour if a PR lands.

daveCode-dot:
## Summary

Closes #4152. The secure-filesystem-server's startup validation already filters inaccessible `allowed_directories` entries, but the diagnostic surface is too thin: per-directory warnings omit the underlying OS error and the aggregate fatal error doesn't enumerate which paths failed. Hosts that surface only the final stderr line (e.g. Claude Desktop in the original report) end up showing users an opaque "Server disconnected" toast with nothing to act on.

This change keeps the existing log strings stable for grep-based consumers and adds the missing reason / enumeration so the diagnostic is actionable.

## Repro

```bash
cd src/filesystem && npm run build

# 1) Single missing directory: silent-ish exit (no reason, no enumeration)
node dist/index.js /nonexistent/foo
# stderr (before): Warning: Cannot access directory /nonexistent/foo, skipping
#                  Error: None of the specified directories are accessible
# exit code 1, but a host showing only the final line cannot tell the user
# *which* directory was bad or *why*.

# 2) Mixed valid/invalid: only the warning line names the path, no reason
node dist/index.js /nonexistent/bar /tmp
# stderr (before): Warning: Cannot access directory /nonexistent/bar, skipping
```

## Fix

`src/filesystem/index.ts` — startup validation now tracks failed entries with their reason and:

- Appends the underlying error message to the per-directory warning, e.g. `Cannot access directory /missing, skipping (ENOENT: no such file or directory, stat '/missing')`.
- Enumerates every failed entry inside the aggregate fatal error so a host capturing only the final stderr line gets the full list:

```
Error: None of the specified directories are accessible:
  - /nonexistent/foo (ENOENT: no such file or directory, stat '/nonexistent/foo')
  - /tmp/secret.txt (not a directory)
```

The four pre-existing `startup-validation.test.ts` cases continue to pass unchanged — the legacy `Warning: Cannot access directory` and `Error: None of the specified directories are accessible` substrings are preserved.

## Tests

`src/filesystem/__tests__/startup-validation.test.ts` adds two regression cases pinned to #4152:

- `should include the underlying error reason in the per-directory warning (#4152)` — asserts `/ENOENT|no such file or directory/i` shows up in stderr alongside the path.
- `should list every offending directory inside the aggregate startup error (#4152)` — slices stderr from the `Error:` marker and verifies *both* failing paths appear in the aggregate, not only as scattered earlier warnings.

Local result, ran from `src/filesystem`:

```
npx vitest run
Test Files  7 passed (7)
     Tests  148 passed (148)
```

No regression in the other six test files (`directory-tree`, `lib`, `path-utils`, `path-validation`, `roots-utils`, `structured-content`).

## Adjacent gaps (out of scope · raising for maintainer decision)

While auditing this code path I noticed `roots-utils.ts` already exposes a `formatDirectoryError(dir, error?, reason?)` helper used by `getValidRootDirectories()`. There are now two error-formatting styles for the same domain:

- `index.ts` startup: `Cannot access directory <dir>, skipping (<reason>)` and `- <dir> (<reason>)` enumeration
- `roots-utils.ts` MCP roots: `Skipping invalid directory: <dir> due to error: <message>`

Standardising on a single format (either by exporting `formatDirectoryError` and adopting it in `index.ts`, or vice-versa) would remove this drift, but it changes user-facing log strings and the right shape is a maintainer call rather than a unilateral choice in this PR. Happy to follow up with a separate PR once you indicate which way you'd like to consolidate.

## Notes

- No `package.json` / version bump — leaving release cadence to the maintainers.
- No new dependencies.
- The `failedEntries` list is built only during startup so there is no runtime cost on the hot path.

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/filesystem/__tests__/startup-validation.test.ts
+++ b/src/filesystem/__tests__/startup-validation.test.ts
@@ -97,4 +97,44 @@ describe('Startup Directory Validation', () => {
     // Should still start with the valid directory
     expect(result.stderr).toContain('Secure MCP Filesystem Server running on stdio');
   });
+
+  // Regression coverage for #4152 — warnings about inaccessible startup
+  // directories must include the underlying error reason so hosts can
+  // surface an actionable diagnostic to end users (the original report
+  // saw a silent transport close because nothing identified the bad path
+  // beyond a generic "Cannot access directory" line).
+  it('should include the underlying error reason in the per-directory warning (#4152)', async () => {
+    const nonExistentDir = path.join(testDir, 'non-existent-dir-with-reason');
+
+    const result = await spawnServer([nonExistentDir, accessibleDir]);
+
+    // Should still skip + continue with the valid directory.
+    expect(result.stderr).toContain('Secure MCP Filesystem Server running on stdio');
+    // Diagnostic must name the directory AND the OS-level error reason.
+    expect(result.stderr).toContain(nonExistentDir);
+    expect(result.stderr).toMatch(/ENOENT|no such file or directory/i);
+  });
+
+  // Regression coverage for #4152 — when ALL startup directories fail, the
+  // aggregate exit error must enumerate each offending path AFTER the
+  // "Error:" marker (not only as scattered per-directory warnings) so a
+  // host capturing only the final fatal line can show users every entry of
+  // allowed_directories that needs fixing.
+  it('should list every offending directory inside the aggregate startup error (#4152)', async () => {
+    const bad1 = path.join(testDir, 'aggregate-fail-1');
+    const bad2 = path.join(testDir, 'aggregate-fail-2');
+
+    const result = await spawnServer([bad1, bad2]);
+
+    expect(result.exitCode).toBe(1);
+
+    // Slice from the "Error:" marker to end-of-stderr — the aggregate
+    // message itself must contain both paths, not only the per-directory
+    // warnings that precede it.
+    const errIdx = result.stderr.indexOf('Error:');
+    expect(errIdx, 'aggregate Error: line missing').toBeGreaterThanOrEqual(0);
+    const aggregate = result.stderr.slice(errIdx);
+    expect(aggregate).toContain(bad1);
+    expect(aggregate).toContain(bad2);
+  });
 });
--- a/src/filesystem/__tests__/startup-validation.test.ts
+++ b/src/filesystem/__tests__/startup-validation.test.ts
@@ -115,6 +115,35 @@ describe('Startup Directory Validation', () => {
     expect(result.stderr).toMatch(/ENOENT|no such file or directory/i);
   });
 
+  // Regression coverage for #4152 — permission-denied (EACCES) directories
+  // must surface the OS reason in the warning the same way ENOENT does. The
+  // original report only mentioned non-existent dirs, but the underlying
+  // try/catch handles every fs.stat failure and hosts on Windows / locked
+  // mounts hit EACCES just as often. Skipped on POSIX root (no chmod 0o000
+  // restriction) and on Windows (chmod has no effect on directory access).
+  const skipChmodEacces = process.platform === 'win32' || (typeof process.getuid === 'function' && process.getuid() === 0);
+  (skipChmodEacces ? it.skip : it)('should include EACCES reason for permission-denied directory (#4152)', async () => {
+    const lockedDir = path.join(testDir, 'locked-dir');
+    await fs.mkdir(lockedDir);
+    // Drop *all* perms on the parent so fs.stat on a child path returns EACCES.
+    const parent = path.join(testDir, 'no-traverse');
+    await fs.mkdir(parent);
+    const child = path.join(parent, 'inner');
+    await fs.mkdir(child);
+    await fs.chmod(parent, 0o000);
+
+    try {
+      const result = await spawnServer([child, accessibleDir]);
+      // Should still continue with the accessible one.
+      expect(result.stderr).toContain('Secure MCP Filesystem Server running on stdio');
+      expect(result.stderr).toContain(child);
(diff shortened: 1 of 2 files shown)


BossChaos:
## Code Review

**PR:** [fix(filesystem): include error reason + enumerate failed dirs at startup (#4152)](https://github.com/modelcontextprotocol/servers/pull/4156) by @daveCode-dot

- ✅ Bug/security fix

*Code review for MCP servers*

PR #4163, opened 2026-05-14
fix(filesystem): reject non-media read_media_file inputs

### LINKED ISSUE #4029: filesystem: read_media_file returns invalid 'blob' content type, violating MCP spec

Lubrsy706:
## Summary
- remove the invalid `blob` content type from the filesystem `read_media_file` output schema
- reject unsupported extensions before reading and encoding the file
- add integration coverage for valid image output and non-media rejection

Fixes #4029.

## Tests
- `npm test -- structured-content`
- `npm test`
- `npm run build`

### DIFF
--- a/src/filesystem/__tests__/structured-content.test.ts
+++ b/src/filesystem/__tests__/structured-content.test.ts
@@ -25,6 +25,13 @@ describe('structuredContent schema compliance', () => {
 
     // Create test files
     await fs.writeFile(path.join(testDir, 'test.txt'), 'test content');
+    await fs.writeFile(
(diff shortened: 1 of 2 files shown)


BossChaos:
## Code Review

**PR:** [fix(filesystem): reject non-media read_media_file inputs](https://github.com/modelcontextprotocol/servers/pull/4163) by @Lubrsy706

- ✅ Bug/security fix

*Code review for MCP servers*

PR #4175, opened 2026-05-16
fix(filesystem): ignore final newline in tail output

Sean-Kenneth-Doherty:
## Summary

- Fix `tailFile` so a normal final newline is not returned as an empty trailing line.
- Add regression coverage that asserts `tailFile` content instead of only checking handle cleanup.
- Keep intentional blank lines intact by removing only the split sentinel from the EOF chunk.

## Root Cause

`tailFile` splits the final chunk on `\n`. For text files ending in a newline, `split("\n")` produces a final empty sentinel, which the reverse reader counted as the last line. That made `read_text_file` with `tail: 1` return an empty string for ordinary POSIX-style files like `line1\nline2\nline3\n`.

## Validation

- `npm exec -w src/filesystem -- vitest run __tests__/lib.test.ts`
- `npm run build -w src/filesystem`

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -553,6 +553,23 @@ describe('Lib Functions', () => {
     });
 
     describe('tailFile', () => {
+      const mockTailReadableFile = (content: string) => {
+        const fileBuffer = Buffer.from(content, 'utf-8');
+        mockFs.stat.mockResolvedValue({ size: fileBuffer.length } as any);
+
+        const mockFileHandle = {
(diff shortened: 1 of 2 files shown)


Sean-Kenneth-Doherty:
Marked ready after rechecking the patch locally.

Why this is ready:

- the fix is scoped to `tailFile` only
- it preserves intentionally blank lines and removes only the EOF split sentinel from the final chunk
- regression tests cover `tail: 1` and `tail: 2` on ordinary newline-terminated text

Validation:

- `npm exec -w src/filesystem -- vitest run __tests__/lib.test.ts` - 46 passed
- `npm run build -w src/filesystem` - passed
- hosted Python/TypeScript test and build checks are passing

PR #4186, opened 2026-05-17
fix(filesystem): preserve UTF-8 multibyte characters across chunk boundaries in head/tail

ishidakei:
## Description

`read_text_file` with `head` or `tail` corrupts UTF-8 multibyte characters that straddle the 1024-byte chunk read boundary. The corrupted bytes become `U+FFFD` (replacement character). Reading the same file in full mode (no `head`/`tail`) is unaffected.

## Server Details
- Server: filesystem
- Changes to: `lib.ts` (`headFile`, `tailFile`)

## Motivation and Context

Both `headFile` and `tailFile` read the file as raw byte chunks via `fileHandle.read(chunk, 0, size, position)` and then call `.toString('utf-8')` on each chunk independently before concatenating the resulting strings. UTF-8 multi-byte sequences (e.g. CJK characters at 3 bytes each, most non-Latin scripts at 2–4 bytes) that span a chunk boundary are split mid-sequence, so each chunk decodes its partial bytes as `U+FFFD` and the original character is lost.

Reproduction (default 1024-byte chunk):
- File: 1023 bytes of ASCII, then `あ` (UTF-8: `E3 81 82`), then a newline.
- `head: 1` returns ASCII followed by `���` instead of `あ`.

The same mechanism breaks `tail` symmetrically: any multi-byte character spanning a 1024-byte boundary measured from EOF is replaced with `U+FFFD`.

This impacts every script outside Basic Latin in practice — Japanese/Chinese/Korean, Cyrillic, Greek, Arabic, accented Latin (German umlauts, French/Spanish accents, etc.), and emoji.

## Changes

- `headFile`: use `node:string_decoder` `StringDecoder` so trailing partial bytes from one read are held internally and combined with the leading bytes of the next read. `decoder.end()` is called on EOF to flush any genuinely truncated sequence.
- `tailFile`: accumulate raw bytes into a `Buffer[]` in iteration order and decode the concatenated buffer once at the end. Newline counting is performed at the byte level (safe because `0x0A` never appears inside a UTF-8 multi-byte sequence by design). The discarded partial first line absorbs any `U+FFFD` introduced at the lowest read position.

Behavior on ASCII input is unchanged. Memory characteristics of `tailFile` are unchanged (chunks accumulated correspond to the same range of bytes the original read).

## How Has This Been Tested?

- Added vitest tests in `src/filesystem/__tests__/encoding.test.ts` using real temp file I/O (no `fs` mocking) that place 3-byte (`あ`) and 4-byte (`😀`) UTF-8 characters at exact 1024-byte chunk boundaries for both head and tail directions. The new tests fail on `main` and pass with this change.
- Existing `lib.test.ts` and other suites continue to pass — full local run shows 156 passed across 8 test files.
- Originally observed via Claude.ai with Filesystem MCP reading a Markdown file containing Japanese kanji: full-file reads were clean while `read_text_file` with `head` or `tail` returned `U+FFFD` at multiples of 1024 bytes from the relevant end.

## Breaking Changes

None. The fix only affects previously incorrect output for non-ASCII input.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have updated the server's README accordingly  *(no README change needed — `read_text_file` already documents UTF-8 support; this fix brings behavior in line with documentation)*
- [x] I have documented all environment variables and configuration options *(N/A)*

### DIFF
--- a/src/filesystem/__tests__/encoding.test.ts
+++ b/src/filesystem/__tests__/encoding.test.ts
@@ -1,0 +1,140 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import fs from 'fs/promises';
+import path from 'path';
+import os from 'os';
+import { headFile, tailFile } from '../lib.js';
+
+// These tests use real file I/O (no fs mocking) because the bug exists in the
+// interaction between byte-level reads and UTF-8 decoding. The implementation
+// reads in 1024-byte chunks, so we construct files where a UTF-8 multi-byte
+// character is placed to straddle a chunk boundary at exact byte offsets.
+
+describe('UTF-8 multi-byte character handling at chunk boundaries', () => {
+  let tmpDir: string;
+
+  beforeEach(async () => {
+    tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), 'mcp-fs-encoding-'));
+  });
+
+  afterEach(async () => {
+    await fs.rm(tmpDir, { recursive: true, force: true });
+  });
+
+  describe('headFile', () => {
+    it('preserves a 3-byte UTF-8 character split across the 1024-byte boundary', async () => {
+      // 'あ' (U+3042) is E3 81 82 in UTF-8.
+      // 1023 'a' chars (bytes 0..1022) + 'あ' (bytes 1023..1025) + '\n' + 'second\n'.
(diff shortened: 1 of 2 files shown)


BossChaos:
## Code Review

**PR:** [fix(filesystem): preserve UTF-8 multibyte characters across chunk boundaries in head/tail](https://github.com/modelcontextprotocol/servers/pull/4186) by @ishidakei

- ✅ Bug fix / error handling

*Reviewing under MCP ecosystem*

Issue #4208, opened 2026-05-19
filesystem: FS_SEARCH_EXCLUDE_PREFIXES is lexical; should excludes be canonical/symlink-aware?


JonoGitty:
**Context.** The proposed fix for #4162 (PR #4212) adds `FS_SEARCH_EXCLUDE_PREFIXES` as a fast lexical prefilter that refuses recursive `search_files` / `directory_tree` on configured slow roots, typically macOS CloudStorage / FileProvider paths.

Discussion on #4162 surfaced that the check is purely lexical: a symlink such as `/allowed/sneaky` whose target lives under an excluded prefix is not caught by the prefix check, and may still reach the slow provider path during canonicalisation / validation.

The case for keeping the #4162 fix scoped to the lexical prefilter, as discussed in that thread:

- The bypass is pre-existing behaviour — the original `searchFilesWithValidation` exclude check is also lexical.
- `validatePath` continues to enforce the `allowed_directories` escape boundary via its own `realpath`, so the security boundary is unaffected. This question is about whether excludes should be symlink-aware.
- The impact of the bypass is performance / hang risk from reaching a slow provider path, not an authorization bypass.
- An unbounded `realpath`-before-compare exclude check risks re-introducing the kind of hang #4162 is trying to avoid.

**Design question.** Should excludes be hardened against symlink bypass?

One possible shape is two-stage:

1. Fast lexical check first, preserving current cheap behaviour.
2. If not excluded lexically, optionally perform bounded canonicalisation, e.g. `withFsTimeout(fs.realpath(...))`, then re-check the canonical path against excluded prefixes, with an explicit fail-policy on timeout — deny, allow, or return a typed error.

This affects both `search_files` and `directory_tree`, so it seems like a design decision rather than a drop-in fix. Not gating #4162; filing this separately so the symlink-aware exclude question is not lost.

norika1207-lab:
This looks worth pinning down as a protocol-level regression rather than only an implementation bug.

For `filesystem search excludes should be canonical or symlink-aware`, the acceptance test I would want is a minimal client/server fixture that records the exact JSON-RPC or HTTP exchange, then asserts both the wire result and the local lifecycle state after the failure path. In MCP deployments, the dangerous failures are often not just wrong return values; they are pending requests, dropped notifications, mismatched session state, or errors that get translated into a generic response and become impossible to diagnose.

[...]

Issue #4162, opened 2026-05-14
filesystem: recursive search can hang on macOS CloudStorage / lazy provider paths


alemuenchen:
Hi — thanks for maintaining the filesystem server.
  
  I'm filing this against `@modelcontextprotocol/server-filesystem` after several weeks of production use on macOS with provider-backed directories in `allowedDirectories`,
  especially:

  - `~/Library/CloudStorage/GoogleDrive-.../...`
  - `~/Library/Mobile Documents/com~apple~CloudDocs/...`

  ### Symptom

  When `allowedDirectories` includes macOS CloudStorage / lazy provider paths, `search_files` and other operations that call `validatePath()` can intermittently hang for minutes.
  Claude Desktop eventually reports the MCP server as unresponsive after roughly 4 minutes. The subprocess remains alive, but subsequent filesystem calls may continue to hang until
   the server is restarted.
  
  This may overlap with reports like #3281, but I'm filing separately because this issue points to specific filesystem code paths.

  ### Code paths

  In [`src/filesystem/lib.ts`](https://github.com/modelcontextprotocol/servers/blob/main/src/filesystem/lib.ts), `validatePath()` calls:

  ```ts
  const realPath = await fs.realpath(absolute);
  ```

  and, in the ENOENT branch:

  ```ts
  const realParentPath = await fs.realpath(parentDir);
  ```

  These calls have no timeout. On macOS provider-backed paths that are not locally materialized, `realpath` can block while the provider fetches metadata.

  `searchFilesWithValidation()` then recursively does:

  ```ts
  const entries = await fs.readdir(currentPath, { withFileTypes: true });
  ...
  await validatePath(fullPath);
  ...
  if (entry.isDirectory()) {
    await search(fullPath);
  }
  ```

  As far as I can tell, recursive search currently has no upper bound on visited entries, no timeout around `readdir`, and no abort path. A large or lazy-materialized tree can
  therefore keep the server busy for a long time.

  ### Why this matters

  This is especially visible with CloudStorage / FileProvider-backed folders such as Google Drive and iCloud Drive, where metadata for subdirectories may be fetched lazily. It can
  also affect very large local trees.

  The libuv threadpool angle may make the symptom worse: filesystem operations use libuv's threadpool, whose default size is 4. A small number of slow filesystem calls can
  therefore degrade unrelated filesystem operations in the same process. I mention this as a contributing factor, not necessarily as the main fix.

  ### Suggested mitigations

[...]

PR #4212, opened 2026-05-20
fix(filesystem): timeouts + max-visited cap for hangs on lazy provider paths (#4162)

JonoGitty:
## Description

Bounds two unbounded code paths in the filesystem server that can hang for minutes on macOS CloudStorage / FileProvider-backed directories: `fs.realpath` in `validatePath`, and recursive `fs.readdir` in `search_files` / `directory_tree`. Addresses the structural concerns from #4162 that #4181 (excludePatterns before realpath) left open.

Follow-up design question about symlink-aware excludes is tracked separately in #4208 — intentionally not in scope here.

## Server Details
- Server: filesystem
- Changes to: tools (`search_files`, `directory_tree`, internal `validatePath`)

## Motivation and Context

`fs.realpath` and `fs.readdir` are called with no timeout. On macOS provider-backed paths (`~/Library/CloudStorage/...`, `~/Library/Mobile Documents/com~apple~CloudDocs/...`, and NFS/SMB), these can block while the provider fetches metadata; recursive search additionally has no upper bound on entries visited. The host (e.g. Claude Desktop) eventually reports the connector unresponsive (~4 min) while the subprocess stays alive but stuck. See #4162 for the full report.

The defensive shape here originated from @alemuenchen, who has run equivalent mitigations in production against Google Drive + iCloud Drive for several weeks and contributed the reference diff on #4162.

### What this adds

`src/filesystem/lib.ts`:
- `withFsTimeout()` wraps `fs.realpath` (in `validatePath`) and `fs.readdir` (in `searchFilesWithValidation`); on timeout it rejects with a typed `FsTimeoutError` rather than hanging.
- `FS_SEARCH_MAX_VISITED` caps recursive search; abort produces a clear "narrow the path / refine the pattern" error.
- `FS_SEARCH_EXCLUDE_PREFIXES` (opt-in, empty default) refuses recursive operations on configured slow roots, via the shared, boundary-aware `assertSearchPathNotExcluded` helper (reuses `isPathWithinAllowedDirectories`).
- ENOENT branch in `validatePath` refactored so a hung parent realpath surfaces as `FsTimeoutError` instead of being masked as "Parent directory does not exist".
- The per-entry `catch` in search re-throws `FsTimeoutError` so a search never returns silently-partial results.

`src/filesystem/index.ts`:
- `directory_tree` gets the same timeout + visited-cap treatment and shares the exclude-prefix helper.

### Configuration

| Variable | Default | Purpose |
|---|---|---|
| `FS_OP_TIMEOUT_MS` | `15000` | Timeout (ms) per `fs.realpath` / `fs.readdir`. |
| `FS_SEARCH_MAX_VISITED` | `50000` | Cap on entries visited by a single `search_files` / `directory_tree` before aborting. |
| `FS_SEARCH_EXCLUDE_PREFIXES` | (empty) | Comma-separated path prefixes recursive ops refuse outright. Lexical prefilter (see #4208 / README). |

Invalid numeric values warn to stderr and fall back to the default, so a typo cannot silently disable a guard.

## How Has This Been Tested?

- New `__tests__/4162-timeouts-and-caps.test.ts` (14 cases): realpath/readdir timeouts, ENOENT preservation + ENOENT-realpath-timeout regression, per-entry `FsTimeoutError` propagation, max-visited abort, exclude-prefix containment (descendant + boundary-aware sibling), env-var validation. Uses vitest fake timers + mocked `fs/promises`, so no real CloudStorage path is needed in CI.
- Full suite: **160/160 pass**, clean build, `lib.ts` coverage 82% stmts / 88% branch / 94% funcs.
- @alemuenchen independently pulled the branch and tested against a **real macOS Google Drive tree** with a live MCP client: no regression on pattern matching, cap + exclude-prefix behaviour confirmed, timeout does not misfire on warm provider paths.

## Breaking Changes

None. All new behaviour is bounded by env vars that default to safe values for ordinary local trees; existing configurations are unaffected.

## Types of changes
- [x] Documentation update

## Checklist
- [x] I have tested this with an LLM client (via @alemuenchen on a live macOS CloudStorage setup; I verified locally with unit tests + build)
- [x] New and existing tests pass locally (160/160)

## Additional context

Deliberately out of scope (worth their own follow-ups, flagged so as not to muddy this PR):
- **libuv threadpool** — timed-out `fs` calls still occupy the threadpool until they settle; a concurrency limiter is a separate mitigation.
- **Single huge directory** — the cap bounds breadth/depth but one `readdir` on a directory with millions of entries is still a single allocation; `fs.opendir` streaming would address this.
- **Symlink-aware excludes** — `FS_SEARCH_EXCLUDE_PREFIXES` is a lexical prefilter; a symlink whose target is under an excluded prefix is not blocked. Design discussion in #4208.

Credit: the core defensive diff is @alemuenchen's, contributed on #4162 with his blessing to take it forward. Tests, `directory_tree` integration, the shared exclude helper, README, and this PR are mine.

### DIFF
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -26,6 +26,11 @@ import {
   tailFile,
   headFile,
   setAllowedDirectories,
+  // #4162: timeout + cap primitives reused by directory_tree
+  withFsTimeout,
+  FS_SEARCH_MAX_VISITED,
+  FsSearchTruncatedError,
+  assertSearchPathNotExcluded,
 } from './lib.js';
 
 // Command line argument parsing
@@ -548,12 +553,38 @@ server.registerTool(
     }
     const rootPath = args.path;
 
+    // #4162: same defensive shape as searchFilesWithValidation -- honour
+    // FS_SEARCH_EXCLUDE_PREFIXES, bound the recursion, and timeout each
+    // readdir so a lazy provider-backed subtree can't keep the server busy
+    // indefinitely. validatePath already wraps fs.realpath via withFsTimeout
+    // in lib.ts, so only readdir + the entry counter need handling here. We
+    // throw immediately when the cap is hit rather than threading an
+    // `aborted` flag through unwinding recursion -- no caller wants a partial
+    // tree, and the immediate throw makes that contract obvious to future
+    // refactors.
+    assertSearchPathNotExcluded(rootPath, 'directory_tree');
+    let visited = 0;
+
     async function buildTree(currentPath: string, excludePatterns: string[] = []): Promise<TreeEntry[]> {
       const validPath = await validatePath(currentPath);
-      const entries = await fs.readdir(validPath, { withFileTypes: true });
+      const entries = await withFsTimeout(
+        fs.readdir(validPath, { withFileTypes: true }),
+        `readdir ${validPath}`,
+      );
       const result: TreeEntry[] = [];
 
       for (const entry of entries) {
+        if (visited >= FS_SEARCH_MAX_VISITED) {
+          throw new FsSearchTruncatedError(
+            `directory_tree aborted after visiting ${visited} entries ` +
+            `(cap FS_SEARCH_MAX_VISITED=${FS_SEARCH_MAX_VISITED}). ` +
+            `Narrow the search path or use excludePatterns.`,
+            visited,
+            FS_SEARCH_MAX_VISITED,
+          );
+        }
+        visited++;
+
         const relativePath = path.relative(rootPath, path.join(currentPath, entry.name));
         const shouldExclude = excludePatterns.some(pattern => {
           if (pattern.includes('*')) {
--- a/src/filesystem/README.md
+++ b/src/filesystem/README.md
@@ -204,6 +204,22 @@ The mapping for filesystem tools is:
 
 > Note: `idempotentHint` and `destructiveHint` are meaningful only when `readOnlyHint` is `false`, as defined by the MCP spec.
 
+## Environment variables
+
+Three environment variables guard against hangs when `allowed_directories` includes a slow or lazy provider-backed path (macOS `~/Library/CloudStorage/...`, `~/Library/Mobile Documents/com~apple~CloudDocs/...`, NFS / SMB mounts, etc.). Defaults are safe for ordinary local trees; tune only if needed.
+
+| Variable | Default | Purpose |
+|---|---|---|
+| `FS_OP_TIMEOUT_MS` | `15000` | Timeout (milliseconds) applied to each `fs.realpath` (in `validatePath`) and `fs.readdir` (in `search_files` and `directory_tree`). On timeout the call rejects with `FsTimeoutError` instead of hanging. |
+| `FS_SEARCH_MAX_VISITED` | `50000` | Hard cap on entries visited by a single `search_files` or `directory_tree` call before aborting with `"Search aborted after visiting N entries..."`. Prevents an unbounded recursion on a large or lazy-materialized tree. |
+| `FS_SEARCH_EXCLUDE_PREFIXES` | (empty) | Comma-separated path prefixes for which recursive search and `directory_tree` are refused outright. `~` is expanded. Matching is boundary-aware: a configured `/data/foo` will not match an unrelated `/data/foobar`. Intended for known-slow roots you'd rather forbid than time out on. |
+
+Invalid numeric values (`"15s"`, `"0"`, negative) print a warning to stderr and fall back to the default, so a typo cannot silently disable a guard.
+
+**Remote-mount tuning:** the 15 s default works well for local trees and warm CloudStorage. On NFS, SMB, or other remote mounts where individual `readdir` calls legitimately take seconds, consider raising `FS_OP_TIMEOUT_MS` (e.g. `30000`–`60000`) to avoid spurious timeouts.
+
+**`FS_SEARCH_EXCLUDE_PREFIXES` is a fast lexical *prefilter*, not a canonical-path exclusion.** The requested root is compared (after `~` expansion and normalisation) against the configured prefixes, with no `realpath` step. A symlink whose target lives under an excluded prefix is therefore **not** blocked — the request would pass the lexical exclude and may still hit the slow provider path through `validatePath`'s realpath. Treat this as a best-effort guard against direct submissions of slow paths, not as a complete symlink-aware exclusion. Security against paths escaping `allowed_directories` is enforced separately by `validatePath`.
+
 ## Usage with Claude Desktop
 Add this to your `claude_desktop_config.json`:
 
--- a/src/filesystem/__tests__/4162-timeouts-and-caps.test.ts
+++ b/src/filesystem/__tests__/4162-timeouts-and-caps.test.ts
@@ -1,0 +1,315 @@
+// Regression tests for issue #4162: recursive search / validatePath can hang
+// for minutes on macOS CloudStorage / lazy provider-backed paths. The patch
+// adds:
+//   - a timeout wrapper around fs.realpath in validatePath
+//   - a timeout wrapper around fs.readdir in searchFilesWithValidation
+//   - a max-visited-entries cap on recursive search
+//   - boundary-aware FS_SEARCH_EXCLUDE_PREFIXES that refuses recursive search
+//     outright on user-configured roots
+//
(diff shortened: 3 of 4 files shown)


temurkhan13:
The `FsTimeoutError` design here is the right shape: a distinct, typed error so the caller can tell a hang apart from `EACCES` (per the comment on the class) instead of swallowing it as an opaque fs failure. The restructure in `validatePath` reinforces it well, by pulling `realParentPath` out of the inner try and explicitly re-throwing `FsTimeoutError` from `parentError` so a timeout cannot get masked into a misleading "Parent directory does not exist." That kind of explicit error-shape discipline at the layer where the failure is still structured is what makes the rest of the PR's defensive shape pay off.

The one place the discipline gets dropped is on the cap side of `searchFilesWithValidation`. When `visited >= FS_SEARCH_MAX_VISITED`, the inner closure sets `aborted = true; return;`, and the outer function returns the accumulated `results` array unchanged. From the caller's perspective, a search that found 30 matches by visiting 49,999 entries and a search that aborted at the 50,000-entry cap with 30 partial matches look identical. The cap-reached signal exists inside `searchFilesWithValidation` (the `aborted` flag, the `visited` counter) but is not visible to anything outside it. A user asking "did this miss results?" cannot answer that question from the return value.

This is the same problem the timeout side carefully avoids. `FsTimeoutError` is the explicit "I bailed out, not a clean empty result" signal for the timeout path. There is no parallel signal for the cap path. The behavior gap is asymmetric: a hang surfaces as a typed error and is impossible to silently lose; a 50k-entry cap collapses into a partial result that looks complete.

Two shapes that would close it without disturbing the rest of the design:

1. Throw on cap, same as timeout: introduce `FsSearchTruncatedError` with the visit count and the directory where the cap was hit, mirroring `FsTimeoutError`. Simplest for callers that already need to handle `FsTimeoutError`; they catch both as "search did not complete."

2. Return a structured result instead of `string[]`: `{ paths: string[]; truncated: boolean; visited: number }`. Callers that want partial results get them, and callers that need completeness can branch on `truncated`. Wider API change, but more recoverable for the search-UI case where partial-with-warning is genuinely useful.

The throw shape is consistent with what the PR already does for timeouts and is probably the lower-friction choice if you want to keep the public signature stable. Either way, the asymmetric shape is the structural thing worth fixing here, because the typed-error discipline elsewhere in this diff is exactly what makes the silent-partial-result on the cap side stand out.

Worth confirming on the test file too. The new `4162-timeouts-and-caps.test.ts` covers cap-reached behavior. If the assertion is "result has &lt;= N entries when cap is N," the silent-truncation case is currently passing (correctly bounded, but indistinguishable from "no more matches existed"). A test that asserts the cap-reached signal is observable from outside the function would catch any future regression that silently re-introduces the asymmetry.

alemuenchen:
@temurkhan13 — thanks for taking the time to read through this so carefully, the framing around `FsTimeoutError` discipline is well put.

There's one piece I'd gently flag, because I think it might be easy to miss in the diff — the cap path in `searchFilesWithValidation` actually does surface a cap-reached signal, just not where the inner closure sets the flag.

The flow is:

1. The inner closure sets `aborted = true; return;` and unwinds the recursion fast (each level re-checks `aborted` on re-entry, by design).
2. After `await search(rootPath)` returns, **before** `return results`, there's:

```ts
if (aborted) {
  throw new Error(
    `Search aborted after visiting ${visited} entries ` +
    `(cap FS_SEARCH_MAX_VISITED=${FS_SEARCH_MAX_VISITED}). ` +
    `Refine the pattern (e.g. '**/name') or narrow the search path.`
  );
}
return results;
```

So a search that hits the cap throws with a clear message including the visit count — the caller doesn't see a silent partial array. `directory_tree` in `index.ts` is even more direct: it throws inline the moment `visited >= FS_SEARCH_MAX_VISITED`, before incrementing.

On the test side, the cap-reached signal is asserted from outside the function:

```ts
it('aborts with a clear error once the visited cap is exceeded', async () => {
  ...
  await expect(searchFilesWithValidation(ROOT, '**/*', ALLOWED))
    .rejects.toThrow(/Search aborted after visiting/);
});
```

That said, there's a real point in your comment that I do think is worth raising as a separate question for @JonoGitty / the maintainers: the timeout path uses a *typed* `FsTimeoutError`, while the cap path uses a plain `Error`. For callers doing `instanceof` discrimination rather than string-matching on the message, a typed `FsSearchTruncatedError` would be more consistent. That's a fair API-shape suggestion — just a different one from "silent truncation," and probably small enough to either fold in here or carry to a follow-up.

Thanks again for the careful read — the instinct that the timeout and cap paths should give callers equivalent visibility is the right one.

JonoGitty:
Thanks both. @alemuenchen — yes, that's the flow exactly: the cap path already throws after `search()` unwinds (and `directory_tree` throws inline), so there was never a silent partial array.

@temurkhan13 — the consistency point is fair and I've folded it in: the cap now throws a typed **`FsSearchTruncatedError`** (carrying `visited` + `maxVisited`), mirroring `FsTimeoutError`, from both `searchFilesWithValidation` and `directory_tree`. Callers can now `instanceof`-discriminate a safety-cap abort instead of string-matching the message. Cap test updated to assert the type + fields. 160/160 still green. Pushed.

temurkhan13:
@alemuenchen — you're right, that's a genuine misread on my end. I grepped the diff for the post-recursion throw and missed it sitting right above `return results`; the cap was already a hard abort, not a silent partial. Appreciate the correction.

The typed-vs-untyped consistency was the substantive thread of it, and `FsSearchTruncatedError` carrying `visited` + `maxVisited` is exactly the right shape for `instanceof` discrimination. @JonoGitty — clean fold-in, thanks for picking that up in the same PR rather than punting to a follow-up.

LuuOW (commented):

src/filesystem/index.ts
```diff
@@ -548,12 +553,38 @@ server.registerTool(
...
+    // #4162: same defensive shape as searchFilesWithValidation -- honour
+    // FS_SEARCH_EXCLUDE_PREFIXES, bound the recursion, and timeout each
+    // readdir so a lazy provider-backed subtree can't keep the server busy
+    // indefinitely. validatePath already wraps fs.realpath via withFsTimeout
+    // in lib.ts, so only readdir + the entry counter need handling here. We
+    // throw immediately when the cap is hit rather than threading an
+    // `aborted` flag through unwinding recursion -- no caller wants a partial
+    // tree, and the immediate throw makes that contract obvious to future
+    // refactors.
+    assertSearchPathNotExcluded(rootPath, 'directory_tree');
```
Kakarottoooo, line 565: One localized edge case worth checking here: `directory_tree` appears to apply the exclude-prefix prefilter to the raw tool argument before `validatePath()` applies filesystem-server relative-path semantics.

My reading is that `assertSearchPathNotExcluded()` normalizes relative paths via `path.resolve(...)`, which resolves against `process.cwd()`, while `validatePath()` resolves relative requests against the configured allowed directories.

That can make the check behave differently from `search_files`, which validates first and then passes the validated root into the recursive search path.

Example shape:

- allowed directory: `/allowed`
- excluded prefix: `/allowed/slow`
- request path: `"slow"`

If the exclude prefilter sees the raw relative path first, it may not match `/allowed/slow`; then `validatePath()` can resolve the traversal root into the excluded subtree.

Would it be safer to validate the root first, then run `assertSearchPathNotExcluded()` against the validated root?

For example:

```ts
const validRootPath = await validatePath(rootPath);
assertSearchPathNotExcluded(validRootPath, 'directory_tree');
// then build from validRootPath

Kakarottoooo (commented):
I reviewed the current PR branch locally with a focus on runtime behavior rather than style nits.

Validation:
- branch: `fix/4162-filesystem-cloudstorage-hangs`
- commit: `fc52c3343d2e647596732714eacb427440e6bf5f`
- `npm run build --workspace @modelcontextprotocol/server-filesystem` → passed
- `npm test --workspace @modelcontextprotocol/server-filesystem` → passed, 8 files / 160 tests

The core timeout / max-visited direction looks good to me overall: bounded aborts are observable, timeout/truncation errors are typed, and the tests are deterministic rather than depending on real CloudStorage paths.

I left one localized inline comment around `directory_tree` applying `assertSearchPathNotExcluded()` before `validatePath()` resolves relative paths against allowed directories. Other than that, no blocking feedback from my pass.

PR #4218, opened 2026-05-21
filesystem: recover from malformed stdio JSON-RPC lines

### LINKED ISSUE #4206: filesystem: malformed JSON-RPC message on stdio causes crash or zombie state
**Describe the bug**
The `secure-filesystem-server` MCP server does not survive a single malformed JSON-RPC line on stdio. After receiving an incomplete JSON object (truncated `initialize`-style frame with an unclosed `params` field), the server either exits or enters a zombie state where it no longer responds to follow-up requests.

A robust MCP server should treat parse errors as recoverable: return a JSON-RPC error (or skip the bad line per spec) and continue serving subsequent messages. Crashing or hanging breaks the host connector, and here the trigger is protocol fuzzing rather than a missing `allowed_directories` path.

**Environment**
OS: Linux
Platform: Cursor Desktop
MCP server: `@modelcontextprotocol/server-filesystem` via `npx -y`
Reported `serverInfo`: `{"name":"secure-filesystem-server","version":"0.2.0"}`

**To Reproduce**
1. Start the server:
```bash
npx -y @modelcontextprotocol/server-filesystem /tmp
```
2. Write the following single line to the server’s stdin (note: invalid JSON — no closing `}` or `params` value):
```json
{"jsonrpc":"2.0", "method": "test", "params": 
```
(newline-terminated, as MCP stdio transport expects one JSON-RPC message per line.)

3. Wait briefly, then send a valid JSON-RPC request (e.g. `ping` or `tools/list`).

**Expected behavior**
For a parse error on one stdin line:
* Do not terminate the Node process.
* Do not block the event loop indefinitely on a half-read line.
* Return a JSON-RPC error for that message (e.g. parse error / invalid request), or discard the bad line per documented transport rules.
* Continue processing subsequent well-formed messages on the same stdio session.
Optional: emit a one-line diagnostic on stderr (host MCP logs often capture this), e.g.
`[filesystem] JSON parse error on stdin line N: Unexpected end of JSON input`

**Actual behaviour**
The MCP server crashed or became unresponsive (zombie state).

kgarg2468:
## Summary
- Log recoverable filesystem transport errors to stderr so malformed stdio JSON-RPC input has an actionable diagnostic.
- Add a regression test that sends a malformed JSON-RPC line followed by a valid `ping` request over the same stdio session.
- Verify the server stays alive and responds to the valid request after the malformed line.

Fixes #4206.

## Testing
- `npm run build`
- `npx vitest run __tests__/startup-validation.test.ts`
- `npm test -- --run __tests__/startup-validation.test.ts`

### DIFF
--- a/src/filesystem/__tests__/startup-validation.test.ts
+++ b/src/filesystem/__tests__/startup-validation.test.ts
@@ -36,6 +36,64 @@ async function spawnServer(args: string[], timeoutMs = 2000): Promise<{ exitCode
   });
 }
 
+async function pingAfterMalformedJson(
+  args: string[],
+  timeoutMs = 2000,
+): Promise<{ exitCode: number | null; stderr: string; stdout: string }> {
+  return new Promise((resolve) => {
+    const proc = spawn('node', [SERVER_PATH, ...args], {
+      stdio: ['pipe', 'pipe', 'pipe'],
+    });
+
+    let stderr = '';
+    let stdout = '';
+    let sentMessages = false;
+    let settled = false;
+
+    const finish = (exitCode: number | null) => {
+      if (settled) {
+        return;
+      }
+      settled = true;
+      clearTimeout(timeout);
+      proc.kill('SIGTERM');
+      resolve({ exitCode, stderr, stdout });
+    };
+
+    const sendMessages = () => {
+      if (sentMessages) {
+        return;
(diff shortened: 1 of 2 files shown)


LuuOW (commented):

Issue #4129, opened 2026-05-10
Filesystem server canonicalizes configured Windows mapped-drive paths to UNC at startup, then rejects identical-target Y:\ inputs


PapaBearDoes:
## Summary

On Windows, when the Filesystem MCP server is configured with paths on a mapped drive (e.g. `Y:\k8s` where `Y:` is mapped to `\\192.168.xxx.xxx\nas`), the configured paths are canonicalized to their UNC form at server startup. However, incoming tool-call paths using the original `Y:\` form are **not** canonicalized to the same UNC form before the allow-list comparison, so they are rejected — even though they refer to the exact same directory on disk that was configured.

The net effect is that the configured allow-list (as shown in the UI) and the runtime allow-list (as reported by `list_allowed_directories`) disagree, and the user has no way to know which form to send without trial-and-error.

## Environment

- **OS:** Windows (mapped drive `Y:` → `\\192.168.xxx.xxx\nas` over SMB)
- **Client:** Claude Desktop (version: <fill in from Help → About>)
- **Server:** `@modelcontextprotocol/server-filesystem` shipped as the Filesystem extension in Claude Desktop
- **Drive mapping:** `Y:` is a persistent SMB mapping to a NAS share, available to the user account before Claude Desktop launches (`dir Y:\` works in `cmd` at the time of repro)

### Claude Desktop Version:
1.6608.2 (ebf1a1) 2026-05-08T23:17:27.000Z

### Filesystem Extension/Connector Version:
0.2.2

The Filesystem extension's Allowed Directories UI (screenshot attached) is configured with:

- `Y:\git`
- `Y:\wiki`
- `Y:\k8s`

These entries are exactly as typed; the UI preserves the `Y:\` form on save and on subsequent views.

## Steps to Reproduce

1. On Windows, map a drive letter (e.g. `Y:`) to an SMB share (e.g. `\\server\share`) so that `dir Y:\` works.
2. Configure the Filesystem extension's Allowed Directories with one or more entries using the drive-letter form (e.g. `Y:\k8s`).
3. Start Claude Desktop and confirm the Filesystem extension is enabled.
4. From a tool call, invoke `list_allowed_directories`.
5. From a tool call, invoke `list_directory` with `path = "Y:\k8s"` (or any subpath).

## Expected Behavior

- `list_allowed_directories` returns the same form that was configured (`Y:\k8s`, etc.), **or** the input path is canonicalized to the same form as the stored allow-list before comparison.
[...]

PR #4219, opened 2026-05-21
fix(filesystem): handle UNC share root allowlists

### LINKED ISSUE #3756: isPathWithinAllowedDirectories fails for UNC paths on Windows (network drives)

he-yufeng:
Fixes #3756.

UNC share roots keep a trailing separator after Windows path normalization. The allow-list check then appended another separator before the prefix comparison, so a file under `\\server\share\` was compared against `\\server\share\\` and rejected.

This keeps the existing equality check and only adds a separator when the normalized allowed directory does not already end with one.

## To verify

- npm exec -w src/filesystem -- vitest run __tests__/path-validation.test.ts --testNamePattern "UNC paths"
- npm exec -w src/filesystem -- vitest run __tests__/path-validation.test.ts
- npm run build -w src/filesystem
- git diff --check


he-yufeng:
Closing this older branch in favor of #4265, which keeps the UNC share-root fix and also covers the mapped-drive-to-UNC allow-list case from #4129. Keeping one PR should make review easier.

PR #4221, opened 2026-05-21
fix(git): resolve server crash on validation errors

### LINKED ISSUE #4213: [git]  raise_exceptions=True
**Describe the bug**

**To Reproduce**

1. Run python repro.py.
```

inner = {"leaf": True}
for _ in range(200):
    inner = {"child": inner}

msgs = [
]

proc = subprocess.Popen(

```

**Expected behavior**

**Logs**
```
    raise message

```
**Additional context**

STiFLeR7:
# fix(git): resolve server crash on validation errors

## Description
This PR resolves an issue in `mcp-server-git` where the server process crashes upon encountering input validation errors or malformed JSON-RPC messages (e.g. deeply nested requests that hit Pydantic's core recursion limit) instead of returning a JSON-RPC error response and remaining operational.

Fixes #4213

### The Problem
During initialization of `mcp-server-git`, the stdio server is run with `raise_exceptions=True`:
```python
await server.run(read_stream, write_stream, options, raise_exceptions=True)
```
When this option is enabled, any exceptions raised by the stream reader (such as a `ValidationError` when validating incoming JSON-RPC messages) are re-raised. This causes the internal task group to tear down, resulting in the server process crashing with exit code 1.

Other servers in this repository (e.g. `mcp-server-fetch` and `mcp-server-time`) run with the default `raise_exceptions=False`. Under the default configuration, validation errors on stdin are caught, logged, returned as a JSON-RPC error response where appropriate, and the server process remains active.

### The Solution
1. Modified the `server.run(...)` call in `mcp_server_git/server.py` to use `raise_exceptions=False` to align it with other servers and prevent crashes on invalid/deeply nested JSON-RPC inputs.
2. Added a unit integration test `test_server_handles_invalid_json_rpc_gracefully` in `src/git/tests/test_server.py` that spawns the server, sends a deeply nested JSON-RPC payload triggering a validation error, and asserts that the process exits cleanly (exit code 0) when stdin is closed instead of throwing a traceback and crashing.

### DIFF
--- a/src/git/src/mcp_server_git/__init__.py
+++ b/src/git/src/mcp_server_git/__init__.py
@@ -4,6 +4,7 @@ import logging
 import sys
 from .server import serve
 
+
 @click.command()
 @click.option("--repository", "-r", type=Path, help="Git repository path")
 @click.option("-v", "--verbose", count=True)
@@ -19,6 +20,7 @@ def main(repository: Path | None, verbose: bool) -> None:
 
     logging.basicConfig(level=logging_level, stream=sys.stderr)
     asyncio.run(serve(repository))
+
 
 if __name__ == "__main__":
     main()
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -20,58 +20,68 @@ from pydantic import BaseModel, Field
 # Default number of context lines to show in diff output
 DEFAULT_CONTEXT_LINES = 3
 
+
 class GitStatus(BaseModel):
     repo_path: str
+
 
 class GitDiffUnstaged(BaseModel):
     repo_path: str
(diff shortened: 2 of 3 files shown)


LuuOW (commented):

STiFLeR7:
Closing as superseded — #4658 (commit 3aa19f55) already fixes this. It removed the explicit `raise_exceptions=True` from the `server.run(...)` call in `serve()`, which restores the SDK default of `raise_exceptions=False` (confirmed against `mcp==1.27.1`'s `Server.run` signature). That's functionally equivalent to what this PR did (explicitly setting `raise_exceptions=False`) — the server no longer crashes on validation errors either way. No further action needed on this PR.

PR #4224, opened 2026-05-21
fix(filesystem): preserve literal \$ in edit_file newText (closes #4157)

### LINKED ISSUE #4157: filesystem: edit_file newText interpreted as String.prototype.replace replacement-pattern (literal $ corrupted)
## Summary
`edit_file` in `@modelcontextprotocol/server-filesystem` silently corrupts content containing literal `$` characters. The root cause is that `applyFileEdits` (in `src/filesystem/lib.ts`) invokes JavaScript's `String.prototype.replace(searchString, replacementString)` where the **second argument is interpreted as a [replacement string pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement)**, not as a literal string.

| Pattern | Replaced by |
|---|---|
| `$$` | a single `$` |
| `$&` | the matched substring |
| `` $` `` | the substring preceding the match |
| `$'` | the substring following the match |

## Affected file & line
`src/filesystem/lib.ts`, inside `applyFileEdits()`:
```ts
if (modifiedContent.includes(normalizedOld)) {
  modifiedContent = modifiedContent.replace(normalizedOld, normalizedNew); // ← bug
  continue;
}
```
The fallback line-by-line path (using `splice` / `join`) is **not** affected.

## Minimal reproduction
```bash
mkdir /tmp/mcp-repro && cd /tmp/mcp-repro
printf 'price: PLACEHOLDER\n' > sample.txt
# from any MCP client, invoke edit_file:
# oldText: "PLACEHOLDER"
# newText: "$$100 USD"
# expected: price: $$100 USD
# actual:   price: $100 USD   ← $$ collapsed to single $
```
Same corruption with `$&`, `` $` ``, `$'`.

## Proposed fix (one line)
```ts
modifiedContent = modifiedContent.replace(normalizedOld, () => normalizedNew);
```
The callback form treats the return value as literal, disabling replacement-pattern interpretation while preserving "replace first occurrence only" semantics.

## Impact
- Affects ~297k weekly downloads of `@modelcontextprotocol/server-filesystem`
- **Silent corruption**: tool reports success and returns a diff, but on-disk content differs from intent
- Particularly affects JS/TS/PHP code (`$` variables), currency, regex source, shell scripts

## Workaround (until fix lands)
Escape `$` in `newText` by doubling it (`$` → `$$`).

## Related
#2033 — different manifestation of `$` handling in `newText` (that issue reports a parse error; this one reports silent content corruption on successful edits).

Willing to open a PR with the one-line fix + parametric tests if maintainers approve direction.

adityachilka1:
Fixes #4157.

This is the exact fix the issue reporter proposed (and is consistent with the
bug root-cause in the issue body).

### The bug

`src/filesystem/lib.ts` passes `normalizedNew` as `String.prototype.replace`'s
replacement argument. When that argument is a string, JS interprets these
tokens in it:

| Token | Interpreted as           |
| ----- | ------------------------ |
| `$$` | a literal `$`             |
| `$&` | the matched substring     |
| `$`` | the portion before match |
| `$'` | the portion after match  |
| `$n` | nth capture (regex only) |

So when a caller does `edit_file` with `newText: "$100 USD"`, the `$1` token
gets eaten (replaced with the content of capture group 1, which is undefined
for a plain-string search, so it's dropped).

### The fix

Use the callback form (`() => normalizedNew`). MDN explicitly documents
this as the way to disable replacement-pattern interpretation:

> If replacement is a function, the matched substring is replaced with the
> return value of the function.

One-line diff:

```diff
-      modifiedContent = modifiedContent.replace(normalizedOld, normalizedNew);
+      modifiedContent = modifiedContent.replace(normalizedOld, () => normalizedNew);
```

### Tests

Two regression tests added to `src/filesystem/__tests__/lib.test.ts`:

1. `$$` preservation (literal `$`)
2. `$&`, `$``, `$'` preservation (replacement-pattern tokens)

Both tests assert that the exact input string is written verbatim. Without
the fix, both tests fail.

### Verification

```
cd src/filesystem && npm test
```

-> **148 / 148** tests pass across 7 test files (the 2 new tests are in there).

### No behavioral regression

The callback always returns a constant (`{n normalizedNew` {n unbound}), so for
inputs that don't contain `$`, the behavior is identical to the string-arg form.
The only change is that `$` characters now survive the replace.

Closes #4157.

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -440,6 +440,40 @@ describe('Lib Functions', () => {
         );
       });
 
+      it("preserves literal $ characters in newText (regression for #4157)", async () => {
+        // String.prototype.replace interprets $$, $&, $`, $' in the
+        // replacement argument when it is a string. The callback form
+        // disables that interpretation. This test pins the callback form
+        // so the bug can't silently regress.
+        mockFs.readFile.mockResolvedValue("price: PLACEHOLDER\n");
+        const edits = [
+          { oldText: "PLACEHOLDER", newText: "$$100 USD" }
+        ];
+        mockFs.rename.mockResolvedValueOnce(undefined);
+        await applyFileEdits("/test/file.txt", edits, false);
+        expect(mockFs.writeFile).toHaveBeenCalledWith(
+          expect.stringMatching(/\/test\/file\.txt\.[a-f0-9]+\.tmp$/),
+          "price: $$100 USD\n",
+          "utf-8"
+        );
+      });
+
+      it("preserves $&, $` and $' replacement-pattern tokens in newText (#4157)", async () => {
+        mockFs.readFile.mockResolvedValue("TARGET\n");
+        const edits = [
+          { oldText: "TARGET", newText: "a $& b $` c $' d" }
+        ];
+        mockFs.rename.mockResolvedValueOnce(undefined);
+        await applyFileEdits("/test/file.txt", edits, false);
+        // Without the fix: $& would expand to "TARGET", $` to "" (before-match),
+        // and $' to "" (after-match), corrupting the output.
+        expect(mockFs.writeFile).toHaveBeenCalledWith(
+          expect.stringMatching(/\/test\/file\.txt\.[a-f0-9]+\.tmp$/),
+          "a $& b $` c $' d\n",
+          "utf-8"
+        );
+      });
+
       it('handles dry run mode', async () => {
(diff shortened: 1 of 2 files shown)


olaservo:
Thanks for the PR!  We ended up going with #4225 as the canonical fix for #4157.  Closing this as a duplicate.  Thank you for taking the time to submit this.

PR #4226, opened 2026-05-22
fix(fetch): block private network URL fetches

### LINKED ISSUE #4143: Security: mcp-server-fetch lacks SSRF protection; cloud-hosted agent hosts can leak IAM credentials
> ### Summary
>
> Two PyPI-published Python MCP servers — `mcp-server-fetch` (Anthropic reference) and `mcp-server-http-request` (community) — accept arbitrary URLs from agent tool calls and pass them to `httpx.get()` with no scheme allowlist, no host denylist, and no protection against link-local cloud-metadata addresses. On cloud-hosted agent hosts (EC2, GCP, Azure VMs, ECS tasks), an agent coerced via prompt injection into calling these tools with a metadata-service URL will exfiltrate IAM credentials.
>
> The vulnerability is fully mitigated on Amazon Linux 2023 instances launched with default IMDSv2-Required metadata options (the AL2023 default since 2024), but remains exploitable on:
> - Instances explicitly configured `HttpTokens=optional`
> - Instances launched from older AMIs (AL2 and earlier)
> - GCP and Azure metadata services (different default postures)
> - Internal HTTP services reachable from the agent host (RFC 1918 ranges, in-cluster service IPs)
>
> The same root cause — no in-package scheme/host validation — means host-level mitigation is insufficient as defense-in-depth. The fix belongs in the package itself.
>
> ### Demonstration
>
> Performed against `mcp-server-fetch` v2025.4.7 on an EC2 `t3.micro` (Amazon Linux 2023 AMI, region us-east-1) with IMDSv2 set to Optional. Full reproduction runbook: https://github.com/desledishant10/mcp-scan/blob/main/docs/audit-runbook-ec2-ssrf-verification.md
>
> Reproduction in summary:
>
> 1. Launch a `t3.micro` AL2023 instance with any IAM role attached (we used a role with `AmazonEC2ReadOnlyAccess`).
> 2. Console → instance → Modify metadata options → IMDSv2: Optional.
> 3. SSH in; install Python 3.11+ and `mcp-server-fetch`.
> 4. Run a 15-line MCP client harness asking `fetch` to retrieve `http://169.254.169.254/latest/meta-data/iam/security-credentials/<role-name>/`.
>
> Result (redacted): the tool returned the JSON `{"Code":"Success", "AccessKeyId":"ASIA****", "SecretAccessKey":"wJal****", "Token":"AQoD****", "Expiration":"..."}` — a complete, valid AWS IAM credential triplet. Same outcome expected on `mcp-server-http-request` because the root cause is identical (we didn't separately verify on EC2 in the same session).
>
> ### Why this is a class issue, not a single-package issue
>
> The same scanning workflow (run `mcp-scan-analyze` against the captured `tools/list` of each package) flags both with [MCP-S-009](https://github.com/desledishant10/mcp-scan/blob/main/docs/static-rules.md) — "URL-fetching tool with no apparent allowlist" — based solely on the absence of a JSON Schema `pattern` constraint and the absence of validation keywords in the tool descriptions. Both packages are *symptomatic* of an ecosystem norm in which MCP server authors model the URL parameter as an unconstrained string and trust the agent to be responsible. That norm is wrong: the threat model for MCP servers explicitly includes adversarially-influenced tool arguments via prompt injection.
>
> ### Suggested remediation
>
> A small `mcp-server-url-safety` utility module (or an inline check in each affected package) that:
>
> 1. Allowlists `http://` and `https://` only.
> 2. Resolves the hostname to its IP and rejects RFC-reserved ranges before making the request:
>    - `169.254.0.0/16` (link-local, all cloud-metadata addresses)
>    - `127.0.0.0/8`, `::1` (loopback)
>    - `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16` (RFC 1918 private)
>    - `fc00::/7` (IPv6 ULA)
>    - `0.0.0.0/8` (current network)
> 3. Allows an environment-variable opt-in to a customizable allowlist for users who legitimately need to fetch from internal services.
>
> The opt-in pattern is what `mcp-server-fetch`'s existing `--ignore-robots-txt` flag models. The default should be safe; users who know what they're doing can opt out explicitly.
>
> ### Embargo
>
> I'd like to coordinate disclosure with the standard 90-day window. Public release planned for 2026-08-10 unless a fix is in active development, in which case happy to extend. Public release would include this writeup plus the reproduction runbook (no exploit payloads beyond what's already in the public scanner).
>
> I'm happy to test a candidate fix against both packages if you'd like a verification round before release.
>
> ### Tooling
>
> Findings produced by MCP-Scan, an open-source security scanner for MCP servers and AI agents I'm building as a capstone project. Code and full audit trail: https://github.com/desledishant10/mcp-scan
>
> ### Contact
>
> Dishant Desle - didesle7@gmail.com
>
> Thanks for taking a look.

kgarg2468:
## Summary
- validate fetch URLs use http/https and resolve only to public IP addresses
- block localhost, private, loopback, link-local, and metadata-service targets
- follow redirects manually so redirected targets are validated before each request

Fixes #4143.

## Tests
- cd src/fetch && uv run pytest tests/test_server.py -q
- cd src/fetch && uv run ruff check src/mcp_server_fetch/server.py tests/test_server.py
- git diff --check

### DIFF
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -1,5 +1,7 @@
+import ipaddress
+import socket
 from typing import Annotated, Tuple
-from urllib.parse import urlparse, urlunparse
+from urllib.parse import urljoin, urlparse, urlunparse
 
 import markdownify
 import readabilipy.simple_json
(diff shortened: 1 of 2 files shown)


desledishant10:
Thanks for picking this up so quickly, @kgarg2468 really appreciate it.
The fix addresses everything the original disclosure asked for (scheme allowlist + reserved-range denylist), and the per-redirect validation is a nice catch I hadn't explicitly called out. Without it, an attacker could host a public URL that 302s to http://169.254.169.254/... and bypass the initial host check. Glad to see it covered here.
Happy to spin up the same EC2 reproduction setup against this branch to confirm the metadata-service request now refuses cleanly, if that would be useful for the review. Let me know.
Tracking the PR alongside the original disclosure record at https://github.com/desledishant10/mcp-scan/blob/main/disclosures/2026-05-12-mcp-fetch-http-request-ssrf.md so the timeline stays public.

LuuOW (commented):

desledishant10:
Per my offer earlier on this PR, here's the independent re-verification, 39 days post-disclosure.

I built a containerized PoC harness that drives the real `mcp-server-fetch` package against a mock IMDS at the canonical 169.254.169.254 metadata IP. Three states verified today:

| State | Source spec | Result |
|---|---|---|
| Pre-fix (disclosed version) | `mcp-server-fetch==2025.4.7` | VULNERABLE fake IAM credentials returned via JSON-RPC |
| Latest PyPI release | `mcp-server-fetch==2026.6.4` (uploaded 2026-06-04) | **Still VULNERABLE** the fix never landed in a release |
| This PR's branch | `refs/pull/4226/head` | **FIX VERIFIED** `"Fetching private or non-public IP addresses is not allowed"` |

Harness: https://github.com/desledishant10/mcp-witness/tree/main/poc/ssrf (no AWS account required; `make demo-full` reproduces in ~5 seconds after first build, `make demo-fixed` installs from this PR's branch directly).

The fix shape (scheme allowlist + reserved-range denylist + per-redirect validation) is correct against the original disclosure. CI is green (16/16 checks), there's been an outside review from @LuuOW, and the fix has now been independently re-verified against the link-local cloud-metadata vector specifically.

Happy to help with anything else needed for merge additional tests, rebase, review notes, anything useful.

PR #4228, opened 2026-05-22
fix(filesystem): handle invalid stdio JSON-RPC frames

### LINKED ISSUE #4206: filesystem: malformed JSON-RPC message on stdio causes crash or zombie state

### LINKED ISSUE #4207: filesystem: ~1MB JSON-RPC line on stdio causes crash or zombie state
**Describe the bug**
The `secure-filesystem-server` MCP server becomes unavailable after receiving a single ~1 MB JSON-RPC message on stdio. 

Unlike truncated JSON or type-confused fields, this payload is syntactically valid JSON-RPC. The failure is consistent with missing per-line size limits and/or unbounded buffering / `JSON.parse` on huge stdin lines, leading to memory pressure, event-loop blocking, or a broken read loop.

**Environment**
OS: Linux
Platform: Cursor Desktop
MCP server: `@modelcontextprotocol/server-filesystem` via `npx -y`
Reported serverInfo: `{"name":"secure-filesystem-server","version":"0.2.0"}`

**To Reproduce**
1. Start the server:
```bash
npx -y @modelcontextprotocol/server-filesystem /tmp
```
2. After a normal session is established (or on a fresh stdio session), write one newline-terminated line whose JSON body is approximately 1 MB. For example:
```json
{
  "jsonrpc": "2.0",
  "method": "echo",
  "params": { "data": "<1,048,576 × 'A'>" },
  "id": 9999
}
```
(Exact construction: `"A" * 1024 * 1024` inside `params.data`, then `json.dumps` + `\n`.)

3. Wait ~200 ms, then send a valid follow-up request.

**Expected behavior**
* Enforce a maximum stdin line / message size (e.g. 64 KiB–256 KiB, aligned with MCP/host limits) before full buffering or `JSON.parse`.
* On oversize input: return a JSON-RPC error (or drop the line per transport rules), log a clear stderr line, and continue processing subsequent messages.
* Server process remains alive and responsive to normal `initialize` / `tools/list` traffic after the oversized line.

**Actual behaviour**
Observed failure modes:
* Node process exits (transport closes abruptly), or
* Process still running but does not respond to the post-fuzz liveness probe (“zombie state”).

kgarg2468:
## Summary
- replace the filesystem server stdio transport with a defensive line-framed transport
- return JSON-RPC parse/invalid-request errors for malformed or oversized frames
- recover after malformed/oversized input so later JSON-RPC messages still process

Fixes #4206.
Fixes #4207.

## Tests
- cd src/filesystem && npx vitest run __tests__/stdio-transport.test.ts
- cd src/filesystem && npm run build
- cd src/filesystem && npx vitest run __tests__/stdio-transport.test.ts __tests__/startup-validation.test.ts
- git diff --check

### DIFF
--- a/src/filesystem/__tests__/stdio-transport.test.ts
+++ b/src/filesystem/__tests__/stdio-transport.test.ts
@@ -1,0 +1,75 @@
+import { describe, it, expect } from 'vitest';
+import { PassThrough } from 'node:stream';
+import { ErrorCode, type JSONRPCMessage } from '@modelcontextprotocol/sdk/types.js';
+import { SafeStdioServerTransport } from '../stdio-transport.js';
+
+function createTransport(maxLineLength = 1024) {
+  const stdin = new PassThrough();
+  const stdout = new PassThrough();
+  const transport = new SafeStdioServerTransport(stdin, stdout, { maxLineLength });
+  const messages: JSONRPCMessage[] = [];
+  const outputLines: string[] = [];
+  let stdoutBuffer = '';
+
+  transport.onmessage = (message) => {
+    messages.push(message);
+  };
+  stdout.on('data', (chunk) => {
+    stdoutBuffer += chunk.toString('utf8');
+    let newlineIndex = stdoutBuffer.indexOf('\n');
+    while (newlineIndex !== -1) {
+      outputLines.push(stdoutBuffer.slice(0, newlineIndex));
+      stdoutBuffer = stdoutBuffer.slice(newlineIndex + 1);
+      newlineIndex = stdoutBuffer.indexOf('\n');
+    }
+  });
+
+  return { stdin, stdout, transport, messages, outputLines };
+}
+
+async function waitForMicrotasks() {
+  await new Promise((resolve) => setImmediate(resolve));
+}
+
+describe('SafeStdioServerTransport', () => {
+  it('returns a JSON-RPC parse error and keeps processing later messages', async () => {
+    const { stdin, transport, messages, outputLines } = createTransport();
+    await transport.start();
+
+    stdin.write('not-json\n');
+    stdin.write('{"jsonrpc":"2.0","id":1,"method":"ping"}\n');
+    await waitForMicrotasks();
+
+    expect(outputLines).toHaveLength(1);
+    expect(JSON.parse(outputLines[0])).toMatchObject({
+      jsonrpc: '2.0',
+      id: null,
+      error: { code: ErrorCode.ParseError },
+    });
+    expect(messages).toHaveLength(1);
+    expect(messages[0]).toMatchObject({ jsonrpc: '2.0', id: 1, method: 'ping' });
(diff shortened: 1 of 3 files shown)


slegarraga:
Nice, defensive framing here — recovering after a bad frame is the right behavior.

One layering thought for the maintainers, since I just looked at the same class of bug in the SDK: the reason this needs ~240 lines of custom transport is that the SDK's own `StdioServerTransport` drops a malformed-but-recoverable frame (e.g. a request missing `jsonrpc`) — it routes the validation error to `onerror` but never replies, so a client tracking that id hangs. That's filed as modelcontextprotocol/typescript-sdk#2247, and a fix there makes the SDK transport reply with a JSON-RPC `Invalid Request` (-32600) when the offending message has a recoverable id.

If that lands in the SDK, the filesystem server (and every other server) could keep using the stock `StdioServerTransport` instead of maintaining a parallel one. Might be worth deciding where this should live long-term — a per-server transport works, but the SDK-level fix removes the need for it everywhere. Happy to share details from the SDK side if useful.

LuuOW (commented):

PR #4229, opened 2026-05-22
fix(filesystem): handle literal $ characters in edit_file

simpleprogrammer2:
This PR fixes Issue #4157 where literal '$' characters in the `newText` of an `edit_file` operation were being incorrectly interpreted as replacement patterns (e.g., $\&, $1). 

### Technical Change
Switched from using a string replacement pattern in `String.prototype.replace()` to a functional replacement. Using a function ensures that the replacement string is treated as a literal, preserving all special characters.

### Verification
- Added a new test suite `src/filesystem/__tests__/bug_repro.test.ts` which specifically verifies the preservation of single and complex '$' patterns.
- Verified that all existing filesystem tests pass.

### DIFF
--- a/src/filesystem/__tests__/bug_repro.test.ts
+++ b/src/filesystem/__tests__/bug_repro.test.ts
@@ -1,0 +1,45 @@
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs/promises';
+import * as path from 'path';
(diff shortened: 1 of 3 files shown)


olaservo:
Thanks for the PR!  We ended up going with #4225 as the canonical fix for #4157.  Closing this as a duplicate.  Thank you for taking the time to submit this.

PR #4230, opened 2026-05-22
fix(git): set raise_exceptions=False for server stability

simpleprogrammer2:
This PR resolves Issue #4213 where the `mcp-server-git` would crash upon receiving malformed JSON-RPC messages or encountering internal exceptions.

### Technical Change
Changed `raise_exceptions=True` to `False` in the `server.run()` call. This ensures that the server remains resilient and returns standard JSON-RPC error responses instead of terminating the process, consistent with other reference servers.


LuuOW (commented):

PR #4232, opened 2026-05-22
fix(git): stop opting into raise_exceptions=True

### LINKED ISSUE #4213: [git]  raise_exceptions=True

whyujjwal:
## Summary
- stop passing `raise_exceptions=True` when starting the Python git server
- align git server behavior with sibling Python servers so JSON-RPC validation errors are returned to the client instead of crashing the process
- add a regression test covering the `serve()` startup path

## Testing
- `uv run pytest tests/test_server.py -q`
- `uv run ruff check tests/test_server.py src/mcp_server_git/server.py`
- `uv run pyright`

Closes #4213

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -584,4 +584,4 @@ async def serve(repository: Path | None) -> None:
 
     options = server.create_initialization_options()
     async with stdio_server() as (read_stream, write_stream):
-        await server.run(read_stream, write_stream, options, raise_exceptions=True)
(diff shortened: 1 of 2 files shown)


LuuOW (commented):

PR #4240, opened 2026-05-24
Fix grammar in time server get_current_time description

adityasingh2400:
The `get_current_time` tool description reads "Get current time in a specific timezones", which mixes the singular article "a" with the plural "timezones".

The tool only accepts a single timezone, and the time server README already documents the intended singular wording ("Get current time in a specific timezone or system timezone"), so the plural here is just a typo. Tool descriptions are surfaced to model clients and influence tool selection, so the wording is worth getting right.

Changed it to "Get current time in a specific timezone". No behavior change; the existing test suite still passes.


a-akimov (approved):

a-akimov:
@adityasingh2400 thank you for your contribution!

PR #4252, opened 2026-05-26
fix(time): default missing timezones to local timezone

### LINKED ISSUE #2853: mcp-server-time: Default Time Zone

kgarg2468:
## Summary
- Let get_current_time use the configured local timezone when timezone is omitted
- Let convert_time default omitted source/target timezones to the configured local timezone while still requiring time
- Update tool schemas and tests to match the documented local-timezone behavior

Fixes #2853.

## Validation
- cd src/time && uv run pytest

### DIFF
--- a/src/time/src/mcp_server_time/server.py
+++ b/src/time/src/mcp_server_time/server.py
@@ -58,8 +58,12 @@ def get_zoneinfo(timezone_name: str) -> ZoneInfo:
 
 
 class TimeServer:
-    def get_current_time(self, timezone_name: str) -> TimeResult:
+    def __init__(self, local_timezone: str | None = None):
+        self.local_timezone = str(get_local_tz(local_timezone))
+
+    def get_current_time(self, timezone_name: str | None = None) -> TimeResult:
         """Get current time in specified timezone"""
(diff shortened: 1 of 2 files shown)


LuuOW (commented):

PR #4265, opened 2026-05-30
fix(filesystem): handle canonical Windows share paths

### LINKED ISSUE #3756: isPathWithinAllowedDirectories fails for UNC paths on Windows (network drives)

### LINKED ISSUE #4129: Filesystem server canonicalizes configured Windows mapped-drive paths to UNC at startup, then rejects identical-target Y:\ inputs

he-yufeng:
﻿## Summary
- allow filesystem requests using Windows mapped-drive paths to pass when realpath resolves inside the configured UNC allow-list
- handle UNC share allow-list entries that already end with a separator without requiring a double separator
- add Windows regression coverage for existing and new files

Fixes #4129.
Fixes #3756.

## To verify
- npm exec -w src/filesystem -- vitest run __tests__/path-validation.test.ts
- npm exec -w src/filesystem -- vitest run __tests__/lib.test.ts
- npm run build -w src/filesystem
- npm test -w src/filesystem
- git diff --check

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -188,6 +188,38 @@ describe('Lib Functions', () => {
         expect(result).toBe(path.resolve(newFilePath));
       });
 
+      it('allows Windows mapped-drive paths that realpath resolves into the allow-list', async () => {
+        if (process.platform !== 'win32') {
+          return;
+        }
+
+        setAllowedDirectories(['\\\\server\\share\\repo']);
+        mockFs.realpath.mockResolvedValueOnce('\\\\server\\share\\repo\\file.txt');
+
+        const result = await validatePath('Y:\\repo\\file.txt');
+
+        expect(result).toBe('\\\\server\\share\\repo\\file.txt');
+      });
+
+      it('allows new files under Windows mapped-drive parents that resolve into the allow-list', async () => {
+        if (process.platform !== 'win32') {
+          return;
+        }
+
+        const enoentError = new Error('ENOENT') as NodeJS.ErrnoException;
+        enoentError.code = 'ENOENT';
+
+        setAllowedDirectories(['\\\\server\\share\\repo']);
+        mockFs.realpath
+          .mockRejectedValueOnce(enoentError)
+          .mockResolvedValueOnce('\\\\server\\share\\repo');
+
+        const newFilePath = 'Y:\\repo\\newfile.txt';
+        const result = await validatePath(newFilePath);
+
+        expect(result).toBe(path.resolve(newFilePath));
+      });
+
       it('rejects when parent directory does not exist', async () => {
         const newFilePath = process.platform === 'win32' ? 'C:\\Users\\test\\nonexistent\\newfile.txt' : '/home/user/nonexistent/newfile.txt';
         
--- a/src/filesystem/__tests__/path-validation.test.ts
+++ b/src/filesystem/__tests__/path-validation.test.ts
@@ -157,6 +157,17 @@ describe('Path Validation', () => {
         expect(isPathWithinAllowedDirectories('C:\\Users\\project2\\', allowed)).toBe(false);
       }
     });
+
+    it('handles Windows UNC share roots with trailing separators', () => {
+      if (path.sep === '\\') {
+        const allowed = ['\\\\server\\share\\'];
+
+        expect(isPathWithinAllowedDirectories('\\\\server\\share\\file.txt', allowed)).toBe(true);
+        expect(isPathWithinAllowedDirectories('\\\\server\\share\\nested\\file.txt', allowed)).toBe(true);
+        expect(isPathWithinAllowedDirectories('\\\\server\\share-other\\file.txt', allowed)).toBe(false);
+        expect(isPathWithinAllowedDirectories('\\\\other\\share\\file.txt', allowed)).toBe(false);
+      }
+    });
   });
 
   describe('Error handling', () => {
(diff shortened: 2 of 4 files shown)


LuuOW (commented):

he-yufeng:
Superseded by #4704 (merged): Windows paths on POSIX are rejected on main now. Closing.

Issue #4117, opened 2026-05-06
memory: safer persistence defaults, atomic writes, quotas, redaction, and destructive-operation guardrails


lcv-leo:
## Summary

While auditing and hardening local MCP server usage, I reviewed `@modelcontextprotocol/server-memory` and built a local hardened wrapper for our environment. I am opening this issue to share findings and possible upstream improvements.

This is not a report about a known data loss incident in the upstream package. It is a defense-in-depth report about persistence defaults and operational safety for a memory tool exposed to autonomous agents.

## Package/runtime reviewed

- Package: `@modelcontextprotocol/server-memory@2026.1.26`
- License observed: MIT
- Author observed: Anthropic, PBC
- Upstream repo/issue tracker: `modelcontextprotocol/servers`
- npm maintainers observed from registry during local review: `jspahrsummers`, `pcarleton`, `thedsp`, `ashwin-ant`, `ochafik`

## Findings

1. **Default persistence path can land inside the package directory.**
   Local source evidence showed a default path based on `path.dirname(fileURLToPath(import.meta.url))`, with `memory.jsonl` under the package `dist` directory when `MEMORY_FILE_PATH` is not configured. This creates operational fragility:
   - package reinstall/update can remove or overwrite memory state
   - global/package directories are not intuitive user data locations
   - backup/sync behavior is unclear

2. **Writes appear to be whole-file writes without explicit atomic replace and lock discipline.**
   For agent memory, interrupted writes or concurrent host usage can corrupt or lose state. This is especially relevant when the same memory server is configured across multiple MCP hosts.

3. **No obvious mutation journal or backup trail.**
   Debugging accidental writes or deletes is hard without append-only audit entries or backup snapshots.

4. **Destructive tools are available by default.**
   `delete_entities`, `delete_observations`, and `delete_relations` are useful, but high-impact. In an agent context, accidental deletion should require explicit operator opt-in or confirmation mode.

5. **No obvious quotas/caps.**
   Large observations or graphs can become a memory/latency issue. Whole-graph reads and writes are also sensitive to unbounded growth.

6. **No obvious secret redaction before persistence.**
   Memory can become a persistent prompt-injection and secret-retention surface if fetched web content, logs, stack traces, or credentials are persisted without sanitization.

7. **No namespace isolation by default.**
   A single global memory file shared across workspace/host contexts can mix unrelated operational memories unless the user manually configures paths.

## Local hardening behavior that worked well

[...]

PR #4272, opened 2026-05-31
fix(memory): default persistence path to user data directory (#4117)

P4ST4S:
## Description

The default memory file path now resolves to a platform-appropriate user data directory instead of landing inside the package install directory. Package upgrades and reinstalls no longer risk silently dropping persistent state. Legacy files left next to the package are migrated automatically on first run.

This is PR 1 of 3 against #4117, per the [sequencing comment](https://github.com/modelcontextprotocol/servers/issues/4117#issuecomment-4588059846). PR 2 (atomic writes) and PR 3 (corruption recovery / last-known-good snapshot) will follow as separate PRs.

## Server Details

- Server: memory
- Changes to: persistence path resolution and backward-compatibility migration. No tool surface changes.

## Motivation and Context

Closes the most consensual point in #4117. The previous default put `memory.jsonl` at `<package_dir>/memory.jsonl` (via `path.dirname(fileURLToPath(import.meta.url))`), which means:

- `npm update` or a fresh install can remove or overwrite the file
- The package directory is not a user-intuitive location for persistent data
- Standard backup tools (which scan user data directories) skip it

The new default follows the same convention as the rest of the desktop tooling ecosystem:

| Platform | Default location |
|---|---|
| Linux | `$XDG_DATA_HOME/mcp-server-memory`, falling back to `~/.local/share/mcp-server-memory` |
| macOS | `~/Library/Application Support/mcp-server-memory` |
| Windows | `%APPDATA%\mcp-server-memory`, falling back to `~/AppData/Roaming\mcp-server-memory` |
| Other POSIX | `~/.local/share/mcp-server-memory` |

`MEMORY_FILE_PATH` keeps its existing semantics (absolute paths used as-is, relative paths resolved against the package directory for backward compatibility with existing configurations).

## How Has This Been Tested?

Unit tests in `__tests__/file-path.test.ts` cover the resolution logic per platform (`linux` with and without `XDG_DATA_HOME`, `darwin`, `win32` with and without `APPDATA`, including invalid values that the XDG spec requires to be ignored). Migration behavior is tested against legacy `memory.jsonl` and legacy `memory.json` formats, including the "existing user-data file wins over legacy" branch.

Manual smoke test against the built binary on macOS:

1. Cleared the user data directory. Started the server with a JSON-RPC `initialize` over stdio. The server responded with a valid `initialize` result and the user data directory was created.
2. Issued a `create_entities` tool call. Verified the write landed in `~/Library/Application Support/mcp-server-memory/memory.jsonl` and that no file was created in the package directory.
3. Placed a legacy `memory.jsonl` in the package `dist/` directory and removed the user data directory. Started the server. The expected `DETECTED:` and `COMPLETED:` log lines were emitted, the legacy file was moved into the user data directory, and content was preserved byte-exact.

Test suite: 54 passed (was 45), no regressions. Coverage on `index.ts` moved from 86.52% to 87.10%.

## Breaking Changes

Users who relied on the file being inside the package directory will see their data migrated on first run rather than being silently lost on the next upgrade. The migration runs only when:

- `MEMORY_FILE_PATH` is unset (so explicit configurations are untouched), AND
- The user data file does not already exist, AND
- A legacy `memory.jsonl` or `memory.json` is present next to the package

If users referenced the old path externally (backup scripts, sync tools, file watchers), they will need to update those references. The log lines on migration make the new location discoverable.

## Types of changes

## Checklist

- [ ] I have updated the server's README accordingly (intentionally deferred — see note below)
- [x] I have tested this with an LLM client (manual stdio JSON-RPC smoke test)

## Additional context

**On the README**: the existing README describes `MEMORY_FILE_PATH` but does not document the default location explicitly. I'd rather update the README in a follow-up commit once the three storage PRs (path, atomic writes, corruption recovery) are settled, so the documentation captures the full picture in one consistent edit rather than three iterative ones. Happy to fold it into this PR if maintainers prefer.

**On the user data directory name (`mcp-server-memory`)**: I used the npm package name. Alternative shorter names (`memory`, `mcp-memory`) are also fine; I'm happy to rename if there's a preference. The name is referenced in one place (`resolveUserDataDir`), so it's a single-line change.

**Out of scope for this PR (tracked in #4117, follow-up PRs):**

- Atomic writes (PR 2): `saveGraph` currently uses direct `fs.writeFile` and is vulnerable to corruption on abrupt shutdown
- Corruption recovery (PR 3): no fallback when the primary file fails to parse
- Quotas, redaction, write provenance, namespace isolation: warrant separate threads; @policylayer-dan flagged some of these as consumer-layer concerns


P4ST4S:
For maintainer context: PR scope is endorsed by @lcv-leo (issue author) in #4117, the storage-integrity approach was confirmed empirically by @Patdolitse (piia-engram), and the broader consumer-layer trust discussion moved to modelcontextprotocol/modelcontextprotocol#2888 to keep this issue focused on byte integrity. PR is mergeable as-is; happy to rebase if the persistence approach shifts.

LuuOW (commented):

PR #4274, opened 2026-06-01
fix(filesystem): verify write_file output landed on disk (#4138)

### LINKED ISSUE #4138: [filesystem] write_file silently returns success but never writes to disk on Windows
**Describe the bug**
write_file silently returns success without writing the file to disk on Windows. edit_file, read_file, and list_directory work correctly on the same paths — only write_file is broken.

**To Reproduce**
Configure the filesystem server against any local directory.
Call write_file with a path to a new file. Response: File created successfully.
Call read_file on the same path. Response: ENOENT: no such file or directory.
Verify externally (Explorer, Test-Path). File does not exist.

**Expected behavior**
Either write the file, or return an error. Silently returning success while doing nothing leaves the caller unable to detect the failure.
Logs
None visible client-side. Server response is a normal success payload.
Additional context

OS: Windows 11
Claude Desktop: Last
Reproduces with both isUsingBuiltInNodeForMcp: true and system Node
Ruled out: NTFS permissions (PowerShell writes work), Defender, Controlled Folder Access, OneDrive, paths with spaces
Survives full Claude Desktop uninstall + %APPDATA%\Claude wipe + reinstall
edit_file writes persist on the same directory; only write_file is affected

0ywfe:
## Description

Closes #4138. `write_file` resolved successfully even when no bytes reached disk. Callers (Claude Desktop and any other MCP client) got `"File created successfully"` and a normal structured result, with no way to tell the file was never written. This adds a post-write verification step so the failure becomes visible.

## Server Details

- Server: `filesystem`
- Changes to: `tools` (`write_file`; no tool schema or API change)

## Motivation and Context

Reported against Windows 11 on both `isUsingBuiltInNodeForMcp` and system Node. The reporter ruled out NTFS permissions, Defender, Controlled Folder Access, OneDrive, and paths with spaces. `edit_file`, `read_file` and `list_directory` all work on the same paths; only `write_file` silently no-ops.

The host-side root cause varies by environment and is not fully pinned down. What is consistent is the failure *mode*: `fs.writeFile` resolves with no error and nothing is on disk afterwards. Rather than chase the host layer, this closes the gap at the boundary the server controls.

Silent success is the wrong contract on any OS, so the guard is deliberately platform-agnostic. It also catches partial writes, antivirus quarantine mid-write, and rename-to-wrong-path bugs, not just the original Windows symptom.

## Implementation

Inside `writeFileContent`, after the write completes:

1. `fs.stat` the target file.
2. Compare on-disk size to `Buffer.byteLength(content, 'utf-8')`.
3. Throw an explicit error on miss or mismatch.

Runs for both write paths, the `wx` fresh-create branch and the `EEXIST` replace branch. Cost is one extra `stat` per `write_file` call. If that is too expensive for the common path I am happy to put it behind a config flag instead.

## How Has This Been Tested?

Full filesystem suite green in CI (16/16 checks). New cases in `src/filesystem/__tests__/lib.test.ts`:

- Regression for #4138: `fs.writeFile` resolves, `fs.stat` rejects with `ENOENT`, and the test asserts the explicit `"reported success but the file is missing on disk"` error rather than a silent pass.
- Size mismatch: `fs.stat` reports 0 bytes for non-empty content, asserting the size-mismatch error.
- Multibyte UTF-8: confirms `Buffer.byteLength('héllo 🌍', 'utf-8')` is the right yardstick, so non-ASCII content is not falsely flagged.
- The existing happy-path test now also asserts the verification stat is consulted.

This revision updates the two #4115 permission tests as well. They mock a single `fs.stat` for the permission read and now need to account for the verification stat too.

## Breaking Changes

No configuration or client changes required. A `write_file` call that previously returned a false success on a broken host will now return an error. That is the point of the change, but it is a behaviour change and worth stating.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have updated the server's README accordingly. No README change needed; no tool schema, argument or configuration is affected.
- [x] I have tested this with an MCP client. Built the server and drove it over stdio JSON-RPC (`initialize` then `tools/call`); `write_file` and `edit_file` succeed normally, so the verification does not false-positive on a healthy write.
- [ ] I have documented all environment variables and configuration options. None added.

## Out of scope

Pinpointing the underlying Windows-host cause. That wants a separate investigation; this is the boundary fix that makes the failure detectable today.

Applying the same verification to `edit_file` / `applyFileEdits`. The reporter confirms `edit_file` works correctly on the same paths, so it is not affected by this failure mode. The check could be extracted into a shared helper in a follow-up if wanted.


LuuOW (commented):

0ywfe:
Gentle nudge on this one — it's been open about 5 weeks. Current state:

- CI is green (all builds + tests pass)
- #4138 is still open, so this is still the live fix for it
- No competing PRs, no conflicts against main

Happy to rebase or adjust scope if a maintainer wants anything changed. Otherwise it's ready whenever someone with write access can take a look.

0ywfe:
Rebased onto current `main`, CI green, no conflicts. #4138 is still open and nothing on `main` addresses it, so this is still the live fix.

Restating the scope, since the original report was Windows-specific and that undersells what the change does. The host-side root cause varies by environment and is not fully pinned down. What is consistent is the failure *mode*: `fs.writeFile` resolves with no error and nothing is on disk afterwards. Rather than chase the host layer, this closes the gap at the boundary the server controls, by `stat`-ing the target after the write and comparing on-disk size to `Buffer.byteLength(content, 'utf-8')`, throwing an explicit error on miss or mismatch.

That turns a false-positive success into something a client can surface, and it is not Windows-specific in effect. The same check catches partial writes, antivirus quarantine mid-write, and rename-to-wrong-path bugs.

Cost is one extra `stat` per `write_file` call. If that is too expensive for the common path, the obvious alternative is gating it behind a config flag, and I am happy to do that rather than argue for it as the default.

This revision also updates the two #4115 permission tests, which mock a single `fs.stat` for the permission read and now need to account for the verification stat as well.

@olaservo, flagging for you as the filesystem maintainer for the August batch. This has been open since 1 June with CI green throughout. A "no, we don't want this" is a perfectly good outcome and more useful than the queue.

PR #4276, opened 2026-06-02
fix(git): accept JSON-encoded string for git_add files argument

### LINKED ISSUE #4242: [git] maybe more lenient with the input
git_add failed with the message: [{"type":"text","text":"Input validation error: '[\"inc/admin/functions.php\", \"inc/admin/menu.php\", \"inc/admin/register.php\", \"inc/admin/setting.php\"]' is not of type 'array'"}]

adityasingh2400:
## Description

Fixes #4242. The git server's `git_add` tool rejects a perfectly reasonable call when a client sends the `files` argument as a JSON-encoded array string rather than a real array.

## Server Details

- Server: git
- Changes to: tools (the `git_add` input schema and handler)

## Motivation and Context

In #4242 a client called `git_add` with `files` set to the string `'["inc/admin/functions.php", "inc/admin/menu.php", ...]'` and got back:

```
Input validation error: '["inc/admin/functions.php", ...]' is not of type 'array'
```

There are two layers to this:

1. `GitAdd.files` was typed `list[str]`, so `GitAdd.model_json_schema()` published `{"type": "array"}`. The low-level MCP SDK validates raw arguments against the tool's `inputSchema` before the handler runs, so a string value is rejected with the exact "is not of type 'array'" message above.
2. Even if that validation were bypassed, `git_add` passed the raw string straight to `repo.git.add("--", *files)`, which iterates the string character by character. That produces `git add -- [ " f i l e 1 ...`, which fails with `fatal: pathspec '[' did not match any files`.

LLM clients fairly often serialize array arguments as JSON strings, so this is a recurring friction point worth handling leniently.

## What changed

- Widened `GitAdd.files` to `list[str] | str` with a description, so the published schema becomes `anyOf: [array of strings, string]` and the SDK no longer pre-rejects a string value.
- Added a small `normalize_file_list` helper used by `git_add`: when `files` is a string it parses a JSON array of strings into a list, and otherwise treats the value as a single path. Real lists pass through unchanged, so the happy path and `["."]` behavior are untouched.

## How Has This Been Tested?

Ran the git server suite from `src/git` with `uv run pytest`: 44 passed, including four new cases:

- `test_git_add_json_encoded_string`: stages two files passed as `'["file1.txt", "file2.txt"]'`.
- `test_git_add_single_path_string`: a bare path string stages exactly that one file, not its characters.
- `test_normalize_file_list`: lists pass through, JSON arrays parse, non-JSON and non-string-list JSON fall back to a single entry.
- The existing `test_git_add_all_files` and `test_git_add_specific_files` still pass.

I confirmed the new behavioral tests fail on the unmodified code: the original `git_add('["file1.txt", "file2.txt"]')` raises `GitCommandError` (`git add -- [ " f i l e 1 ...`). I also validated end to end that `GitAdd.model_json_schema()` now accepts the array, the JSON-string, and the single-path forms through `jsonschema.validate`, matching how the SDK gates calls.

`uv run pyright` reports 0 errors and `uv run ruff check .` passes.

## Breaking Changes

None. Existing array inputs behave exactly as before. This only adds acceptance of additional input shapes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The leniency is intentionally narrow: only a JSON array whose elements are all strings is expanded into multiple paths. Any other string, including malformed JSON or a JSON object, is treated as a single literal path, which keeps the existing strict pathspec semantics for genuine single-file calls.

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from pathlib import Path
 from typing import Sequence, Optional
@@ -42,7 +43,10 @@ class GitCommit(BaseModel):
 
 class GitAdd(BaseModel):
     repo_path: str
-    files: list[str]
+    files: list[str] | str = Field(
+        ...,
+        description="The files to stage. Normally a list of paths. A JSON-encoded array string such as '[\"a.py\", \"b.py\"]' or a single path string are also accepted for leniency.",
+    )
 
 class GitReset(BaseModel):
     repo_path: str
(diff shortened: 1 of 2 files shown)


LuuOW (commented):

adityasingh2400:
Rebased onto current main to clear a conflict with the new path traversal tests.

Both sets of tests are kept. The ordering is deliberate: `normalize_file_list` runs first, then the traversal validation runs over the normalized list, so a JSON-encoded array containing an escaping path is still rejected rather than slipping past the check as a single opaque string.

Locally on the rebased branch: 46 tests pass, pyright reports 0 errors, ruff is clean.

PR #4278, opened 2026-06-02
[codex] Improve filesystem startup directory diagnostics

### LINKED ISSUE #4238: [server-filesystem] Single missing allowed-directory causes silent process exit due to Promise.all rejection
## Summary

`@modelcontextprotocol/server-filesystem` v0.2.0 (`secure-filesystem-server`) crashes
on startup if **any** allowed directory in `argv` does not exist. The crash is caused
by `Promise.all(dirs.map(fs.stat))` rejecting on the first ENOENT, which terminates
the entire startup with no fallback and no actionable stderr captured by hosts that
don't surface child-process stderr (e.g. Claude Desktop on Windows MSIX, see linked
report).

## Reproduction

```bash
node dist/index.js "C:\exists" "C:\does-not-exist"
```

Output:
Error accessing directory C:\does-not-exist: Error: ENOENT: no such file or directory,
stat 'C:\does-not-exist'
at async Object.stat (node:internal/fs/promises:1038:18)
at async file:///.../dist/index.js:43:23
at async Promise.all (index 1)
at async file:///.../dist/index.js:41:1

Process exits 1. No partial startup, no warning, no continuation with valid dirs.

## Impact

When run under a host that does not pipe child-process stderr to the user (Claude
Desktop on the MSIX/Microsoft Store build of Windows is one such host), the user sees
only "Server transport closed unexpectedly" in the host log with no further detail.
Diagnosing the missing-directory cause requires running `node dist/index.js` manually
outside the host — a step most users won't think to take.

In my case the single offending entry was `C:\NetPhantom 8 QS` (a directory I had
removed). The other 8 allowed directories were valid, but the entire server died.

## Expected behavior

Either:

1. **Skip missing directories with a clear warning to stderr** and start with the
   valid ones. This matches how most "allowlist" tools behave and is the friendliest
   to users whose allowed_directories list drifts.
2. **Fail hard with a clear stderr message naming the offending directory**, then
   `process.exit(1)`. This preserves strict-validation semantics but at least makes
   the error survive `process.on('uncaughtException')`/host stderr capture.

Either way, replacing `Promise.all` with `Promise.allSettled` and inspecting each
result allows the server to report **all** missing/invalid directories in one pass
instead of dying on the first.

## Suggested patch

Around `dist/index.js:41-43` (the startup validation block), something like:

```ts
const results = await Promise.allSettled(
  allowedDirectories.map(async (dir) => {
    const stat = await fs.stat(dir);
    if (!stat.isDirectory()) {
      throw new Error(`Not a directory: ${dir}`);
    }
    return dir;
  }),
);

const valid: string[] = [];
const errors: string[] = [];
results.forEach((r, i) => {
  if (r.status === 'fulfilled') valid.push(r.value);
  else errors.push(`${allowedDirectories[i]}: ${(r.reason as Error).message}`);
});

if (errors.length > 0) {
  console.error(`[server-filesystem] Skipping ${errors.length} invalid director(y/ies):`);
  errors.forEach((e) => console.error(`  - ${e}`));
}

if (valid.length === 0) {
  console.error('[server-filesystem] No valid allowed directories; exiting.');
  process.exit(1);
}

// proceed with `valid`
```

## Environment

- `@modelcontextprotocol/server-filesystem` v0.2.0
- Node.js: bundled with Claude Desktop (Windows MSIX build) + tested with system Node
- OS: Windows 11 Pro Canary build, MSIX-installed Claude Desktop v1.8555.2.0
- Server name reported in initialize: `secure-filesystem-server` v0.2.0

tarunag10:
## Summary

This improves filesystem server startup diagnostics for invalid allowed-directory arguments.

The current source already skips invalid directories and only exits when none of the specified directories are accessible, which matches the behavior requested in #4238. However, the stderr warnings for inaccessible paths were still generic and the hard-fail path did not list the invalid arguments or their concrete failure reasons. In host environments that hide or truncate child-process stderr, every extra bit of actionable startup output matters.

## Root Cause / User Impact

A missing or invalid allowed directory can be difficult to diagnose from a host-level "transport closed" failure. The current startup validation says that a directory cannot be accessed, but it drops the underlying error message such as `ENOENT`, and when all directories are invalid it exits without enumerating the failed inputs.

## Fix

- Record invalid allowed directories during startup validation with a concrete reason.
- Emit warnings in a consistent `Skipping invalid allowed directory` format that includes the directory and reason.
- When no requested directories are accessible, print a compact invalid-directory list before exiting.
- Tighten the startup validation test harness so child processes are terminated after the server emits its startup banner instead of relying only on a blind timeout.
- Extend startup tests to assert missing-directory stderr includes the path and `ENOENT`, and that all-invalid startup reports each failed directory.

## Validation

- `npm ci`
- `npm run build --workspace @modelcontextprotocol/server-filesystem`
- `npx vitest run src/filesystem/__tests__/startup-validation.test.ts --config src/filesystem/vitest.config.ts`
- `npm test --workspace @modelcontextprotocol/server-filesystem`
- `git diff --check`

Fixes #4238.

### DIFF
--- a/src/filesystem/__tests__/startup-validation.test.ts
+++ b/src/filesystem/__tests__/startup-validation.test.ts
@@ -18,6 +18,9 @@ async function spawnServer(args: string[], timeoutMs = 2000): Promise<{ exitCode
     let stderr = '';
     proc.stderr?.on('data', (data) => {
       stderr += data.toString();
+      if (stderr.includes('Secure MCP Filesystem Server running on stdio')) {
+        proc.kill('SIGTERM');
+      }
     });
 
     const timeout = setTimeout(() => {
@@ -66,8 +69,10 @@ describe('Startup Directory Validation', () => {
     const result = await spawnServer([nonExistentDir, accessibleDir]);
 
     // Should warn about inaccessible directory
-    expect(result.stderr).toContain('Warning: Cannot access directory');
+    expect(result.stderr).toContain('Warning: Skipping invalid allowed directory');
     expect(result.stderr).toContain(nonExistentDir);
+    expect(result.stderr).toContain('ENOENT');
+    expect(result.stderr).toContain('no such file or directory');
 
     // Should still start successfully
     expect(result.stderr).toContain('Secure MCP Filesystem Server running on stdio');
@@ -82,6 +87,10 @@ describe('Startup Directory Validation', () => {
     // Should exit with error
     expect(result.exitCode).toBe(1);
     expect(result.stderr).toContain('Error: None of the specified directories are accessible');
+    expect(result.stderr).toContain('Invalid allowed directories:');
+    expect(result.stderr).toContain(nonExistent1);
+    expect(result.stderr).toContain(nonExistent2);
+    expect(result.stderr).toContain('ENOENT');
   });
 
   it('should warn when path is not a directory', async () => {
@@ -91,7 +100,8 @@ describe('Startup Directory Validation', () => {
     const result = await spawnServer([filePath, accessibleDir]);
 
     // Should warn about non-directory
-    expect(result.stderr).toContain('Warning:');
+    expect(result.stderr).toContain('Warning: Skipping invalid allowed directory');
+    expect(result.stderr).toContain(filePath);
     expect(result.stderr).toContain('not a directory');
 
     // Should still start with the valid directory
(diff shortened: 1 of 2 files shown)


LuuOW (commented):

PR #4279, opened 2026-06-02
fix(time): preserve McpError INVALID_PARAMS code in call_tool error handler

basilalshukaili:
## Summary

`call_tool` in `src/time/src/mcp_server_time/server.py` wraps its entire body in `try … except Exception as e: raise ValueError(…)`. `McpError` (raised by `get_zoneinfo` for unknown timezone names) is a subclass of `Exception`, so it was silently caught and re-raised as a plain `ValueError`. This discarded the structured `INVALID_PARAMS` error code that the MCP transport layer uses to distinguish client errors from server faults — causing MCP clients that inspect error codes to see an opaque server error instead of a proper "invalid argument" response.

## Fix

Add an explicit `except McpError: raise` guard *before* the broad handler so `McpError` propagates to the transport layer with its code intact:

```python
except McpError:
    raise           # preserve INVALID_PARAMS / structured error code
except Exception as e:
    raise ValueError(f"Error processing mcp-server-time query: {str(e)}")
```

Also fixes a minor description typo: `"Get current time in a specific timezones"` → `"Get current time in a specific timezone"`.

## Tests

Added three assertions that verify `McpError` carries `INVALID_PARAMS` when an invalid timezone is supplied to `get_current_time` and both positions of `convert_time`. These tests fail on the previous code (a `ValueError` is raised instead of `McpError`).

```
uv run pytest src/time/test/time_server_test.py  # 41 passed
```

### DIFF
--- a/src/time/src/mcp_server_time/server.py
+++ b/src/time/src/mcp_server_time/server.py
@@ -131,7 +131,7 @@ async def serve(local_timezone: str | None = None) -> None:
         return [
             Tool(
                 name=TimeTools.GET_CURRENT_TIME.value,
-                description="Get current time in a specific timezones",
+                description="Get current time in a specific timezone",
                 inputSchema={
                     "type": "object",
                     "properties": {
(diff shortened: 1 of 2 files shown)


basilalshukaili:
Following up here — all checks are green. This is a small, targeted fix to preserve the `INVALID_PARAMS` code that the generic `call_tool` handler was overwriting. Happy to tweak if anything needs changing. Thanks for taking a look!

LuuOW (commented):

PR #4285, opened 2026-06-06
fix(everything): prevent URL elicitation error-path infinite loop

cliffhall:
## Description

Fixes an infinite-loop bug in the `everything` server's `trigger-url-elicitation` tool error path (`errorPath=true`).

Previously the tool threw `UrlElicitationRequiredError` (`-32042`) carrying the **same URL that failed**, and re-threw it on every retry. A client would complete the "prerequisite", retry, hit the identical error, and loop — tripping its own loop detection ("The server requested the same URL again after you completed it").

Two changes break the loop:

1. The prerequisite elicitation now points at a **different** URL than the failing request (`https://modelcontextprotocol.io`).
2. The server records that a prerequisite was issued for a request, keyed on the inputs a client resends verbatim on retry (`sessionId` + `url` + caller-supplied `elicitationId`). When the client satisfies the prerequisite and retries the same call, the retry is recognized, **ignores `errorPath`**, and proceeds via the normal request path instead of re-throwing.

## Server Details
- Server: everything
- Changes to: tools (`trigger-url-elicitation`), tests, docs

## Motivation and Context

Without this, the error-path demonstration of URL elicitation is unusable: any client following the spec's retry-after-prerequisite flow loops forever. This makes the reference behavior correct and demonstrates the intended error-path → satisfy-prerequisite → retry → proceed sequence.

## How Has This Been Tested?

- Tested interactively in an MCP client: the error path now resolves after satisfying the prerequisite and the original request proceeds without looping.
- Unit tests (`src/everything/__tests__/tools.test.ts`): the `-32042` case asserts the prerequisite points at a different URL than the failing request; a new test performs a plain retry (identical args, no echoed id) and asserts the retry takes the request path (`elicitation/create`) instead of throwing.
- `npm run build` and full `vitest` suite pass (105 tests).

### Screen capture

https://github.com/user-attachments/assets/f401db94-7dff-461b-a857-8a2472019a1d

## Breaking Changes

None. Behavior change is confined to the `errorPath=true` retry flow.

## Types of changes
- [x] Documentation update

## Checklist
- [x] I have updated the server's README accordingly (server docs under `src/everything/docs/`)

## Additional context

Design note on retry correlation: a real client retries with the **original** tool-call arguments and does not echo the prerequisite's server-generated `elicitationId`. The marker is therefore keyed on stable resent inputs rather than the resolved/random `elicitationId` (which would change per call and never match). The marker is one-shot — cleared on the recognized retry — so a fresh attempt re-arms the error path.


cliffhall:
@claude review

cliffhall:
Thanks for the thorough review — all points addressed in 5e115592.

**🔴 NUL bytes (blocking):** Fixed. The delimiters were meant to be spaces but ended up as raw NUL bytes during editing, which is why git flagged the file binary and hid the diff. Replaced them with the `\\u0000` escape sequence — same collision-proof delimiter at runtime, but the source is plain UTF-8 again and the file now renders inline in the PR diff (`76/8`, no longer `Bin`).

**🟡 Unbounded growth of `issuedErrorPathElicitations`:** Kept the `Set` for the reference server but added a comment calling out the tradeoff explicitly and pointing at the production approach (a `Map<string, timestamp>` with TTL eviction). Entries are still only removed on a recognized retry, so this is a deliberate demo simplification rather than an oversight.

**🟡 Module-global state / test isolation:** Added an exported `__resetIssuedErrorPathElicitations()` helper and a `beforeEach` in the `trigger-url-elicitation` describe block that clears it, so the suite no longer relies on each case happening to use a distinct session/id.

**🟢 Nits:**
- One-shot toggle surprise — documented in `docs/features.md`: the marker is one-shot per `(session, url, elicitationId)` and cleared on the recognized retry.
- Prerequisite message — now uses a dedicated message ("Open this link to satisfy the prerequisite, then retry the request.") instead of reusing the caller's, fixing the semantic mismatch with the `modelcontextprotocol.io` URL.

Full `vitest` suite still green (105 passing) and `npm run build` clean.

cliffhall:
@claude review

cliffhall:
Thanks for the re-review and the local verification — glad it's all confirmed.

On the optional follow-up: yes, the split is intentional. An explicit-`elicitationId` caller and an omit-`elicitationId` caller are deliberately tracked under different keys — the contract is "retry with the **same** arguments," so an explicit id is part of the request identity that must match on retry, while the omit case correlates on `session` + `url` alone (the random fallback is never used in the key, precisely so it stays stable across the retry). No change needed, just confirming it's by design.

olaservo (approved):
Thanks for finding and resolving this!

PR #4334, opened 2026-06-13
fix(git): keep stdio validation errors recoverable

### LINKED ISSUE #4213: [git]  raise_exceptions=True

he-yufeng:
## Description

Fixes #4213 by letting the git server use the default `Server.run(...)` behavior, matching the time server. Validation errors from stdin parsing should be returned through the JSON-RPC error path instead of being re-raised into the server task group and terminating the process.

## Server Details
- Server: git
- Changes to: stdio transport error handling

## Motivation and Context

`mcp-server-git` was the outlier among the Python reference servers: it passed `raise_exceptions=True` into `Server.run(...)`. For malformed JSON-RPC input, that makes transport-level validation errors abort the server process instead of keeping the stdio session recoverable.

The change removes that override and adds a focused regression test that verifies the git server no longer opts into exception-raising transport behavior.

## How Has This Been Tested?

- `python -m py_compile src\git\src\mcp_server_git\server.py src\git\tests\test_server.py`
- `python -m pytest src\git\tests\test_server.py::test_stdio_transport_errors_do_not_abort_server -q`
- `python -m ruff check src\git\src\mcp_server_git\server.py src\git\tests\test_server.py`
- `git diff --check`

I also ran the full `python -m pytest src\git\tests\test_server.py -q` on Windows. All test assertions reached pass state, but the existing GitPython temp repository fixture hit Windows file-handle cleanup errors during teardown (`PermissionError` while removing `.git` temp dirs). I did not count that as a clean full-suite pass.

## Breaking Changes

None. This aligns git server transport error behavior with the other Python servers instead of changing tool semantics.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

No README/config changes are needed because this is an internal transport recovery behavior fix.

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -584,4 +584,4 @@ async def serve(repository: Path | None) -> None:
 
     options = server.create_initialization_options()
     async with stdio_server() as (read_stream, write_stream):
-        await server.run(read_stream, write_stream, options, raise_exceptions=True)
+        await server.run(read_stream, write_stream, options)
--- a/src/git/tests/test_server.py
+++ b/src/git/tests/test_server.py
@@ -1,7 +1,10 @@
+import asyncio
+from contextlib import asynccontextmanager
 import pytest
 from pathlib import Path
 import git
 from git.exc import BadName
+import mcp_server_git.server as server_module
 from mcp_server_git.server import (
     git_checkout,
     git_branch,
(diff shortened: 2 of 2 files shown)


he-yufeng:
Gentle bump. This keeps the stdio git server's validation errors as structured errors instead of collapsing them, so callers can still tell what failed. Happy to rebase if useful.

he-yufeng:
Second bump. The git server still runs its stdio loop with raise_exceptions=True, so one malformed request takes the whole server down. The June security pass touched git_add, not this path.

he-yufeng:
Superseded by #4658 (merged 2026-09-02): raise_exceptions is gone from server.run on main. Closing.

PR #4383, opened 2026-06-16
fix(memory): skip duplicate entities and relations within a single batch

JSap0914:
## Description

`create_entities` and `create_relations` only de-duplicated against the **existing** graph, never against the items in the incoming request. So passing the same entity name (or an identical relation) twice in a single call persisted duplicate records to `memory.jsonl`.

This contradicts the documented behavior in the README — *"Ignores entities with existing names"* and *"Skips duplicate relations"* — and breaks the implicit name-uniqueness invariant the rest of the manager relies on (`addObservations`/`deleteEntities`/`deleteObservations` all key on `name`, and `find()` only returns the first match).

The fix de-duplicates within the input batch as well, keeping the first occurrence. Cross-call de-duplication is unchanged.

## Server Details
- Server: memory
- Changes to: core logic (`KnowledgeGraphManager.createEntities`, `KnowledgeGraphManager.createRelations`)

## Motivation and Context

Reproduction (before this change):

```jsonc
// create_entities
[{ "name": "Alice", "entityType": "person", "observations": ["a"] },
 { "name": "Alice", "entityType": "person", "observations": ["b"] }]
// -> graph now contains TWO "Alice" entities

// create_relations
[{ "from": "Alice", "to": "Bob", "relationType": "knows" },
 { "from": "Alice", "to": "Bob", "relationType": "knows" }]
// -> graph now contains TWO identical relations
```

After the change each of the above creates a single record, matching the documented "ignores existing names" / "skips duplicate relations" contract.

## How Has This Been Tested?

Added two regression tests to `src/memory/__tests__/knowledge-graph.test.ts` (within-batch duplicate entities, within-batch duplicate relations) and ran the existing vitest suite:

```
$ npx vitest run            # in src/memory
 Test Files  2 passed (2)
      Tests  47 passed (47)
```

Both new tests fail on `main` (the manager returns/persists 2 records) and pass with this change. `npx tsc --noEmit` is clean.

## Breaking Changes

None. Behavior only changes for inputs that were already malformed (duplicate items in one request); valid de-duplicating callers are unaffected.

## Types of changes
- [ ] Documentation update

## Checklist
- [ ] I have updated the server's README accordingly *(no change needed — fix aligns behavior with the existing README description)*
- [ ] I have tested this with an LLM client *(verified via the vitest suite instead)*

> Prepared with AI assistance; the diff, tests, and verification were reviewed by me.

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -59,6 +59,20 @@ describe('KnowledgeGraphManager', () => {
       const newEntities = await manager.createEntities([]);
       expect(newEntities).toHaveLength(0);
     });
+
+    it('should ignore duplicate entity names within a single batch', async () => {
+      const entities: Entity[] = [
+        { name: 'Alice', entityType: 'person', observations: ['first'] },
+        { name: 'Alice', entityType: 'person', observations: ['second'] },
+      ];
+
+      const newEntities = await manager.createEntities(entities);
+      expect(newEntities).toHaveLength(1);
+
+      const graph = await manager.readGraph();
+      expect(graph.entities).toHaveLength(1);
+      expect(graph.entities[0].name).toBe('Alice');
+    });
   });
 
   describe('createRelations', () => {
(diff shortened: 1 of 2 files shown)


olaservo (approved):
Looks good to me, I rebased from main and fixed conflicts.

PR #4392, opened 2026-06-19
Add MCP Observatory CI check

KryptosAI:
This adds a lightweight MCP Observatory check for the reference `server-everything` MCP server.\n\nWhy it helps:\n\n- verifies MCP tools, prompts, and resources still respond correctly\n- catches schema drift and common security footguns before release\n- posts a readable PR report for maintainers\n- gives users a compatibility signal when evaluating MCP servers\n\nI validated the target locally with:\n\n```bash\nnpx @kryptosai/mcp-observatory@latest test --target mcp-observatory.target.json --security --deep\n```\n\nResult: passed, with 13 tools, 4 prompts, and 7 resources detected from `npx -y @modelcontextprotocol/server-everything@latest`.\n\nIt runs in GitHub Actions and does not require an MCP Observatory account. If this is too broad for the repo, I can scope it differently or adjust the workflow.


KryptosAI:
Small review note: I’m happy to narrow this PR if that makes it easier to evaluate, for example keeping only the `sequential-thinking` reference check first and leaving broader server coverage for a follow-up.

The workflow is intended to stay read-only and advisory: it verifies the MCP server starts, inventories tools/prompts/resources, and surfaces compatibility/schema/security findings without requiring an account or hosted service.

KryptosAI:
Versioned follow-up: MCP Observatory v0.27.0 is now published with optional SARIF output and GitHub Code Scanning support.

That means this check can stay as a normal read-only compatibility gate, or maintainers can later opt into security-native findings with `setup-ci --sarif` and `security-events: write` if Code Scanning is useful for this repo.

Docs: https://github.com/KryptosAI/mcp-observatory/blob/main/docs/github-code-scanning-for-mcp.md

PR #4451, opened 2026-07-01
fix(memory): persist the knowledge graph atomically (temp file + rename)

ameyypawar:
## Summary

`KnowledgeGraphManager.saveGraph` rewrote the whole memory file with a single `fs.writeFile`. An interrupted process, or a concurrent reader observing the file mid-write, could see a truncated or partially written `memory.jsonl` — corrupting the store.

Write to a temporary file in the same directory and atomically `rename` it into place, cleaning up the temp file on error. This mirrors the atomic-write pattern already used by the filesystem server (`src/filesystem/lib.ts`).

Addresses the atomic-write finding (#2) in #4117.

## Change

- `src/memory/index.ts`: `saveGraph` now writes to `${memoryFilePath}.<randomBytes>.tmp` and then `fs.rename`s it over the target, unlinking the temp on failure. Adds `import { randomBytes } from 'crypto'`. File contents are otherwise unchanged (`lines.join("\n")`).

## Tests

`src/memory/__tests__/knowledge-graph.test.ts`:
- asserts the write goes to a temp file in the target's **own directory** and is renamed onto the exact target path (same filesystem → atomic);
- forces the `rename` to fail and asserts the temp file is cleaned up (the error branch);
- the existing round-trip / JSONL tests continue to pass unchanged.

Verified locally: `npm test` → 52/52 passing, `npm run build` (`tsc`) clean.

## Notes

- On the same filesystem, `rename(2)` is atomic, so a reader always sees either the old file or the fully written new one. The temp is created in the target's directory specifically to keep the rename on one filesystem.
- This matches the filesystem server's behavior on Windows: a `rename` can still fail if another process holds the file open without share-delete. In that case the save surfaces a clean error and the existing file is left intact (no corruption). A cross-platform retry-on-lock would be a reasonable follow-up for both servers.

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -417,10 +417,14 @@ describe('KnowledgeGraphManager', () => {
         ]);
 
         // saveGraph writes to a temp file, then atomically renames it into place.
-        expect(renameSpy).toHaveBeenCalledWith(
-          expect.stringMatching(/\.tmp$/),
-          testFilePath
-        );
+        expect(renameSpy).toHaveBeenCalledTimes(1);
+        const [tmpArg, destArg] = renameSpy.mock.calls[0];
+        expect(String(destArg)).toBe(testFilePath);
+        // The temp file must live in the target's own directory so the rename
+        // stays on the same filesystem (and is therefore atomic); guard that
+        // invariant rather than just matching any ".tmp" suffix.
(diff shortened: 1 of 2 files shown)


olaservo:
The atomic temp-file-plus-rename write landed via #4642 for #4614, with the same `randomBytes` temp name and same-directory rename this PR uses. Closing as superseded. Thanks for the PR!

PR #4456, opened 2026-07-02
feat(fetch): add retry/backoff for transient 429/5xx and network errors

### LINKED ISSUE #4449: fetch: no retry/backoff on transient 429/5xx or network errors
### Problem

The `fetch` MCP server (`src/fetch/`) proxies HTTP requests on behalf of agents. When the target URL returns a transient error — `429 Too Many Requests`, `500 Internal Server Error`, `502 Bad Gateway`, `503 Service Unavailable`, `504 Gateway Timeout` — the server returns the error immediately to the agent with no retry.

Agents (Claude Code, Kilo Code, OpenCode) that call `fetch` for documentation lookups, API calls, or web scraping then receive an error result and either halt or waste a full LLM turn handling the error, when a 1-2 second retry would have succeeded.

### Expected behavior

For idempotent fetches (GET, HEAD), the fetch server should retry transient errors with exponential backoff:

- Retryable: `429`, `500`, `502`, `503`, `504`, and network errors (`ECONNRESET`, `ETIMEDOUT`, `ENOTFOUND` with a short TTL)
- Non-retryable: `400`, `401`, `403`, `404`, `405`, `409`, `410`, `422`
- Default retry policy: 3 attempts, initial delay 1s, backoff factor 2x, max delay 10s
- Configurable via env: `FETCH_MAX_RETRIES`, `FETCH_RETRY_DELAY_MS`
- Respect `Retry-After` header when present on 429 responses

POST/PUT/DELETE should NOT be retried by default (not idempotent); opt-in via config.

### Why

Transient errors are the norm for public APIs and documentation sites under load. A retry layer in the MCP server is more efficient than having every agent implement its own retry logic, and it keeps agent prompts clean (agents describe what they need, not how to handle HTTP transience).

### Environment

- MCP servers: `@modelcontextprotocol/server-fetch` (latest)
- Runtime: Node.js v24.18.0, Windows 11 Enterprise
- Use case: agents fetching npm docs, GitHub raw content, and public APIs through `fetch` MCP tool

chakshu-dhannawat:
## Summary

Resolves #4449.

The Python fetch server currently raises `McpError` immediately on any HTTP error or non-2xx status. This means a single transient 429 (rate-limit), 503 (overloaded server), or momentary network timeout causes the fetch to fail completely, forcing the LLM to retry the whole tool call.

Changes:
- **Retry on 429, 500, 502, 503, 504** with configurable `max_retries` (default 3)
- **Retry on `httpx.TransportError`** (covers `TimeoutException`, `ConnectError`, etc.)
- **Respect `Retry-After` header** on 429; fall back to full-jitter exponential backoff capped at 30 s
- **`--max-retries` CLI flag** so operators can tune or disable retries (`--max-retries 0`)
- **12 new tests** covering: 503→success, 429 Retry-After, exhausted retries, 404 non-retry, network error retry, network error exhaustion, and `_retry_wait` unit tests

Non-retryable errors (400, 401, 403, 404, etc.) still raise immediately — unchanged behavior.

## Test plan

- [x] All 32 tests pass: `uv run pytest -v`
- [x] No ruff issues: `uv run ruff check .`
- [x] pyright: 0 errors, 0 warnings: `uv run pyright`
- [x] Tested with `--max-retries 0` to verify backwards-compatible opt-out

### DIFF
--- a/src/fetch/src/mcp_server_fetch/__init__.py
+++ b/src/fetch/src/mcp_server_fetch/__init__.py
@@ -16,9 +16,15 @@ def main():
         help="Ignore robots.txt restrictions",
     )
     parser.add_argument("--proxy-url", type=str, help="Proxy URL to use for requests")
+    parser.add_argument(
+        "--max-retries",
+        type=int,
+        default=3,
+        help="Maximum number of retries for transient errors (429/5xx/network). Default: 3",
+    )
 
     args = parser.parse_args()
-    asyncio.run(serve(args.user_agent, args.ignore_robots_txt, args.proxy_url))
+    asyncio.run(serve(args.user_agent, args.ignore_robots_txt, args.proxy_url, args.max_retries))
 
 
 if __name__ == "__main__":
--- a/src/fetch/src/mcp_server_fetch/server.py
+++ b/src/fetch/src/mcp_server_fetch/server.py
@@ -1,3 +1,5 @@
+import asyncio
+import random
 from typing import Annotated, Tuple
 from urllib.parse import urlparse, urlunparse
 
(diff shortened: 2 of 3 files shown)


chakshu-dhannawat:
Hi, I've refreshed this against main. Could a maintainer take a look when possible? Thanks.

PR #4458, opened 2026-07-02
fix(memory): reject relations with missing entities

### LINKED ISSUE #4457: memory: create_relations accepts relations to non-existent entities (dangling relations, silent graph corruption)
### Summary

The Memory server's `create_relations` does not check that the referenced entities exist, so it silently persists relations pointing at entities that were never created (dangling relations). `add_observations` in the same file *does* validate entity existence, so the behaviour is inconsistent.

### Reproduction

Against `@modelcontextprotocol/server-memory` (current `main`), on a fresh graph:

1. Call `create_relations` with `[{ "from": "Ghost_A", "to": "Ghost_B", "relationType": "knows" }]` — no entities exist yet.
   → returns success (`isError: false`).
2. Call `read_graph` → `entities: []`, `relations: [{ "from": "Ghost_A", "to": "Ghost_B", ... }]` — a dangling relation.

For contrast, `add_observations` with `{ "entityName": "Ghost_C", "contents": [...] }` correctly fails with `Entity with name Ghost_C not found`.

Live output:
```
create_relations(nonexistent) -> isError: False
  returned: [ { "from": "Ghost_A", "to": "Ghost_B", "relationType": "knows" } ]
add_observations(nonexistent) -> isError: True
  returned: Entity with name Ghost_C not found
read_graph -> entities: 0 | relations: 1
```

A minimal (~30-line) stdio reproduction using the MCP Python client is here: https://github.com/chrisbartoloburlo/llmcontract-mcp-audit/blob/main/servers/memory/probe.py

### Root cause

`src/memory/index.ts`, `createRelations` filters only for duplicates and then pushes, with no existence check — unlike `addObservations`, which throws `Entity with name ... not found`:

```ts
async createRelations(relations: Relation[]): Promise<Relation[]> {
  const newRelations = relations.filter(r => !graph.relations.some(er =>
    er.from === r.from && er.to === r.to && er.relationType === r.relationType));
  graph.relations.push(...newRelations);   // no check that from/to exist
  await this.saveGraph(graph);
  return newRelations;
}
```

### Impact

`read_graph` returns referentially-inconsistent state that persists to `MEMORY_FILE_PATH`. Downstream consumers that assume referential integrity (graph traversal, visualization, an agent reasoning over the graph) get silently corrupt data — whereas the analogous observation case fails loudly.

### Suggested fix

In `createRelations`, validate that every relation's `from`/`to` names an existing entity (mirroring `addObservations`) and throw, or skip-and-report, otherwise.

---

_Found with [llmcontract](https://github.com/chrisbartoloburlo/llmcontract), a session-type runtime monitor for agent/tool protocols: encoding the documented "entities before relations" workflow as a contract flags these sequences that the server accepts. Full case study (protocol, probe, replay tests): https://github.com/chrisbartoloburlo/llmcontract-mcp-audit/tree/main/servers/memory_

gingerninja85:
## Summary
- validate that both endpoints exist before creating memory graph relations
- reject missing `from` or `to` entities with the same error style used by `addObservations`
- add regression tests that ensure dangling relations are not persisted

Fixes #4457

## Test Plan
- `cd src/memory && npm test`
- `cd src/memory && npm run build`

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -103,6 +103,36 @@ describe('KnowledgeGraphManager', () => {
       const newRelations = await manager.createRelations([]);
       expect(newRelations).toHaveLength(0);
     });
+
+    it('should reject relations with a missing from entity', async () => {
+      await manager.createEntities([
+        { name: 'Alice', entityType: 'person', observations: [] },
+      ]);
+
+      await expect(
+        manager.createRelations([
+          { from: 'Ghost', to: 'Alice', relationType: 'knows' },
+        ])
+      ).rejects.toThrow('Entity with name Ghost not found');
+
+      const graph = await manager.readGraph();
+      expect(graph.relations).toHaveLength(0);
+    });
+
+    it('should reject relations with a missing to entity', async () => {
+      await manager.createEntities([
+        { name: 'Alice', entityType: 'person', observations: [] },
+      ]);
+
+      await expect(
+        manager.createRelations([
+          { from: 'Alice', to: 'Ghost', relationType: 'knows' },
+        ])
+      ).rejects.toThrow('Entity with name Ghost not found');
+
+      const graph = await manager.readGraph();
+      expect(graph.relations).toHaveLength(0);
+    });
   });
 
   describe('addObservations', () => {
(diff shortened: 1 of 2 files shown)


gingerninja85:
This is still mergeable with the memory package tests and builds passing. The change keeps relation creation from accepting dangling entity references and includes regression coverage. Happy to adjust if there is a preferred error shape

olaservo:
Thanks for this. #4477 makes the same change to `createRelations` (same check, same error message) and also includes the fail-fast batch test and the README line documenting the new behaviour, so I am going with that one. Closing as a duplicate.

PR #4470, opened 2026-07-04
fix(git): normalize git_log schema across filtered and unfiltered branches

### LINKED ISSUE #4469: git: git_log output schema differs between filtered vs unfiltered branches
## Problem
`git_log` returns two different string shapes depending on whether `start_timestamp`/`end_timestamp` is passed, breaking downstream parsing.

## Steps
1. Run `mcp-server-git` (v0.6.2) against any repo.
2. Call `git_log` with no timestamps.
3. Call `git_log` with `start_timestamp='1 week ago'`.
4. Compare the two returned entries for the same commit.

## Expected
Same output schema regardless of filter branch.

## Actual
Unfiltered branch (`src/git/src/mcp_server_git/server.py:191-197`) uses `!r`, producing:
```
Commit: 'a1b2c3...'
Author: <git.Actor "Name <email>">
Message: 'subject\n\nbody\n'
```
Filtered branch (`server.py:180-185`) emits raw values via `--format=%H%n%an%n%ad%n%s%n`:
```
Commit: a1b2c3...
Author: Name
Message: subject
```
Also, filtered branch loses commit body entirely (`%s` = subject only) while unfiltered emits `commit.message` (subject + body).

## Environment
- mcp-server-git 0.6.2 (`src/git/pyproject.toml`)
- Python >=3.10, gitpython>=3.1.50

Thanks for maintaining modelcontextprotocol/servers!

Sagargupta16:
## Summary

Closes #4469.

\`git_log\` returned two different string shapes depending on whether \`start_timestamp\`/\`end_timestamp\` was passed:

**Before (unfiltered branch — `server.py:187-197`)**

Used \`commit.hexsha!r\` / \`commit.author!r\`, which produced repr-quoted values:

\`\`\`
Commit: 'a1b2c3d4e5f6...'
Author: <git.Actor \"Name <email>\">
Date: 2026-07-01 12:00:00+00:00
Message: 'subject\n\nbody\n'
\`\`\`

**Before (filtered branch — `server.py:180-185`)**

Used `git log --format=%H%n%an%n%ad%n%s%n` and split, producing raw values but losing the commit body (\`%s\` is subject-only):

\`\`\`
Commit: a1b2c3d4e5f6...
Author: Name
Date: Wed Jul 1 12:00:00 2026 +0000
Message: subject
\`\`\`

Downstream parsers that split on \`\n\` and \`:\` had to special-case which branch produced the entry.

## Fix

Collapse both branches onto a single \`repo.iter_commits(max_count=..., since=..., until=...)\` call, so there is exactly one code path — schema drift is impossible by construction. Also drop the \`!r\` formatting so both filtered and unfiltered entries emit raw values (bare commit hash, plain author \`Name <email>\`, full \`commit.message\` including body).

The flag-injection defense (\`start_timestamp.startswith(\"-\")\`) is preserved and now applies uniformly whether or not the fast path is taken.

## Tests

Added \`test_git_log_schema_matches_across_filter_branches\` that asserts:

- Same key order (\`Commit\`, \`Author\`, \`Date\`, \`Message\`) across both filtered and unfiltered calls
- Neither branch emits repr-quoted commit hashes (guards against the leading-\`'\` regression)

All existing \`git_log\` tests pass; the unrelated \`test_validate_repo_path_symlink_escape\` failure on Windows is pre-existing on upstream (verified via stash).

## Compatibility note

The `Date:` line for the unfiltered branch now uses `commit.authored_datetime` (already the case before), and the filtered branch's format string is gone. If any external consumer was pattern-matching on the raw git-format date (\`Wed Jul 1 12:00:00 2026 +0000\`), they now get the ISO-style datetime for both branches, which is the consistent shape #4469 asks for.

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -157,45 +157,27 @@ def git_reset(repo: git.Repo) -> str:
     return "All staged changes reset"
 
 def git_log(repo: git.Repo, max_count: int = 10, start_timestamp: Optional[str] = None, end_timestamp: Optional[str] = None) -> list[str]:
-    if start_timestamp or end_timestamp:
-        # Defense in depth: reject timestamps starting with '-' to prevent flag injection
-        if start_timestamp and start_timestamp.startswith("-"):
-            raise ValueError(f"Invalid start_timestamp: '{start_timestamp}' - cannot start with '-'")
-        if end_timestamp and end_timestamp.startswith("-"):
-            raise ValueError(f"Invalid end_timestamp: '{end_timestamp}' - cannot start with '-'")
-        # Use git log command with date filtering
-        args = []
-        if start_timestamp:
-            args.extend(['--since', start_timestamp])
-        if end_timestamp:
-            args.extend(['--until', end_timestamp])
-        args.extend(['--format=%H%n%an%n%ad%n%s%n'])
+    # Defense in depth: reject timestamps starting with '-' to prevent flag injection
+    if start_timestamp and start_timestamp.startswith("-"):
+        raise ValueError(f"Invalid start_timestamp: '{start_timestamp}' - cannot start with '-'")
+    if end_timestamp and end_timestamp.startswith("-"):
+        raise ValueError(f"Invalid end_timestamp: '{end_timestamp}' - cannot start with '-'")
 
-        log_output = repo.git.log(*args).split('\n')
+    rev_kwargs: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


Sagargupta16:
Closing this as superseded. #4658 landed the same normalization, collapsing the filtered and unfiltered paths into a single iter_commits call, and added test_git_log_filtered_unfiltered_parity which covers the same assertion as the test in this branch. Nothing here is missing from main now. Thanks for sorting it upstream.

PR #4491, opened 2026-07-07
fix(git): git_log date-filtered path misparses every commit after the first

tao-hpu:
## Bug

`git_log` with `start_timestamp`/`end_timestamp` uses `--format=%H%n%an%n%ad%n%s%n`. The trailing `%n`, plus the newline git prints between log entries, makes each commit emit **5 lines**, while the parser walks the output in strides of **4** (`server.py`, `for i in range(0, len(log_output), 4)`). Alignment breaks from the second commit on.

Actual output on a 3-commit repo with `start_timestamp="2000-01-01"`:

```
Commit: 15d3a078...          <- commit 3, correct
...
Commit:                      <- blank: this is the separator line
Author: 36688c31c43d9bb...   <- the hash landed in the author column
Date: T                      <- the author landed in the date column
Message: Wed Jul 8 01:04:18  <- the date landed in the message column
```

## Fix

Drop the trailing `%n` so each commit is exactly 4 lines. One-character change; the non-filtered path is untouched.

## Tests

The date-filtered path had no coverage (`test_git_log`/`test_git_log_default` only exercise the plain path). Added `test_git_log_with_date_filter`, which asserts per-field alignment for every returned entry. Verified it fails on the old format string and passes with the fix. Full suite: 44 passed.

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -169,7 +169,10 @@ def git_log(repo: git.Repo, max_count: int = 10, start_timestamp: Optional[str] 
             args.extend(['--since', start_timestamp])
         if end_timestamp:
             args.extend(['--until', end_timestamp])
-        args.extend(['--format=%H%n%an%n%ad%n%s%n'])
+        # No trailing %n: git already separates entries with a newline, so a
+        # trailing %n would emit 5 lines per commit and break the groups-of-4
+        # parsing below for every commit after the first.
+        args.extend(['--format=%H%n%an%n%ad%n%s'])
 
         log_output = repo.git.log(*args).split('\n')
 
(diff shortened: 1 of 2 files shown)


tao-hpu:
Friendly ping — still happy to iterate if anything needs changing.

Summary: date-filtered `git_log` used `--format=...%n` which made each commit 5 lines while the parser stepped by 4, so every commit after the first was misaligned. Dropped the trailing `%n` and added `test_git_log_with_date_filter` (fails on old format, passes with the fix). Full suite was 44 passed when opened.

Happy to rebase or adjust if review has preferences.

tao-hpu:
Closing this. `git_log` on main now builds its output from `repo.iter_commits()` instead of splitting `--format` output by line, so the misparse this PR fixed no longer exists.

Issue #4488, opened 2026-07-06
[windows-mcp] Scheduled task auto-start fails (0xC0000005) when uvx installed via WinGet symlink; InteractiveToken logon trigger cannot resolve symlink before shell initialization


Chris-CGG:
## Environment
- OS: Windows 11 (domain-joined, service account `manifold.svc`)
- Claude Desktop: latest (2026)
- windows-mcp version: installed via WinGet (`astral-sh.uv`)
- Deployment pattern: MCP server running as Windows Scheduled Task (logon trigger, `InteractiveToken`)

## Description

When an MCP server is installed via WinGet and launched from a Windows Scheduled Task, the server crashes on auto-start with exit code `0xC0000005` (access violation). The root cause is that WinGet installs binaries as **symbolic links** under `AppData\Local\Microsoft\WinGet\Links\`, and Windows Scheduled Tasks with `InteractiveToken` logon type cannot follow these symlinks before the user shell is fully initialized.

The server works correctly when started manually (`Start-ScheduledTask`) or launched directly by Claude Desktop — only the at-logon scheduled task trigger fails.

## Symptoms

- Scheduled task fires at logon (`Event ID 119` — task launched due to user logon)
- Task exits immediately with `LastTaskResult: 3221225477` (`0xC0000005`)
- Port never becomes available
- Claude Desktop shows the connector as registered but tools never load in new sessions

## Root Cause

WinGet installs `uvx.EXE` as a zero-byte symbolic link:

```
C:\Users\<user>\AppData\Local\Microsoft\WinGet\Links\uvx.EXE
  → C:\Users\<user>\AppData\Local\Microsoft\WinGet\Packages\astral-sh.uv_Microsoft.Winget.Source_8wekyb3d8bbwe\uvx.exe
```

The `start-server.cmd` generated by `windows-mcp install` points to the symlink path. When Task Scheduler fires the `InteractiveToken` logon trigger, the user's shell environment is not yet fully initialized, and symlink resolution fails — producing an access violation in the spawned process.

## Workaround

Replace the symlink path in `start-server.cmd` with the real executable path under `WinGet\Packages\`:

```batch
@echo off
setlocal

:: Guard: exit cleanly if another instance is already running on port 8000
netstat -ano | findstr "127.0.0.1:8000" >nul 2>&1
if %ERRORLEVEL% EQU 0 (
[...]

PR #4499, opened 2026-07-08
docs: clarify Windows launch commands

### LINKED ISSUE #4487: [server-filesystem] Windows: server fails to start when Node.js installed at default path (C:\Program Files\nodejs) due to space in command path
## Environment
- OS: Windows 11 (domain-joined, service account `manifold.svc`)
- Claude Desktop: latest (2026)
- Node.js install path: `C:\Program Files\nodejs\` (default Windows installer location)
- server-filesystem version: `@modelcontextprotocol/server-filesystem@2026.1.14`

## Description

The filesystem MCP server fails to start on Windows when Node.js is installed in the default location (`C:\Program Files\nodejs`) because the path contains a space. Claude Desktop passes the `command` value through `cmd.exe`, which splits the path at the space and attempts to execute `C:\Program` as a standalone command.

This affects every Windows user who installs Node.js to its default location — which is the majority of users.

## Error

```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

## Steps to Reproduce

1. Install Node.js to default location: `C:\Program Files\nodejs\`
2. Configure `claude_desktop_config.json`:
   ```json
   "filesystem": {
     "command": "npx",
     "args": ["-y", "@modelcontextprotocol/server-filesystem@2026.1.14", "C:\\path\\to\\dir"]
   }
   ```
3. Launch Claude Desktop — server immediately disconnects on every attempt.

## What Was Tried

Every quoting approach via `cmd.exe` fails:

| Config | Error |
|---|---|
| `"command": "npx"` | Claude Desktop resolves to `C:\Program Files\nodejs\npx.cmd`, passes unquoted → splits at space |
| `"command": "cmd.exe", "args": ["/C", "\"C:\\Program Files\\nodejs\\npx.cmd\"", "-y", "..."]` | `The filename, directory name, or volume label syntax is incorrect` |
| Single-string cmd /C with inner quotes | `'"C:\Program Files\nodejs\npx.cmd"' is not recognized` — quotes passed as literals |

## Workaround

Use the Windows 8.3 short path, which has no spaces and requires no quoting:

```json
"filesystem": {
  "command": "C:\\PROGRA~1\\nodejs\\npx.cmd",
  "args": [
    "-y",
    "@modelcontextprotocol/server-filesystem@2026.1.14",
    "C:\\Users\\username\\allowed-directory"
  ]
}
```

`PROGRA~1` is the stable 8.3 short name for `Program Files` on all Windows systems. This has been tested and confirmed working on Windows 11.

## Suggested Fix

The Claude Desktop documentation for Windows should prominently document the 8.3 short path workaround. Alternatively, the server package or Claude Desktop launcher should internally resolve the Node.js path using `where.exe npx` and apply proper quoting before passing to the shell.

## Related Issues

- #447 — covers spaces in *allowed directory paths* (different from this issue, which is about spaces in the *command* path)

tarunag10:
## Summary
- clarify that Windows `npx` MCP configs should keep `npx` in `args` when Node.js is installed under `C:\Program Files\nodejs`
- add an absolute `npx.cmd` example with the executable quoted for `cmd`
- document the WinGet `uvx` shim scheduled-task failure mode and the real-executable workaround

Fixes #4487
Related to #4488

## Validation
- `git diff --check`

## Notes
This is docs-only because the reported failures come from client launch configuration / Windows startup resolution rather than code in this repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,24 @@ Additional examples of using the Claude Desktop as an MCP client might look like
 }
 ```
 
-On Windows, apply the same wrapper to each `npx`-based entry above by changing `"command"` to `"cmd"` and prepending `"/c", "npx"` to the existing `args`. Leave `uvx` entries unchanged.
+On Windows, apply the same wrapper to each `npx`-based entry above by changing `"command"` to `"cmd"` and prepending `"/c", "npx"` to the existing `args`. Keep `npx` in `args` instead of setting `"command"` to the full `npx.cmd` path; this avoids command-path parsing failures when Node.js is installed under `C:\Program Files\nodejs`.
+
+If you need to use the absolute `npx.cmd` path, quote the executable for `cmd`:
+
+```json
+{
+  "command": "cmd",
+  "args": [
+    "/c",
+    "\"C:\\Program Files\\nodejs\\npx.cmd\"",
+    "-y",
+    "@modelcontextprotocol/server-filesystem",
+    "C:\\Users\\username\\Desktop"
+  ]
+}
+```
+
+Leave interactive `uvx` entries unchanged. If a Windows scheduled task or logon-started client fails to launch a `uvx` server from a WinGet shim path such as `%LOCALAPPDATA%\Microsoft\WinGet\Links\uvx.exe`, point `"command"` at the real `uvx.exe` in the WinGet package directory, or install `uv` with an installer that places a normal executable on `PATH`.
 
 ## 🛠️ Creating Your Own Server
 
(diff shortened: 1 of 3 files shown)


ahmedaminebouchmal:
The cmd/npx guidance is useful. Please avoid describing PROGRA~1 as guaranteed on every Windows installation: NTFS 8.3 names may be disabled or unavailable. Keep the npx-in-args approach primary and make the short-path workaround conditional, with a normal executable-on-PATH fallback.

Issue #1481, opened 2025-04-17
Memory Server not adding to JSON Array


bbarrington0099:
**Describe the bug**
Running the server locally with no changes to the repo, have env set to a custom MEMORY_FILE_PATH. I have tried starting the server with the memory.json being both an empty JSON Object and Array. When a new entity is created, it removes the existing structure of the empty file and adds a new entity, then when another entity is created it just adds a new JSON Object so there are 2 JSON Objects in the file but nothing wrapping them so the file is invalid

**To Reproduce**
1. Have one entity created
2. Have a second entity created
3. See that the .json file is errored

**Expected behavior**
I expect multiple entities to go into and be pulled from a valid JSON Array

**Logs**
{"type":"entity","name":"Brandon","entityType":"User","observations":["Primary user as provided by himself."]}
{"type":"entity","name":"MCP","entityType":"Self","observations":["This is the self entity for the assistant, created as per server instructions."]}

**Additional context**
Client created to use OpenAI API, other tools are working as expected, when only one entity is present the Memory tools work as well.

lamba:
In my case the memories being saved (not sure where) because Claude recalls those memories across sessions. But it's not clear where they are being stored. They are certainly not being stored in the location I specified.

Here's my config file

```
{
  "tools": [
    {
      "name": "memory",
      "type": "memory_server",
      "storage_path": "/Users/lamba/Dropbox/Claude-MCP/memory.json"
    }
  ],
  "mcpServers": {
    "memory": {
      "command": "npx",
      "args": [
        "-y",
        "@modelcontextprotocol/server-memory",
        "--file",
        "/Users/lamba/Dropbox/Claude-MCP/memory.json"      
      ],
      "env": {
        "MEMORY_FILE_PATH": "/Users/lamba/Dropbox/Claude-MCP/memory.json"
      }
    },
    "filesystem": {
      "command": "npx",
      "args": [
        "-y",
        "@modelcontextprotocol/server-filesystem",
        "/Users/lamba/Desktop",
        "/Users/lamba/Dropbox"
      ]
    }
  }
}


```

leependu:
I've also been unable to get my `memory.json` saved to the specified path. Here's my config:

```
    "memory": {
      "command": "npx",
      "args": [
        "-y",
        "@modelcontextprotocol/server-memory"
      ],
      "env": {
[...]

PR #4510, opened 2026-07-10
fix(memory): atomic save + per-line parse guard against corruption (#1481)

ZhangJing-gugugaga:
﻿## Problem

Two defects in the memory server (issue #1481, "Memory MCP JSON Parsing Error - All Tools Failing") combine to make **every tool fail on startup** whenever the memory file is even slightly malformed:

1. `saveGraph` overwrote the file in place with `fs.writeFile`. A process crash during the write left a truncated last JSONL line — producing an invalid file.
2. `loadGraph` parsed each JSONL line inside the reducer with **no per-line guard**. A single malformed line aborted the *entire* `loadGraph` call, and since `main()` constructs the `KnowledgeGraphManager` at startup, the server died before it could serve a single request.

This turns a survivable half-write into a total, repeated startup crash.

## Fix (`src/memory/index.ts`)

- `saveGraph` now writes to a `${path}.${pid}.tmp` file and `fs.rename`s it into place — an atomic replace so a crash mid-write can never leave the real file truncated.
- `loadGraph` reads first (so ENOENT still returns an empty graph), then parses each line inside its own `try/catch`. Malformed lines are skipped with a stderr warning instead of killing the server.

## Verification

`cd src/memory && npx vitest run` — **all 30 tests pass** (unchanged, 3 files / 50 tests across the whole memory suite also green). The new code does not alter the JSONL storage schema, so no migration concerns.

Added no new test cases here — a regression test for the truncated-file-resilience path requires truncating a real file on disk and is a good follow-up.

### DIFF
--- a/src/memory/index.ts
+++ b/src/memory/index.ts
@@ -70,10 +70,20 @@ export class KnowledgeGraphManager {
   constructor(private memoryFilePath: string) {}
 
   private async loadGraph(): Promise<KnowledgeGraph> {
+    let data: string;
     try {
-      const data = await fs.readFile(this.memoryFilePath, "utf-8");
-      const lines = data.split("\n").filter(line => line.trim() !== "");
-      return lines.reduce((graph: KnowledgeGraph, line) => {
+      data = await fs.readFile(this.memoryFilePath, "utf-8");
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


ZhangJing-gugugaga:
Superseded: the memory fix was folded into PR #4509 ( branch) for a single review. Closing this duplicate.

PR #4515, opened 2026-07-11
fix(filesystem): write in place to preserve file identity on edits

### LINKED ISSUE #4512: filesystem: `write_file`/`edit_file` destroy file creation time (birthtime) and file identity due to atomic-rename write strategy
## Summary

`write_file` and `edit_file` replace existing files via a temp-file + `fs.rename` pattern (`writeFileContent` and `applyFileEdits` in `src/filesystem/lib.ts`). There is a security rationale in the code comments: atomic rename prevents symlink race conditions (TOCTOU) between path validation and write. However, this strategy has an undocumented side effect: **every write to an existing file creates a new inode**, which
- **destroys the file's creation timestamp** (`birthtime` on macOS/APFS and BSD, creation time on Windows/NTFS, `crtime` on ext4), and
- **breaks file identity**: hard links are severed, and inode-based file watchers/editors tracking the file lose it.

Notably, the server itself treats the creation timestamp as first-class metadata: `get_file_info` returns it as a dedicated `created:` field. Reporting this field but silently destroying it on each of its own write operations is internally inconsistent.

## To reproduce

1. Create a file, note `birthtime` (e.g. `stat -f "%SB" file.md` on macOS, or via the server's own `get_file_info`).
2. Call `edit_file` (or `write_file`) on it.
3. Read `birthtime` again → it now equals the time of the edit.

(`move_file` correctly preserves `birthtime`, since rename keeps the inode.)

## Root cause

`writeFileContent` falls back to temp-file + `rename` for existing files; `applyFileEdits` uses the same pattern unconditionally. `rename` replaces the target with a *different* file, so all creation metadata and identity of the original are lost.

## Suggested fix

Preserve the symlink-safety property while writing in place:

1. `fs.open(filePath, fs.constants.O_RDWR | fs.constants.O_NOFOLLOW)` — fails with `ELOOP` if the path is a symlink, giving the same protection the rename pattern provides (no writes through symlinks swapped in after validation).
2. `fstat` the handle and verify it is a regular file (defense in depth; also allows re-checking the resolved path if desired).
3. `ftruncate` + write through the handle → same inode, `birthtime`, hard links, and watchers all preserved.

On POSIX platforms (macOS, Linux, BSD) `O_NOFOLLOW` is well supported. On Windows, symlink creation requires elevated privileges by default and the threat model differs; the current rename path could be retained there, or `FILE_FLAG_OPEN_REPARSE_POINT` semantics used.

Trade-off: In-place writes lose crash-atomicity (a crash mid-write can leave a partially written file, which the rename pattern avoids). If that property is considered essential, an alternative would be an opt-in mode (env var or tool argument, e.g. `preserveFileIdentity`) selecting the in-place strategy — or at minimum, documenting the metadata-destroying behavior of the current implementation in the tool descriptions.

ayushbunny11:
write_file/edit_file used a temp-file+rename strategy for existing files, which replaces the inode on every write and destroys the file's creation time, hard links, and inode-based file watches.

On POSIX, write in place instead: open with O_RDWR | O_NOFOLLOW to keep the same symlink protection the rename gave us, fstat to confirm the target is still a regular file, then truncate and write. This trades away crash-atomicity (a crash mid-write can now leave a truncated file, where rename left the original untouched).

On win32, keep the rename strategy: O_NOFOLLOW isn't reliably enforced there, and symlink creation already requires elevated privileges by default, so the threat model differs.

Fixes #4512

## Description
`write_file`/`edit_file` used a temp-file+rename strategy for existing files, which replaces the inode on every write and destroys the file's creation time, hard links, and inode-based file watches.

On POSIX, write in place instead: open with `O_RDWR | O_NOFOLLOW` to keep the same symlink protection the rename gave us, `fstat` to confirm the target is still a regular file, then truncate and write. This trades away crash-atomicity (a crash mid-write can now leave a truncated file, where rename left the original untouched).

On win32, keep the rename strategy: `O_NOFOLLOW` isn't reliably enforced there, and symlink creation already requires elevated privileges by default, so the threat model differs.

Fixes #4512

## Server Details

- Server: `filesystem`
- Changes to: tools (`write_file`, `edit_file`)

## Motivation and Context
The server's own `get_file_info` tool reports a file's `birthtime`, but every edit through `write_file`/`edit_file` silently reset it, because the rename-based write replaced the file's inode. This also broke hard links (a link to the file kept pointing at stale pre-edit content) and confused inode-based file watchers, which saw "delete + create" instead of "modify" on every edit.

## How Has This Been Tested?
- `npm run build` passes (typecheck).
- `npm test` passes — 157/157, including new coverage for the EEXIST in-place path, the non-regular-file rejection, and the win32 rename fallback.
- Manually verified on a real filesystem (Linux) that inode and `birthtime` are preserved across an in-place write, and that opening a symlink with `O_NOFOLLOW` correctly fails with `ELOOP`.
- Tested with Claude Code as an MCP client against the built server: registered the local build, ran `edit_file` and `write_file` against a real file with a hard link, and confirmed inode/birthtime stayed constant across edits while the hard link tracked the change.

## Breaking Changes
None. Tool inputs/outputs are unchanged; this only changes the write mechanism on disk. The only observable behavior difference is that crashes mid-write can now leave a truncated file on POSIX, instead of leaving the original untouched.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import fs from 'fs/promises';
+import { constants as fsConstants } from 'fs';
 import path from 'path';
 import os from 'os';
 import {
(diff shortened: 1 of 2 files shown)


0ywfe:
Hi @ayushbunny11 — checking in rather than competing, because I think our two PRs are blocking each other.

This and #4516 implement the same fix for #4512, opened the same day. You posted the approach on the issue before I pushed, so priority is yours; my concern is only that two overlapping PRs give a reviewer an easy reason to defer both, and #4512 has now been open two months with neither merged.

Your call on Windows was the right one and I had it wrong. My original revision used `O_RDWR | O_NOFOLLOW` unconditionally; since `fs.constants.O_NOFOLLOW` is undefined on Windows, that silently collapses to a plain `O_RDWR` open and *removes* the symlink protection while appearing to enforce it. I've since adopted your win32 rename fallback and the `fstat` regular-file check.

Are you still planning to take this forward? Either resolution works for me:

- **You continue #4515.** Rebase it onto `main` (it's currently conflicting — the filesystem code moved a fair bit in the 28 Aug batch) and feel free to lift anything useful from mine: the `__tests__/write-preserves-identity.test.ts` real-filesystem tests, and the handling for #4115, which landed on 28 Aug and restores permissions with a `stat`+`chmod` after the rename — writing in place makes that compensation unnecessary on POSIX, so its tests need retargeting either way. I'll close #4516 and you'll have no argument from me.
- **I continue #4516.** It's rebased, green, and carries the above. Happy to credit you in the commit message for the Windows catch.

I genuinely don't mind which. If I don't hear back in a week or so I'll assume you'd rather not carry it and will leave mine open, but I'd rather ask than assume.

ayushbunny11:
Hi @0ywfe. 
Go ahead with #4516 since it is already rebased, green and has the necessary tests in place. I am fine with the credits in the commit message.

ayushbunny11:
Closing the PR. For more information can check #4516. Thanks @0ywfe.

0ywfe:
Thanks @ayushbunny11 — appreciated, and sorry it took a duplicate standoff to get there.

The credit is now in the commit message on #4516: the Windows problem is attributed to you, with the note that an earlier revision of mine had the unconditional `O_NOFOLLOW` open and would have dropped the symlink protection on win32 while appearing to enforce it. That was a real defect and you caught it before any reviewer did.

The `fstat` regular-file check from your read of the issue is in there too.

PR #4520, opened 2026-07-13
git: clarify when NOT to call status/diff/log — cuts agent extra-call rate (measured +21.7 pts)

sujugithub:
## What

Rewrites 7 tool description strings in `src/git/src/mcp_server_git/server.py`. **No code or behavior changes** — only the text agents read when choosing tools.

## Why

I measured how a mid-tier agent actually uses this server (12 realistic scenarios × 5 runs each, live server, every tool call recorded). Tool *selection* is already excellent — 100% hit rate, the diff-trio naming works. The failure mode is **ritual verification calls**: after `git_add`, agents call `git_status` or `git_diff_staged` to check it worked; before `git_show HEAD`, they call `git_log` to find a hash they don't need. A quarter of all runs were padded this way.

## Measured effect of this diff (nothing else changed)

| metric | before | after |
|---|---|---|
| tool hit rate | 100% | 100% |
| argument correctness | 100% | 100% |
| strict success (right tool + args + no extra calls) | 75.0% | **96.7%** |

Method: agent = gpt-oss-120b (deliberately mid-tier — weaker agents are hurt most by underspecified descriptions), N=5 runs per scenario, descriptions applied via in-memory override so before/after used the identical server build. Happy to share the full per-run data and scenario suite, or re-run against any revision of the wording.

## Notes

- The pattern in each rewrite: state that the operation is complete on success ("does not require calling git_status to verify"), and say which sibling tool covers the adjacent intent.
- One deviation from the measured text: the tested `git_show` wording implied `revision` was optional; since `GitShow.revision` is required, this PR says `Pass revision=\"HEAD\"` instead — same intent, schema-accurate.
- The measurement harness (scenarios + runner) is being open-sourced; I'll link it here once public.

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -323,7 +323,7 @@ async def serve(repository: Path | None) -> None:
         return [
             Tool(
                 name=GitTools.STATUS,
-                description="Shows the working tree status",
+                description="Shows the current working tree and staged changes status. Use this only when the user explicitly asks for the repository status; do not use to verify the results of git_add, git_reset, or git_checkout.",
                 inputSchema=GitStatus.model_json_schema(),
                 annotations=ToolAnnotations(
                     readOnlyHint=True,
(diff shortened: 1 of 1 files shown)


sujugithub:
The measurement harness is now public: https://github.com/sujugithub/hitrate — the git scenario suite is [`scenarios/git/git.yaml`](https://github.com/sujugithub/hitrate/blob/main/scenarios/git/git.yaml) and the per-run data behind the table above is in [`results/`](https://github.com/sujugithub/hitrate/tree/main/results) (files dated 2026-07-13, `git-*`). Happy to re-run against any revised wording.

PR #4531, opened 2026-07-15
fix(filesystem): reject existing move destinations

dev-willbird1936:
## Description

Make the filesystem server's `move_file` honor its documented no-overwrite contract. Existing destination files, directories, and symlinks now return a tool error, including when another process creates the destination during the move.

## Server Details

- Server: filesystem
- Changes to: `move_file`

## Motivation and Context

POSIX `rename()` can replace an existing destination. A separate existence check followed by `rename()` still has a check-to-use race, so a concurrent destination could be overwritten. The implementation now copies with exclusive destination creation and removes the source only after the copy succeeds.

## How Has This Been Tested?

- Focused `structured-content.test.ts` suite: 13 passed.
- Full filesystem package suite: 155 passed.
- Filesystem package TypeScript build: passed.
- `git diff --check`: passed.
- LLM-client integration was not exercised; tests call the server tool through the MCP client test harness.

## Breaking Changes

None to the documented API. A successful move now uses copy-then-remove rather than an atomic same-filesystem rename so that portable Node.js code can enforce no replacement. If copying fails, the source is retained.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have read the MCP Protocol Documentation
- [ ] I have updated the server's README accordingly (not required; the documented contract is unchanged)
- [ ] I have documented all environment variables and configuration options (not applicable)

## Additional context

`fs.cp()` is called with `force: false` and `errorOnExist: true`, which makes destination creation exclusive and closes the overwrite race. Directory success and rejection without tree merging are covered by regression tests.

### DIFF
--- a/src/filesystem/__tests__/structured-content.test.ts
+++ b/src/filesystem/__tests__/structured-content.test.ts
@@ -122,6 +122,81 @@ describe('structuredContent schema compliance', () => {
       // The content should contain success message
       expect(structuredContent.content).toContain('Successfully moved');
     });
+
+    it('should reject an existing destination without replacing it', async () => {
+      const sourcePath = path.join(testDir, 'source.txt');
+      const destPath = path.join(testDir, 'existing.txt');
+      await fs.writeFile(sourcePath, 'source content');
+      await fs.writeFile(destPath, 'existing content');
+
+      const result = await client.callTool({
+        name: 'move_file',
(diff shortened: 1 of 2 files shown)


olaservo:
Closing in favour of #4630, which has been merged for #4628. Thanks for the PR.

PR #4289, opened 2026-06-07
fix: declare zod runtime dependencies

he-yufeng:
Fixes #4288.

## Summary

`@modelcontextprotocol/server-sequential-thinking` imports `zod` directly from its runtime entrypoint, but the package did not declare `zod` as a dependency. With strict package isolation, consumers can install the server without a local `zod`, then hit `ERR_MODULE_NOT_FOUND` at runtime.

While checking the same dependency boundary, I found two other published server packages with the same direct runtime import:

- `@modelcontextprotocol/server-memory`
- `@modelcontextprotocol/server-filesystem`

This patch declares `zod` in those three packages, using the same `^4.0.0` range already used by `@modelcontextprotocol/server-everything`.

## Validation

- `npm install --package-lock-only --ignore-scripts`
- `npm install --ignore-scripts`
- `npm run build --workspace=@modelcontextprotocol/server-sequential-thinking`
- `npm run build --workspace=@modelcontextprotocol/server-memory`
- `npm run build --workspace=@modelcontextprotocol/server-filesystem`
- `npm test --workspace=@modelcontextprotocol/server-sequential-thinking`
- `npm test --workspace=@modelcontextprotocol/server-memory`
- `npm test --workspace=@modelcontextprotocol/server-filesystem`
- `npm exec prettier -- --check src/filesystem/package.json src/memory/package.json src/sequentialthinking/package.json package-lock.json`
- `git diff --check`

Issue #4288, opened 2026-06-07
zod declared in both dependencies and peerDependencies causes ERR_MODULE_NOT_FOUND with pnpm strict isolation


GrzywN:
**Describe the bug**
zod is declared in both `dependencies` and `peerDependencies` in package.json. With strict package managers (pnpm with enableGlobalVirtualStore or default strict isolation), the peer dependency declaration takes precedence — pnpm does not co-locate zod alongside the package in the virtual store. Packages that depend on `@modelcontextprotocol/sdk` but do not explicitly declare `zod` themselves (e.g. `@modelcontextprotocol/server-sequential-thinking`) fail at runtime with ERR_MODULE_NOT_FOUND when the SDK tries to import `zod` via ESM.

**To Reproduce**
1. Create a project using pnpm with `enableGlobalVirtualStore: true` (or default strict mode)
2. Install any package that depends on @modelcontextprotocol/sdk without declaring zod itself:
`pnpm add @modelcontextprotocol/server-sequential-thinking`
3. Run the package:
`pnpm exec mcp-server-sequential-thinking` or `pnpx -y @modelcontextprotocol/server-sequential-thinking@2025.12.18`

**Expected behavior**
Package starts successfully. zod is a declared dependency of the SDK — it should be available at runtime without the consumer explicitly installing it.

**Logs**
```
❯ pnpm exec mcp-server-sequential-thinking
node:internal/modules/run_main:107
    triggerUncaughtException(
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'zod' imported from /home/user/.pnpm-store/v11/links/@modelcontextprotocol/server-sequential-thinking/2025.12.18/b16d7229fe63eba8f5c9df434194638ecedcbeff81abdf52f136fde8d100ab1e/node_modules/@modelcontextprotocol/server-sequential-thinking/dist/index.js
Did you mean to import "zod/index.cjs"?
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:301:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:859:18)
    at defaultResolve (node:internal/modules/esm/resolve:992:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:691:20)
    at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:708:38)
[...]

PR #4539, opened 2026-07-20
fix(memory, sequentialthinking, filesystem): declare zod as a dependency (#4330)

### LINKED ISSUE #4330: server-memory and server-sequential-thinking fail at startup due to missing zod dependency
**Describe the bug**

The published npm packages `@modelcontextprotocol/server-memory` and `@modelcontextprotocol/server-sequential-thinking` fail immediately on startup because they import `zod`, but `zod` is not listed in their published dependencies.

This reproduces outside of Continue and appears to be a package publishing/dependency issue.

**To Reproduce**

1. Install Node.js 22
2. Run:

```bash
pnpm dlx @modelcontextprotocol/server-memory
```

or

```bash
pnpm dlx @modelcontextprotocol/server-sequential-thinking
```

3. Observe that the process exits immediately with a module resolution error.

**Expected behavior**

The MCP server should start successfully and wait for stdio connections.

**Logs**

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'zod' imported from .../dist/index.js

Did you mean to import "zod/index.cjs"?
```

Environment:

```text
Node.js v22.22.3
pnpm v11.3.0
Arch Linux
```

**Additional context**

The published package contains imports from `zod`:

```bash
grep -R 'from "zod"' node_modules/@modelcontextprotocol/server-memory
```

Output:

```text
node_modules/@modelcontextprotocol/server-memory/dist/index.js:import { z } from "zod";
```

However, the published dependencies do not include `zod`:

```bash
pnpm view @modelcontextprotocol/server-memory dependencies
```

Output:

```json
{
  "@modelcontextprotocol/sdk": "^1.25.2"
}
```

I observed the same behavior with `@modelcontextprotocol/server-sequential-thinking`.

gianghungtien:
## Description

`memory`, `sequentialthinking` and `filesystem` all `import { z } from "zod"` but none of them declared `zod` in `package.json`. This PR declares it, and adds a regression test so a phantom dependency cannot silently return.

Fixes #4330.

## Server Details

- Server: `memory`, `sequentialthinking`, `filesystem`
- Changes to: `package.json` dependencies (+ tests). No tool, resource, or prompt behaviour changes.

## Motivation and Context

Each of the three servers imports `zod` directly:

| Server | Import |
|---|---|
| memory | `src/memory/index.ts:6` |
| sequentialthinking | `src/sequentialthinking/index.ts:5` |
| filesystem | `src/filesystem/index.ts:13` |

None declared it. The import resolves only by accident: `@modelcontextprotocol/sdk` depends on `zod`, and npm's hoisted `node_modules` layout happens to place it somewhere these packages can reach. That is a phantom dependency — it works until the layout changes.

Under a strict layout (pnpm without hoisting, yarn PnP) it is not reachable, and the servers die on startup. Reproduced against the current published packages:

```console
$ cat .npmrc
node-linker=isolated
hoist=false

$ pnpm add @modelcontextprotocol/server-memory \
           @modelcontextprotocol/server-sequential-thinking \
           @modelcontextprotocol/server-filesystem

$ node node_modules/@modelcontextprotocol/server-memory/dist/index.js
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'zod' imported from
  .../@modelcontextprotocol/server-memory/dist/index.js
```

All three fail identically. The issue reports `memory` and `sequentialthinking`; `filesystem` has the same defect and is included here.

Note this is not only a strict-layout concern. The SDK's zod range is `^3.25 || ^4.0`, so even under npm the servers silently bind to whichever major the SDK happens to hoist — a future SDK release could flip that under them.

### Why `^4.0.0`

- It matches the `everything` server, the one TS server that already declared `zod`, so the repo stays consistent.
- It sits inside the SDK's `^3.25 || ^4.0` peer range, so `zod` still dedupes to **one** instance shared with the SDK. That matters: two zod copies would make schemas built here unrecognisable to the SDK's `instanceof` checks. Verified after the fix — a single `zod@4.4.3` is installed and linked.

## How Has This Been Tested?

**Reproduction, before and after.** Packed all three servers from this branch and installed the tarballs into a clean project using the same strict config that fails above (`node-linker=isolated`, `hoist=false`):

```console
$ node node_modules/@modelcontextprotocol/server-memory/dist/index.js
Knowledge Graph MCP Server running on stdio
$ node node_modules/@modelcontextprotocol/server-sequential-thinking/dist/index.js
Sequential Thinking MCP Server running on stdio
$ node node_modules/@modelcontextprotocol/server-filesystem/dist/index.js .
Secure MCP Filesystem Server running on stdio
```

All three now start over stdio where all three previously aborted.

**Regression test.** Each package gets `__tests__/dependencies.test.ts`, which scans the sources that compile into `dist/` for bare import specifiers and asserts each is declared in `dependencies` (relative paths and Node builtins are skipped, deep/scoped specifiers reduce to the package name). Confirmed it actually catches the bug by removing `zod` from `src/memory/package.json` and re-running:

```console
AssertionError: expected [ '@modelcontextprotocol/sdk' ] to deeply equal ArrayContaining{…}
- ArrayContaining [
+ [
    "@modelcontextprotocol/sdk",
-   "zod",
  ]
```

It is written with vitest, per CONTRIBUTING.md.

**Suites and build.** `npm ci` from a clean tree (so the lockfile is verified the way CI installs it), `npm run build`, then per-package `npm test`:

| Server | Result |
|---|---|
| memory | 4 files, 52 tests passed |
| sequentialthinking | 2 files, 17 tests passed |
| filesystem | 8 files, 158 tests passed |

`tsc --noEmit` is clean for all three.

**Not tested with an LLM client.** This change adds no tools and alters no schemas or behaviour — the servers either start or they do not, which is what the reproduction above covers end to end. Happy to run a client check if you would like one anyway.

## Breaking Changes

None. No client configuration changes. Existing installs on hoisted npm layouts are unaffected, since they already resolve to the same zod v4 — the declaration just makes that guaranteed instead of incidental.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have updated the server's README accordingly — not applicable, no user-facing configuration or behaviour change
- [ ] I have tested this with an LLM client — see the note above; verified by launching each server over stdio under the layout that reproduces the bug
- [ ] I have documented all environment variables and configuration options — not applicable, none added

## Additional context

`package-lock.json` is regenerated to record the new declarations for the three workspaces; validated with a clean `npm ci`.

The regression test is duplicated per package rather than shared from a common helper, since each server is published standalone and its `tsconfig.json` sets `rootDir` to the package directory — reaching outside it would break the build. Tests are already excluded from compilation via `**/*.test.ts`, so nothing new ships in `dist/`.

Worth noting the same class of bug is easy to reintroduce elsewhere; the test is deliberately generic, so dropping it into another server would guard that one too.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3835,7 +3835,8 @@
         "@modelcontextprotocol/sdk": "^1.29.0",
         "diff": "^8.0.3",
         "glob": "^10.5.0",
-        "minimatch": "^10.0.1"
+        "minimatch": "^10.0.1",
+        "zod": "^4.0.0"
       },
       "bin": {
         "mcp-server-filesystem": "dist/index.js"
@@ -3855,7 +3856,8 @@
       "version": "0.6.3",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
-        "@modelcontextprotocol/sdk": "^1.29.0"
+        "@modelcontextprotocol/sdk": "^1.29.0",
+        "zod": "^4.0.0"
       },
       "bin": {
         "mcp-server-memory": "dist/index.js"
@@ -3875,7 +3877,8 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
         "chalk": "^5.3.0",
-        "yargs": "^17.7.2"
+        "yargs": "^17.7.2",
+        "zod": "^4.0.0"
       },
       "bin": {
         "mcp-server-sequential-thinking": "dist/index.js"
--- a/src/filesystem/__tests__/dependencies.test.ts
+++ b/src/filesystem/__tests__/dependencies.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync, readdirSync } from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { builtinModules } from 'module';
+
+/**
+ * Guards against phantom dependencies: modules that are imported by the shipped
+ * source but not declared in package.json. Such imports resolve by accident
+ * under npm's hoisted node_modules layout, then fail at startup for anyone
+ * using a strict layout (pnpm, yarn PnP) with ERR_MODULE_NOT_FOUND.
+ */
+describe('declared dependencies', () => {
+  const packageDir = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+
+  const packageJson = JSON.parse(
+    readFileSync(path.join(packageDir, 'package.json'), 'utf-8')
+  ) as { dependencies?: Record<string, string> };
+
+  // Source files that get compiled into dist/, i.e. everything except tests and configs.
+  const sourceFiles = readdirSync(packageDir)
+    .filter((file) => file.endsWith('.ts'))
+    .filter((file) => !file.endsWith('.test.ts') && file !== 'vitest.config.ts');
+
+  const IMPORT_PATTERN = /(?:import|export)[\s\S]*?from\s*['"]([^'"]+)['"]|import\s*\(\s*['"]([^'"]+)['"]\s*\)/g;
+
+  /** Reduces a specifier like `zod/v4` or `@scope/pkg/sub` to its package name. */
+  const toPackageName = (specifier: string): string => {
+    const segments = specifier.split('/');
+    return specifier.startsWith('@') ? segments.slice(0, 2).join('/') : segments[0];
+  };
+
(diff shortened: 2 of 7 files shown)


olaservo:
#4289 has landed (`ba167cef`) with the same `zod` declarations for the three packages, and #4288 is closed. The phantom-dependency test in this PR is a reasonable idea on its own; if you want to pursue it, a fresh PR with only the test against current `main` would be easy to review. Closing this one as superseded.

PR #4555, opened 2026-07-27
fix(memory): serialize graph mutations to prevent concurrent write race

### LINKED ISSUE #1819: Memory server race condition causes corrupted JSON
**Describe the bug**
calling tool multiple times can cause the json file to become corrupted due to writing malformed data

**To Reproduce**
trigger multiple calls to Memory from a single LLM response (e.g. create entity, delete observation, add observation, and create relation)

**Expected behavior**
It should not be possible for the tool to write corrupted data to fle.

_Ideally_, it would correctly handle concurrent writes to ensure atomic operations

_minimally_, it would validate any data before being written, to ensure that malformed data is not saved and it fails gracefully.

**Additional context**
Currently This has to be resolved manually every time, once the file is corrupted it is not possible for the tool to recover/fix by itself, every operation fails

Nandini-Inuguru:
## Problem

Every mutation method in `KnowledgeGraphManager` (`createEntities`,
`createRelations`, `addObservations`, `deleteEntities`,
`deleteObservations`, `deleteRelations`) independently does
`load → mutate → save`, with no synchronization between calls.

When two tool calls land close together (e.g. multiple mutations
dispatched from a single LLM turn), both read the same starting
state, mutate their own in-memory copy, and write it back —
whichever write lands last silently discards the other's changes.
Interleaved writes to the same file can also corrupt it outright.

Fixes #1819

## Fix

Adds an in-process async mutex (`KnowledgeGraphManager.withLock`)
that serializes all six mutation methods through a single queue, so
only one load→mutate→save sequence runs at a time. Read-only methods
(`readGraph`, `searchNodes`, `openNodes`) are untouched — this is
scoped to the write race specifically, not a broader concurrency
rewrite.

## Testing

Added three regression tests under a new `concurrent mutations`
describe block, firing multiple mutation calls without awaiting each
other (simulating concurrent tool calls):

- Concurrent `createEntities` calls don't lose entities
- Concurrent `createRelations`/`addObservations` calls don't lose
  data
- 25 concurrent `createEntities` calls produce valid JSONL with no
  corrupted lines

**Verified the tests actually catch the bug**: temporarily reverted
`index.ts` to the version on `main` and reran the same three tests —
all three fail (lost entities: 20 expected → 10 received; lost
relations: 2 expected → 1 received; corrupted JSONL causing a
`JSON.parse` `SyntaxError`). Restored the fix and all 39 tests
(36 existing + 3 new) pass.

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)

## Test Coverage
- [x] I added/updated unit tests

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [ ] I have updated documentation if needed


olaservo (approved):
Looks good to me, This closes the lost-update part of #1819, and #4642 already covers the torn-write part.
 I rebased from main and fixed conflicts.

Issue #4475, opened 2026-07-05
Decompose PR #4452 (v2 TypeScript SDK migration) into scoped issues with a tracking issue


cliffhall:
## Goal

Decompose PR #4452 (**feat: MCP v2 (draft)** — migrates the TypeScript servers to the v2 MCP TypeScript SDK; 52 files, +295/−802, explicitly do-not-merge) into **scoped, individually-testable issues on the V2 board, under a single tracking issue** — then re-implement each one deliberately as its own PR, using #4452's diff as the reference implementation.

This follows the pattern we used successfully on the inspector for a similarly-oversized PR: see [modelcontextprotocol/inspector#1579](https://github.com/modelcontextprotocol/inspector/issues/1579) ("decomposing it into scoped issues and re-implementing each one deliberately, using the PR's diff as the reference implementation"). #4452 stays open as the reference while the work proceeds, and is **ultimately closed in favor of the rewrite** — none of its commits merge directly.

## Branch discipline (differs from normal work)

- All of this work happens on the **`v2/main` branch** — every sub-issue must state explicitly that **its PR targets `v2/main`, not `main`**.
- ⚠️ Because `v2/main` is not the default branch, `Closes #N` in those PR bodies is only a cross-reference — it will **not** auto-close the issue on merge (the same caveat inspector documents for its `v2/main`). On merge: manually close the issue and move its card to Done. Keep the `Closes #N` line anyway so issues close if/when `v2/main` merges to `main`.
- Prerequisite: push the `v2/main` branch to the repo (it exists locally) and confirm CI runs against it (`typescript.yml`/`python.yml` trigger on all pushes, so it will).

## What the decomposition should produce

1. **One tracking issue** (modeled on inspector#1579): background, link to #4452 as reference (pin the head SHA so the reference is stable), the sub-issue list grouped into waves — independent issues first so they can be worked in parallel, dependent ones after.
2. **Scoped sub-issues**, each with: a clearly-defined scope of *what* it changes, pointers to the relevant files/symbols in #4452's diff, its own test expectations, the `v2` label, a card on board 43, and the **targets-`v2/main`** note.

### Starting sketch of the axes (to be refined during decomposition)

The PR's own file distribution suggests the grouping:

| Area | Files in #4452 | Notes |
|---|---|---|
| `src/everything` | 41 (+106/−578) | The bulk — likely splits further (SDK server/registration API migration, transports, tools, resources/prompts, tests) |
[...]

PR #4563, opened 2026-07-28
fix(fetch): cap mcp dependency below 2.0

### LINKED ISSUE #4560: fetch: ImportError on startup with mcp SDK 2.0.0 (McpError renamed to MCPError)
### Describe the bug

`mcp-server-fetch` fails to start with an `ImportError` when resolved against
`mcp` SDK 2.0.0 (released 2026-07-28).

### To Reproduce

```
uvx mcp-server-fetch
```

### Actual behavior

```
  ...
  File ".../mcp_server_fetch/__init__.py", line 1, in <module>
    from .server import serve
  File ".../mcp_server_fetch/server.py", line 6, in <module>
    from mcp.shared.exceptions import McpError
ImportError: cannot import name 'McpError' from 'mcp.shared.exceptions'.
Did you mean: 'MCPError'?
```

The process exits with code 1 before the initialize handshake, so clients
simply report the server as unavailable with no visible cause.

### Root cause

`mcp` 2.0.0 renamed the exception class `McpError` to `MCPError`.
`src/fetch/pyproject.toml` declares `mcp>=1.1.3` with no upper bound, so any
fresh resolution (e.g. `uvx`, which does not pin) now pulls 2.0.0 and breaks.

The last fetch release (2026.7.10) predates the SDK 2.0.0 release, so every
user on an unpinned launcher is affected as of 2026-07-28.

### Suggested fix

Short term: cap the dependency, e.g. `mcp>=1.1.3,<2`.
Longer term: migrate to the 2.x API (`MCPError`), as done for `everything`
in #4551.

### Workaround for other users

```
uvx --with "mcp<2" mcp-server-fetch
```

Verified: initialize handshake succeeds, serverInfo reports mcp-fetch 1.29.0.

### Environment

- OS: Windows 11
- Launcher: uv 0.11.8 (uvx, unpinned)
- mcp-server-fetch 2026.7.10 / mcp 2.0.0

guptaishaan:
Fixes #4560

`mcp` 2.0.0 renamed `McpError` to `MCPError`, so `server.py:6` raises `ImportError` at
module import and the process exits before the initialize handshake. `src/fetch` declared
`mcp>=1.1.3` with no upper bound, so unpinned launchers like `uvx` started resolving 2.0.0
as soon as it shipped.

This caps the dependency at `mcp>=1.1.3,<2` and regenerates `uv.lock`. Resolution is
unchanged at mcp 1.23.0, only the recorded specifier moves.

I deliberately did not add an alias import. In 2.0.0 the constructor changed too, from
`McpError(ErrorData(code=..., message=...))` to `MCPError(code, message, data=None)`, so
importing `MCPError` under the old name would import cleanly and then mis-construct all 8
call sites. That trades a loud startup crash for quiet runtime breakage. The real 2.x
migration is a separate job for `v2/main` per #4475.

Also adds a regression test that parses `pyproject.toml` and asserts the declared `mcp`
specifier does not admit 2.0.0. It fails on the unbounded specifier and passes with the
cap. It needs no new dev dependency, `packaging` already comes in with pytest, and
`tomllib` is fine because CI reads `.python-version`, which is 3.11 here.

Verified on Linux, Python 3.11 and 3.12:

- Reproduced the exact traceback from the issue in a clean venv with `mcp==2.0.0`, same
  line, same `Did you mean: 'MCPError'?`.
- Confirmed the new test fails before the cap and passes after.
- Fresh unpinned install of the patched package now resolves mcp 1.29.0, and a real stdio
  `initialize` handshake returns `serverInfo` `mcp-fetch 1.29.0` with exit 0 and empty
  stderr. That matches what the reporter saw with the `uvx --with "mcp<2"` workaround.
- `ruff check` and `pyright` clean. `uv lock --check` and `uv sync --frozen` clean.

Not verified: Windows 11 / uv 0.11.8, the reporter's exact environment, and the literal
`uvx mcp-server-fetch` against a published release rather than a local fresh resolve.
Nothing in the change is platform-sensitive. This patch does not make fetch work on mcp
2.x, it only stops 2.x from being selected.

Unrelated pre-existing failure, present before this change and left alone:
`test_empty_content_returns_error` fails when Node.js is absent, because `readabilipy`
falls back to pure-Python mode.

Separately, and not changed here since this issue is scoped to fetch: `src/git`
(`mcp>=1.0.0`) and `src/time` (`mcp>=1.23.0`) are also unbounded and exposed to the same
resolution.

Thanks to @Jason26214 for the report, which included the root cause, the exact traceback,
and a working `uvx --with "mcp<2"` workaround.


olaservo:
Thanks for catching this first, you filed the cap the same day `mcp` 2.0.0 landed. Fixed in #4663, released as `2026.8.18`, which caps `mcp<2` across fetch, git and time and raises the floor to 1.29.0 so the resolved range also clears GHSA-vj7q-gjh5-988w, GHSA-jpw9-pfvf-9f58 and GHSA-hvrp-rf83-w775. Closing as superseded.

PR #4565, opened 2026-07-28
fix(fetch): port to mcp SDK v2

### LINKED ISSUE #4560: fetch: ImportError on startup with mcp SDK 2.0.0 (McpError renamed to MCPError)

jayzuccarelli:
Fixes #4560

mcp 2.0.0 is more than the `McpError` -> `MCPError` rename: the low-level
`Server` decorator API is gone, replaced by constructor callbacks with new
handler signatures and result-object returns, and `MCPError` now takes
`code`/`message` directly instead of `ErrorData`. So the server can't be
fixed with an alias import - it needs a port.

This ports src/fetch to the v2 API:
- handlers move to `on_list_tools` / `on_call_tool` / `on_list_prompts` /
  `on_get_prompt`, returning `ListToolsResult` / `CallToolResult` /
  `ListPromptsResult` / `GetPromptResult`
- `raise MCPError(code=..., message=...)` at the existing raise sites,
  message text unchanged
- `mcp>=2,<3` in pyproject, uv.lock regenerated
- tests updated for the exception rename

Tool/prompt descriptions, robots.txt handling and truncation behavior are
unchanged. #4563 caps `mcp<2` as immediate triage; this is the follow-up
that moves fetch onto the new SDK, same shape as #4564 does for src/git.

Verified against mcp 2.0.0: `uv run pytest` (20 passed), `uv run --frozen
pyright` (clean), and a live stdio session (initialize, tools/list,
tools/call, prompts/get, plus MCPError propagation on invalid params).


jayzuccarelli:
Checking in on this one. 16 checks green and mergeable as-is.

The change ports `fetch` to SDK v2. I can walk through any part of the diff, or split it into smaller steps if that would be easier to review.

jayzuccarelli:
Rebased on main (5b2bad9d). Upstream pinned `mcp<2` in the meantime and added a README note that the v2 port is in progress; this PR is that port, so the note is dropped here and the pin goes back to `mcp>=2,<3` (lock resolves to 2.2.0). 20 tests, pyright and ruff green.

PR #4576, opened 2026-07-29
fix(sequentialthinking): read server version from package.json

KarlLeen:
Fixes #4575

Read `serverInfo.version` from `package.json` at runtime instead of hardcoding `"0.2.0"`.

Published `@modelcontextprotocol/server-sequential-thinking@2026.7.4` still ships `dist/index.js` with `version: "0.2.0"` while its `package.json` is `"2026.7.4"`. Same class of bug as #4406 for memory; related to #360.

## Changes
- Add `version.ts` with `resolvePackageVersion()` (source + `dist/` layouts)
- Wire `McpServer` to `SERVER_VERSION`
- Tests: not `"0.2.0"`, optional dist resolve, stdio `initialize` smoke via SDK client

## How Has This Been Tested?
- `cd src/sequentialthinking && npm run build && npm test` → 17 passed
- `rm -rf dist && npm test` → 15 passed, 2 skipped
- Stdio smoke: SDK `Client` + `StdioClientTransport` against `dist/index.js`; `getServerVersion().version` matches `package.json`

## Note
Aligns `serverInfo` with `package.json` (in-tree `0.6.2`). CalVer on the next publish follows whatever release writes into `package.json` (see #3870).

## Server Details
- Server: sequentialthinking
- Changes to: serverInfo version reporting

## Breaking Changes
None.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] I have read the MCP Protocol Documentation
- [x] My changes follows MCP security best practices
- [x] I have tested this with an LLM client (stdio initialize smoke)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally

Issue #4575, opened 2026-07-29
sequentialthinking: serverInfo.version hardcoded as "0.2.0" (published package is 2026.7.4)


KarlLeen:
**Describe the bug**
`serverInfo.version` returned during MCP initialization for `@modelcontextprotocol/server-sequential-thinking` is hardcoded as `"0.2.0"` in source/`dist`, so it does not match the installed npm package version. Same class of bug as #4406 for memory; related to the broader sync request in #360.

**To Reproduce**
1. Inspect the published package (or current `main`):
   ```
   npm pack @modelcontextprotocol/server-sequential-thinking@2026.7.4
   tar -xOf modelcontextprotocol-server-sequential-thinking-2026.7.4.tgz package/dist/index.js | rg 'version:'
   tar -xOf modelcontextprotocol-server-sequential-thinking-2026.7.4.tgz package/package.json | rg '"version"'
  
2. Observe `dist/index.js` contains `version: "0.2.0"` while `package/package.json` has `"version": "2026.7.4"`.
3. Or start the server and send `initialize` — `serverInfo.version` is `"0.2.0"`.

On `main` (`src/sequentialthinking/index.ts`):
```ts
const server = new McpServer({
  name: "sequential-thinking-server",
  version: "0.2.0",
});
```

**Expected behavior**
`serverInfo.version` should match the package version from `package.json` (in-tree currently `0.6.2`; published CalVer such as `2026.7.4` after release bumps that field).

**Logs**
Published `@modelcontextprotocol/server-sequential-thinking@2026.7.4`:
- `package.json` → `"version": "2026.7.4"`
- `dist/index.js` → `version: "0.2.0"`

**Additional context**
- Root cause: the version string is hardcoded in `src/sequentialthinking/index.ts` and compiled into `dist/`, so calendar-versioned releases keep advertising `0.2.0`.
- Impact: confusing when diagnosing client/server issues — reported version looks like a very old release even though the installed package is current.
- Suggested fix: read the version from `package.json` at runtime (source + `dist/` layouts), as proposed for memory in #4406 / PRs such as #4407.
- Note:  this aligns `serverInfo` with whatever is in `package.json`; keeping that field bumped on CalVer release is separate (see #3870). Related to the broader sync ask in #360.
- Happy to send a PR.

Medsaad:
This issue exists in all TS servers in the repo. I'll try to add a another PR that address these at onece.

PR #4557, opened 2026-07-27
fix: read memory version from package.json, add missing zod deps

Timelovers:
Fixes #4406, fixes #4330.

Two small fixes:

1. **memory server version** — reads `serverInfo.version` from `package.json` at startup instead of hardcoding `"0.6.3"`, so it stays in sync with the published package version.

2. **missing zod dependency** — `zod` is imported at runtime by both `server-memory` and `server-sequential-thinking` but was only available transitively through `@modelcontextprotocol/sdk`. With strict package managers (pnpm), this causes `ERR_MODULE_NOT_FOUND` at startup. Added `zod: "^3"` to both `dependencies`.

Issue #4463, opened 2026-07-03
v2 release pipeline, Phase 1: OIDC trusted publishing — ✅ complete (validation release 2026.7.4)


cliffhall:
## Scope

Phases 0–1 of the release-pipeline rebuild: recover from the June publish failures, move npm publishing to **OIDC trusted publishing** (tokenless), and validate it with a production release. **✅ Complete** — validation release [`2026.7.4`](https://github.com/modelcontextprotocol/servers/releases/tag/2026.7.4) published all four TypeScript packages via OIDC with provenance attestations, ~2 minutes dispatch-to-release.

**Phase 2 — semver via changesets (TS) + GitHub-Release-triggered publishing — continues in #4472.**

## Background

The `@modelcontextprotocol/*` npm packages are bound to [trusted publishers](https://docs.npmjs.com/trusted-publishers):

| Setting | Value |
|---|---|
| Publisher | GitHub Actions |
| Repository | `modelcontextprotocol/servers` |
| **Workflow filename** | **`release.yml`** |
| **Environment** | **`release`** |
| Allowed actions | `npm publish` |

npm only accepts a publish whose OIDC token attests it came from a `release.yml` run using the `release` environment — `release.yml` is the **single publish path**.

### What the June incidents exposed

1. **Dead publish paths**: `typescript.yml`/`python.yml` had `release:`-triggered publish jobs that *never fired* (the cron created releases with `GITHUB_TOKEN`, which suppresses workflow triggers) and could never succeed under OIDC. The TS one still referenced the dead `NPM_TOKEN`.
2. **Tag-before-test**: the cron tagged and pushed versions before running a single test, stranding versions on any failure (June 3, June 16 — see #4465 for the abandoned recovery).
3. **Phantom versions**: version bumps lived only on orphaned tag commits — main's package.json files were never touched.
4. **Approval-delay bugs**: environment-approval waits crossing midnight UTC produced a tag (`2026.6.3`) whose packages were stamped `2026.6.4`; failed runs sat perpetually "waiting" — 31 zombie runs accumulated.
5. **Expired token**: both June attempts failed every npm leg with a disguised auth error (`E404` on the publish PUT); the token could not be replaced (tokenless policy adopted instead).

## Phase 0 — hygiene ✅

- [x] Close #4465 (retry of the stranded June versions abandoned — tokenless migration; `2026.6.4`/`2026.6.16` deliberately never exist on npm)
[...]

PR #4608, opened 2026-08-04
Fix: TS server version drift

Medsaad:
> **Draft .. parked until #4576 lands.** That PR establishes the pattern for `sequentialthinking`; this one applies it to the remaining three TypeScript servers. Opening it early so the work is visible and nobody duplicates it. I'll mark it ready once #4576 merges, and rebase if its implementation changes during review.

## Description

Follow-up to #4576, as suggested by @KarlLeen there. `filesystem`, `memory`, and `everything` hardcode a version literal in their `McpServer` constructor .. the same defect #4575 reports for `sequentialthinking`. Those literals cannot be kept correct, because `scripts/release.py` rewrites `package.json["version"]` at release time, so every server eventually misreports its version over the wire.

This replaces each literal with `SERVER_VERSION`, read from the package's own `package.json` at runtime.

| Server | Literal in source | `package.json` | |
|---|---|---|---|
| filesystem | `0.2.0` | 0.6.3 | drifted today |
| memory | `0.6.3` | 0.6.3 | matches now, drifts at next release |
| everything | `2.0.0` | 2.0.0 | matches now, drifts at next release |

`memory` and `everything` match at rest but drift the moment a release stamps a CalVer version into their manifests, so all three are fixed together. `sequentialthinking` is deliberately untouched here .. it belongs to #4576.

## Server Details

- Server: filesystem, memory, everything
- Changes to: reported `serverInfo.version` (no tools, resources, or prompts changed)

## Motivation and Context

Related to #4575.

`scripts/release.py:70-76` (`NpmPackage.update_version`) rewrites `package.json["version"]` to a date-based version generated by `gen_version()` at release time. A version committed in TypeScript source is therefore stale by construction .. it is never the version that ships. That makes this one systemic defect rather than a series of independent oversights, and it is why the fix has to read the manifest at runtime rather than sync a literal.

This stays correct if #4463 (semver via changesets) changes how versions are stamped, since it reads whatever ends up in the shipped manifest.

### Implementation

`version.ts` is duplicated at each package root rather than extracted into a shared package. A shared `src/shared/` workspace package would be auto-detected as publishable by both `find_changed_packages` (`release.py`) and the CI `detect-packages` matrix, would need its own npm trusted-publisher registration per `RELEASING.md`, and would make each server's release depend on a separately-versioned package. That is disproportionate for ~30 lines, so the duplication is deliberate.

The implementation matches #4576 deliberately, so all four servers end up with the same file. Two details worth noting for review:

- **The search is bounded.** The manifest is either alongside the module (running from source, e.g. under vitest) or one level up (running from `dist/`). Those two locations are checked and nothing else, it never walks up into the monorepo and cannot pick up a parent `package.json`.
- **`version.ts` must stay at the package root.** Moving it into a subdirectory breaks the `dist/` case. This matters for `everything`, where the surrounding code is organised into subfolders.

There is one intentional difference from #4576:

```diff
-    } catch {
-      // Try the next candidate when running from dist/ or source.
+    } catch (error) {
+      // Only a missing manifest is skippable; a corrupt or unreadable one is a real failure.
+      if ((error as NodeJS.ErrnoException)?.code !== 'MODULE_NOT_FOUND') {
+        throw error;
+      }
     }
```

A blanket `catch` swallows every failure, so a corrupt or permission-denied `package.json` reports `Could not locate package.json for server version` .. misleading for a file that was found and is unreadable for a different reason. Narrowing to `MODULE_NOT_FOUND` lets real failures surface. Happy to drop this if reviewers would rather the four files stay byte-identical, and equally happy to open it against #4576 instead.

## How Has This Been Tested?

- `npm run build` clean across all packages (`everything` type-checks its tests, and passes).
- Suites pass: everything 113, filesystem 158, memory 56. 18 of those are new (6 per package). `sequentialthinking` unchanged at 14.
- Ran a server in Claude Desktop against the local build: connects, negotiates, and its tools work normally.

Each package's `version.ts` gets six tests, in two groups:

- **Four drive the resolver directly**, with `createRequire` mocked: the happy path against the real manifest, throwing when neither candidate exists (asserting the search stops at the package root), propagation of non-`MODULE_NOT_FOUND` errors (`EACCES`), and propagation of a malformed manifest. The mock is a pass-through by default, so the happy-path assertion runs against the real filesystem in the same file; only tests that need a failure inject one.
- **Two exercise the real build** .. importing `dist/version.js`, and spawning `node dist/index.js` to read `serverInfo.version` back over a stdio handshake (`everything` 2.0.0, `filesystem` 0.6.3, `memory` 0.6.3; `filesystem` reported `0.2.0` before this change). These use `it.skipIf` so an unbuilt tree still passes, matching #4576. They do run in CI, since `npm ci` triggers `prepare` .. `npm run build`.

The two propagation tests and the bounded-search test were mutation-tested: reverting the catch to a blanket `catch {}` fails the first two, and adding a third candidate path beyond the package root fails the third.

## Breaking Changes

None. No client configuration changes. Servers now report their actual published version instead of a stale literal, which is the intent of the fix.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have updated the server's README accordingly — n/a, no README documents the server version
- [x] I have documented all environment variables and configuration options — n/a, none added

## Additional context

This overlaps existing work, and I'd rather flag that up front than have a reviewer find it:

- **#4576** fixes `sequentialthinking`. This PR is the follow-up @KarlLeen proposed there and stays out of its way.
- **#4557** fixes `memory` and `sequentialthinking` (alongside a zod dependency fix). If that lands first, I'll drop `memory` here and reduce this to `filesystem` and `everything`.
- **#4406**, the equivalent report for `memory`, was closed as not planned.

If maintainers would rather have one combined change, I'm happy for this to be folded into #4576 instead. Credit to @KarlLeen for reporting #4575 and to the authors of #4576 and #4557 .. the direction here is the same as theirs.

### DIFF
--- a/src/everything/__tests__/version.test.ts
+++ b/src/everything/__tests__/version.test.ts
@@ -1,0 +1,103 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { resolvePackageVersion } from '../version.js';
+
+vi.mock('node:fs', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:fs')>();
+  return { ...actual, readFileSync: vi.fn(actual.readFileSync) };
+});
+
+const actualFs = await vi.importActual<typeof import('node:fs')>('node:fs');
+const readFileSyncMock = vi.mocked(readFileSync);
+
+const packageRoot = dirname(dirname(fileURLToPath(import.meta.url)));
+const { version } = JSON.parse(actualFs.readFileSync(join(packageRoot, 'package.json'), 'utf8'));
+
+/** fs errors carry a `code`; only ENOENT/ENOTDIR mean "not at this path". */
+const fsError = (code: string) => Object.assign(new Error(code), { code });
+
+beforeEach(() => {
+  readFileSyncMock.mockReset();
+  readFileSyncMock.mockImplementation(actualFs.readFileSync);
+});
+
+describe('resolvePackageVersion', () => {
+  it('reports the version from package.json', () => {
+    expect(resolvePackageVersion()).toBe(version);
+  });
+
+  it('does not fall back while the manifest is readable', () => {
+    expect(resolvePackageVersion('unused-fallback')).toBe(version);
+  });
+
+  it('falls back when no manifest is found', () => {
+    readFileSyncMock.mockImplementation(() => {
+      throw fsError('ENOENT');
+    });
+
+    expect(resolvePackageVersion('1.2.3-fallback')).toBe('1.2.3-fallback');
+  });
+
+  it('falls back to 0.0.0-dev by default', () => {
+    readFileSyncMock.mockImplementation(() => {
+      throw fsError('ENOENT');
+    });
+
+    expect(resolvePackageVersion()).toBe('0.0.0-dev');
+  });
+
+  it('checks the parent directory, covering the dist/ layout', () => {
+    readFileSyncMock
+      .mockImplementationOnce(() => {
+        throw fsError('ENOENT');
+      })
+      .mockImplementationOnce(() => JSON.stringify({ version: '9.9.9' }));
+
+    expect(resolvePackageVersion()).toBe('9.9.9');
+  });
+
+  it('treats ENOTDIR as a missing manifest', () => {
+    readFileSyncMock
+      .mockImplementationOnce(() => {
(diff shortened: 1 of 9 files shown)


olaservo:
#4576 has landed for sequentialthinking, so this can be un-parked. Note #4407 covers memory with the same `version.ts` shape; one of the two should drop the memory part once the other merges.

olaservo:
#4407 has landed for memory, so this can drop `src/memory/` and cover filesystem and everything only. The `version.ts` shape there matches what you have.

PR #4610, opened 2026-08-04
Add mcpbeat to Resources

theevoq:
mcpbeat monitors the MCP ecosystem rather than listing it: every remote address in the registry receives a real JSON-RPC `initialize` handshake every 15 minutes, and the tool list of each responding server is recorded together with the date it was read.

Current numbers, measured today:

- 20 107 servers tracked, 10 570 of them reachable over the network
- 7 900 answering right now; 2 279 are listed as active in the registry but never respond
- 1.4M checks performed so far

Adoption comes from actual npm and PyPI download counts rather than GitHub stars, and Agent Skills are deduplicated by SHA-256 of their content, so a skill repackaged into forty repositories counts once.

Everything is published as CSV under CC BY 4.0 at https://mcpbeat.com/data/ — no key, no signup, so any number on the site can be verified against the raw rows.

### DIFF
--- a/ADDITIONAL.md
+++ b/ADDITIONAL.md
@@ -54,6 +54,7 @@ This is a curated collection of community-built frameworks and resources that si
 - **[Discord Server](https://glama.ai/mcp/discord)** – A community discord server dedicated to MCP by **[Frank Fiegel](https://github.com/punkpeye)**
 - **[Install This MCP](https://installthismcp.com)** - Reduce Installation Friction with beautiful installation guides
 - **[Klavis AI](https://www.klavis.ai)** - Open Source MCP Infra. Hosted MCP servers and MCP clients on Slack and Discord.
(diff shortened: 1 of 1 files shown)


Ecocitizenz:
This is valuable because it measures observed state rather than merely listed state.

The 15-minute initialize checks and tool snapshots expose an important distinction for the MCP ecosystem:

registry discovery ≠ current operational state.

A registry can establish that a server has been published and where it is expected to exist. Runtime observation can establish that something responded at a particular time. Neither, by itself, establishes the full authority/evidence chain a relying party may need: who operates it, which deployment/version is authoritative, what changed, and whether that state has subsequently been withdrawn or superseded.

I think keeping those layers explicit will become increasingly useful as MCP moves into enterprise use:

Discovery → Observed state → Operator/authority → Evidence → Freshness.

That is very close to the resolver-first model we are building with ECZ-ID: not replacing registries or monitoring, but making independently resolvable identity, authority and current evidence available alongside them.

— Adam Al-Zayed
Founder | EcoCitizenz Ltd
ECZ-ID — MCP & Agent Trust
https://mcp.ecocitizenz.com/
Resolver: https://resolver.ecocitizenz.org/passport/ECZ-GB-RBS1NW

PR #4611, opened 2026-08-04
fix(fetch): cap mcp dependency below 2.0

### LINKED ISSUE #4600: MCP client for fetch failed to start
⚠ MCP client for `fetch` failed to start: MCP startup failed: handshaking with
  MCP server failed: connection closed: initialize response

**resolve**

[mcp_servers.fetch]
type = "stdio"
command = "uvx"
args = ["--with", "mcp<2", "mcp-server-fetch"]

Root cause: mcp-server-fetch 2026.7.10 allows mcp>=1.1.3, so uvx selects incompatible mcp 2.0.0, where McpError was renamed to MCPError. Python 3.13 fails identically, so Python 3.14 is not responsible.

I verified the pinned command with a real MCP handshake: server=mcp-fetch version=1.29.0, tools=fetch. Apply the argument change in [config.toml](tauri://localhost/home/mrking/.codex/config.toml:95), then restart Codex; the Inspector’s “Web is up” message only confirms the Inspector UI started, not that the fetch server connected.

from codex

guptaishaan:
Fixes #4600

`mcp>=1.1.3` has no upper bound, so a fresh `uvx mcp-server-fetch` now resolves mcp 2.0.0, which renamed `McpError` to `MCPError` and dropped the low-level `Server` decorators. `server.py` dies on the import at line 6 before writing anything to stdout, which the client sees as `connection closed: initialize response`.

This caps the requirement at `mcp>=1.1.3,<2` and updates `uv.lock` to match. The lock already pinned mcp 1.28.1, so CI never saw this; only fresh resolves are affected.

I did not port fetch to mcp 2.0. It is not just the exception rename: `list_tools`, `list_prompts`, `call_tool`, and `get_prompt` are all gone from the low-level `Server`, and `server.py` uses all four. That is a migration, not a bug fix.

Verified on Linux, Python 3.12, uv 0.12.0, with a real JSON-RPC stdio handshake:

- published `mcp-server-fetch==2026.7.10`: no initialize response, `ImportError: cannot import name 'McpError'`
- wheel built from this branch: resolves `mcp==1.29.0`, `initialize ok: server=mcp-fetch version=1.29.0`, `tools=['fetch']`

Added `tests/test_dependencies.py`, which fails on the old specifier and passes on the new one. `ruff check` and `pyright` are clean. `tests/test_server.py::test_empty_content_returns_error` fails on my machine both with and without this patch, because there is no Node.js on PATH and readabilipy falls back to pure-Python mode. Unrelated to this change.

Not verified: I reproduced with my own stdio client, not with Codex, and the "after" run used a locally built wheel rather than a published one.

Two things for you to decide. First, `src/git` and `src/time` have the same unbounded range and break the same way under mcp 2.0.0 (`mcp-server-time` on the same `McpError` import, `mcp-server-git` on `'Server' object has no attribute 'list_tools'`). I left them out to keep this scoped to the reported issue, happy to add the same one-line cap to both. Second, if you would rather port to mcp 2.0 than cap, close this.

Thanks to @tking007 for the report, including the correct diagnosis and the `--with "mcp<2"` workaround.


olaservo:
Superseded by #4663, released as `2026.8.18`. Same cap, with the floor raised to 1.29.0. Closing — see the note on #4563.

PR #4617, opened 2026-08-06
Make memory graph writes atomic

### LINKED ISSUE #4614: saveGraph() in src/memory uses non-atomic fs.writeFile — risk of corrupted memory file on interruption

```ts
```

## Suggested fix

```ts
}
```

## Environment

joenermunch:
## Description

Memory graph updates are now written to a unique temporary file in the same directory. The completed file is renamed over the target only after the write succeeds. Failed writes remove the temporary file and leave the previous graph intact.

Fixes #4614.

## Server Details

- Server: memory
- Changes to: persistence

## Motivation and Context

Direct writes truncate the target before the replacement content is complete. An interrupted write can therefore corrupt the only persisted copy of the graph.

## How Has This Been Tested?

- Added a regression test that simulates a partial write failure and verifies the original graph remains unchanged.
- Verified temporary files are removed after failure.
- `npm test --workspace @modelcontextprotocol/server-memory -- --maxWorkers=4` (51 tests)
- `npm run build --workspace @modelcontextprotocol/server-memory`

## Breaking Changes

None.

## Types of changes

## Checklist

## Additional context

The temporary file is created beside the memory file so the rename stays on the same filesystem. Its randomized name also avoids collisions between concurrent processes.

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { promises as fs } from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
@@ -18,6 +18,8 @@ describe('KnowledgeGraphManager', () => {
   });
 
   afterEach(async () => {
+    vi.restoreAllMocks();
+
     // Clean up test file
     try {
       await fs.unlink(testFilePath);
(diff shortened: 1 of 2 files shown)


Selven81:
Thanks for this — I know it's closed, but I wanted to note that it was the
first of five PRs on #4614 and the first to state that the temporary file has
to live beside the target so the rename stays on one filesystem. That's the
constraint that matters most in practice: a temp file in `os.tmpdir()` fails
with `EXDEV` when the memory file is on a different mount, which is exactly the
container setup where I hit the original bug.

I credited it in the summary I posted on the issue:

PR #4618, opened 2026-08-06
fix(memory): write memory graph atomically to prevent corruption

### LINKED ISSUE #4614: saveGraph() in src/memory uses non-atomic fs.writeFile — risk of corrupted memory file on interruption

Arnab758:
## What

Fixes #4614.

`saveGraph()` wrote the memory file directly with `fs.writeFile`, so an interruption mid-write (process crash, kill, power loss) could leave a truncated or corrupted memory file — permanently losing the agent's long-term memory.

## How

Write the graph to a unique temp file, then `rename()` it into place. Rename is atomic on POSIX and Windows, so the memory file is always either the old complete version or the new complete version — never a half-written one.

- Random suffix (`randomBytes(16)`) prevents collisions when saves overlap
- Temp file is cleaned up on failure so failed saves never leak artifacts

## Tests

Added a regression test that simulates an interruption mid-write (partial bytes written, then error) and verifies:
- The original memory file is untouched
- No temp files leak

`tsc --noEmit` passes. Full suite: 51/51 tests pass.

## Notes

This complements (not duplicates) #4555, which serializes concurrent graph mutations. #4555 fixes concurrent writes racing each other; this fixes a single write being interrupted mid-write — two different failure modes.

Known limitation: a hard kill between write and rename can still leave a temp file behind (inherent to the temp+rename pattern). The memory file itself is never corrupted.

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { promises as fs } from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
@@ -18,6 +18,8 @@ describe('KnowledgeGraphManager', () => {
   });
 
   afterEach(async () => {
+    vi.restoreAllMocks();
+
     // Clean up test file
     try {
       await fs.unlink(testFilePath);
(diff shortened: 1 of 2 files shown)


Arnab758:
Hi maintainers — just bumping this gently. The change is small (atomic write for saveGraph, ~20 lines) and all 37 memory server tests pass locally, including a new regression test for interrupted writes. Would appreciate a review when anyone has a moment. 🙏

Selven81:
Thanks for picking this up. I read all five PRs on #4614 and merged the
strongest parts to run against my own setup — two came from here:

- `randomBytes(16)` for the temp filename. `saveGraph` is async and can be in
  flight more than once in a single process, so anything derived from `pid` +
  timestamp can collide. This doesn't.
- the note in the code that a hard kill between write and rename can still
  leave an orphan `.tmp` while the memory file itself stays intact. Being
  explicit about what a fix does *not* cover is worth as much as the fix.

You're also right that #4555 is complementary rather than competing — I kept
that framing in the summary.

Full write-up of what came from where:

olaservo:
Closing in favour of #4642, which has been merged for #4614. Thanks for the PR.

PR #4620, opened 2026-08-07
fix(time): expand LOCAL_TIMEZONE in Docker ENTRYPOINT

ks72:
Server: time
Change: Docker startup / LOCAL_TIMEZONE configuration
Type: Bug fix
Breaking change: None
Reproduction: documented Docker command currently exits 1
Verification: same command starts successfully after patch


ks72:
Friendly ping for triage. This fixes the documented `-e LOCAL_TIMEZONE` Docker usage for the time reference server, which currently fails on startup because the exec-form ENTRYPOINT passes `${LOCAL_TIMEZONE}` literally.

The change is limited to the Dockerfile, the documented command was verified before/after, and CI is green.

Happy to add a regression test or adjust the approach if maintainers prefer.

PR #4622, opened 2026-08-08
fix(everything): re-check GZIP_ALLOWED_DOMAINS on redirect hops

SashaMIT:
## Description

`gzip-file-as-resource` validated `GZIP_ALLOWED_DOMAINS` only against the initial URL, then called `fetch` with the default redirect policy (`follow`). An allowlisted host could respond with a `3xx` to a host outside the allowlist, and the tool would still download and gzip that body.

This change follows redirects manually (`redirect: "manual"`), re-applies the same protocol and domain checks on every hop, and caps redirect depth.

## Server Details
- Server: everything
- Changes to: tools (`gzip-file-as-resource`)

## Motivation and Context

`GZIP_ALLOWED_DOMAINS` is documented as the domain allowlist for this tool. Without hop re-checks, the control does not hold whenever an allowed origin redirects.

Local confirmation (pre-fix): with allowlist `127.0.0.1`, an origin on `127.0.0.1` that `302`s to `http://[::1]:.../secret` returned the IPv6 target body under default `fetch` redirect following. Post-fix, the same setup rejects `[::1]` before the second request.

## How Has This Been Tested?
- Added vitest coverage that mocks a `302` from an allowed host to `127.0.0.1` and asserts rejection + that the redirect target is never fetched
- `npm test` in `src/everything` (108 tests) green
- Dual-stack local HTTP repro green post-fix

## Breaking Changes
None for the default (empty allowlist). Operators with `GZIP_ALLOWED_DOMAINS` set get the documented allowlist behavior on redirect chains as well as on the first URL.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

Framed as a product bug in the existing allowlist control (reference `everything` server). `SECURITY.md` notes these servers are reference implementations and not a security-reporting surface; happy to adjust if maintainers prefer a different lane.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/everything/__tests__/tools.test.ts
+++ b/src/everything/__tests__/tools.test.ts
@@ -1217,5 +1217,63 @@ describe('Tools', () => {
         handler!({ name: 'test.gz', data: 'ftp://example.com/file.txt', outputType: 'resource' })
       ).rejects.toThrow('Unsupported URL protocol');
     });
+
+    it('should re-check GZIP_ALLOWED_DOMAINS on redirect targets', async () => {
+      vi.stubEnv('GZIP_ALLOWED_DOMAINS', 'allowed.example');
+      vi.resetModules();
+
+      const fetchMock = vi.fn(async (input: RequestInfo | URL, init?: RequestInit) => {
+        const href = String(input);
+        const redirect = init?.redirect ?? 'follow';
+        if (href === 'https://allowed.example/file') {
+          if (redirect === 'follow') {
+            // Pre-fix behavior would auto-follow into a non-allowed host.
+            return new Response('SECRET_FROM_PRIVATE_HOST', { status: 200 });
+          }
+          return new Response(null, {
+            status: 302,
+            headers: { Location: 'http://127.0.0.1/secret' },
+          });
+        }
+        if (href === 'http://127.0.0.1/secret') {
+          return new Response('SECRET_FROM_PRIVATE_HOST', { status: 200 });
+        }
+        throw new Error(`unexpected fetch: ${href}`);
+      });
+      vi.stubGlobal('fetch', fetchMock);
+
+      const { registerGZipFileAsResourceTool: registerFresh } = await import(
+        '../tools/gzip-file-as-resource.js'
+      );
+
+      const mockServer = {
+        registerTool: vi.fn(),
+        registerResource: vi.fn(),
+      } as unknown as McpServer;
+
+      let handler: Function | null = null;
+      (mockServer.registerTool as any).mockImplementation(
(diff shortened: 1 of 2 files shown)


Ecocitizenz:
The per-hop revalidation is the important property here. An allowlist checked only against the initial URL protects the string supplied by the caller, not the destination the tool ultimately reaches.

One additional test worth considering is a redirect chain that changes the address representation without appearing to change the logical destination, for example:

* hostname → IPv4 literal;
* IPv4 → IPv6;
* public hostname resolving to a private or loopback address;
* repeated redirects across permitted and non-permitted hosts;
* relative `Location` followed by an absolute cross-origin redirect;
* userinfo or unusual URL canonicalisation around the authority component.

The invariant should be evaluated on every hop after URL resolution and before the next network request:

`resolved protocol + canonical destination + applicable network policy → permit or reject`

If DNS resolution can change between validation and connection, production implementations may also need an egress/network-layer control rather than treating URL validation as complete SSRF containment.

The current patch closes the demonstrated redirect-following gap cleanly and the “target is never fetched” assertion is especially important.

SashaMIT:
Great callouts, thank you. Added coverage in 9791fdb for the representation-change cases: hostname to IPv4 and IPv6 literals, permitted-to-non-permitted chains, relative Location then absolute cross-origin, and the userinfo authority trick. Each scenario asserts the non-allowed destination is never fetched, and there is a positive control that a relative redirect staying on the allowed host still succeeds.

On DNS resolution changing between validation and connection: agreed, that is egress-layer territory. URL-layer validation alone is not complete SSRF containment there; I have noted it as out of scope in the PR description.

Ecocitizenz:
Thanks — that is a strong set of regression cases, especially the representation-change and userinfo authority cases.

I agree with keeping the DNS-resolution race explicitly out of scope for the URL layer. 

That boundary is important:
URL validation can establish that the requested authority is permitted at decision time; egress enforcement is what determines where the connection is actually allowed to terminate.

Keeping those responsibilities separate also makes the resulting security evidence much clearer. A relying party can distinguish:

what the application validated,
what the network layer actually permitted,
and what decision/result was observed at runtime.

That separation becomes useful when MCP deployments need to produce reviewable evidence later rather than just enforce a rule in-process.

Nice improvement to the PR.

— Adam Al-Zayed
Founder | EcoCitizenz Ltd
ECZ-ID — MCP & Agent Trust
https://mcp.ecocitizenz.com/⁠�

Resolver: https://resolver.ecocitizenz.org/passport/ECZ-GB-RBS1NW⁠�

PR #4630, opened 2026-08-11
fix(filesystem): make move_file fail instead of silently overwriting the destination

### LINKED ISSUE #4628: server-filesystem: move_file silently overwrites an existing destination (data loss)
**Version:** `2026.7.10`
**Node:** `v25.2.1`

## Summary

`move_file` overwrites an existing destination file without warning and reports success. The README states the opposite:

> - **move_file**
>   * Move or rename files and directories
>   * Inputs: `source` (string), `destination` (string)
>   * **Fails if destination exists**

The tool description returned over the wire says the same thing:

> If the destination exists, the operation will fail.

Because no existence check is performed, the destination file's contents are destroyed and unrecoverable. This is a data-loss bug, not just a documentation error: an agent that reads "fails if destination exists" will reasonably treat `move_file` as a safe, non-clobbering operation and use it without first checking whether the target is occupied.

The impact is amplified by the fact that this server ships **no delete tool**. Users (and agents) reasonably infer that the toolset cannot destroy existing files, only create and modify them. `move_file` quietly provides arbitrary file deletion: move anything onto a target and the target is gone.

That inference is actively encouraged by the README, which is itself inconsistent here. The Features section advertises:

> Read/write files · Create/list/**delete** directories · Move files/directories

No delete tool exists in the API section or in the tool list returned at runtime. So a reader is told deletion is available, cannot find it, concludes the toolset is non-destructive by design — and then loses data to `move_file`.

## Reproduction

Two files with **different names**, **different contents**, and **different sizes**, so the result is unambiguous.

1. `write_file` -> `<allowed_dir>/_movetest/alpha_source.txt`

   ```
   FILE ALPHA
   role: SOURCE
   unique marker: AAAA-1111-SOURCE
   length: this file is deliberately short.
   ```

2. `write_file` -> `<allowed_dir>/_movetest/beta_destination.txt`

   ```
   FILE BETA
   role: DESTINATION
   unique marker: BBBB-2222-DESTINATION
   ...(padding so the sizes differ)...
   ```

3. `list_directory_with_sizes` on `_movetest`:

   ```
   [FILE] alpha_source.txt                     97 B
   [FILE] beta_destination.txt                307 B
   Total: 2 files, 0 directories
   ```

4. `move_file` with `source = alpha_source.txt`, `destination = beta_destination.txt`

**Expected:** error — destination exists.

**Actual:** success, no warning:

```
Successfully moved <...>/_movetest/alpha_source.txt to <...>/_movetest/beta_destination.txt
```

5. `list_directory_with_sizes` on `_movetest`:

   ```
   [FILE] beta_destination.txt                 97 B
   Total: 1 files, 0 directories
   ```

6. `read_text_file` on `beta_destination.txt` returns ALPHA's contents. The marker `BBBB-2222-DESTINATION` is gone; 307 bytes were destroyed by an operation documented to refuse.

## Cause

Presumably a bare `fs.rename` with no prior existence check — which clobbers on every platform Node supports.

POSIX `rename(2)` replaces an existing destination by definition. On Windows, libuv's `fs__rename` calls `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING` (added in joyent/libuv#283 precisely to match Linux behaviour), and per the Win32 docs that flag replaces the contents of the destination file. The libuv docs accordingly describe `uv_fs_rename` as equivalent to `rename(2)`.

So the overwrite is the platform default in both cases, and the README documents an intent that was never implemented in code.

## Note on tool annotations

`move_file` currently carries `destructiveHint: true`, but the README's own rationale for it is *"Deletes source file"* — i.e. the annotation is justified by the loss of the **source**, and the destination is still assumed safe. See #2988, which proposed `destructiveHint: false` on the explicit grounds that *"a repeat typically errors because source no longer exists and README says 'fails if destination exists'"*.

So the reasoning around this tool's safety has been anchored on a README claim the code does not honour. Worth revisiting the note and rationale even though the flag itself happens to be correct.

## Suggested fix

Either:

**(a) Implement the documented behaviour** — `stat` the destination first and reject if it exists:

```ts
try {
  await fs.stat(validDestPath);
  throw new Error(`Destination already exists: ${destination}`);
} catch (e) {
  if (e.code !== "ENOENT") throw e;
}
await fs.rename(validSourcePath, validDestPath);
```

Note this is TOCTOU-racy but adequate for this server's threat model; a stricter option on Linux is `renameat2` with `RENAME_NOREPLACE`.

**(b) Keep the overwrite behaviour** but fix the README and the tool description to say so plainly, mirroring `write_file`'s existing wording ("exercise caution with this"), and add an explicit `overwrite: boolean` input defaulting to `false`.

(a) is preferable: it matches every published description of the tool, and clients that gate on `destructiveHint` today do so believing the destination is protected.

eeshsaxena:
Fixes #4628.

`move_file`'s description and the README both say the move fails if the destination already exists:

> If the destination exists, the operation will fail.

But the handler calls `fs.rename` directly, which silently overwrites the destination and still reports success. Because this server ships no delete tool, an agent that reads the contract reasonably treats `move_file` as non-clobbering, so this quietly turns into unadvertised, irreversible file deletion: move any file onto an existing target and the target's contents are gone.

## Fix

Added a `moveFile` helper in `lib.ts` that checks the destination first and rejects if anything already occupies it, then renames only when the target is free:

```ts
try {
  await fs.lstat(destinationPath);
} catch (error) {
  if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
    await fs.rename(sourcePath, destinationPath);
    return;
  }
  throw error;
}
throw new Error(`Destination already exists: ${destinationPath}`);
```

`lstat` (not `stat`) is used so an existing symlink at the destination is detected rather than followed. The `move_file` handler now calls this helper. Behaviour now matches the documented contract, and the successful-move path is unchanged.

## Tests

Added two cases to `__tests__/lib.test.ts`:
- moves when the destination doesn't exist (rename is called)
- rejects with "Destination already exists" and never calls rename when the destination is present

The second test fails against the old `fs.rename`-only behaviour and passes with the fix. Full file suite green:

```
cd src/filesystem && npm run build && npx vitest run __tests__/lib.test.ts
# 48 passed
```

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -14,6 +14,7 @@ import {
   getFileStats,
   readFileContent,
   writeFileContent,
+  moveFile,
   // Search & filtering functions
   searchFilesWithValidation,
   // File editing functions
@@ -307,6 +308,29 @@ describe('Lib Functions', () => {
         await writeFileContent('/test/file.txt', 'new content');
         
         expect(mockFs.writeFile).toHaveBeenCalledWith('/test/file.txt', 'new content', { encoding: "utf-8", flag: 'wx' });
+      });
+    });
+
+    describe('moveFile', () => {
+      it('moves the file when the destination does not exist', async () => {
+        const enoent = Object.assign(new Error('not found'), { code: 'ENOENT' });
+        mockFs.lstat.mockRejectedValueOnce(enoent);
+        mockFs.rename.mockResolvedValueOnce(undefined);
+
+        await moveFile('/test/source.txt', '/test/dest.txt');
+
+        expect(mockFs.rename).toHaveBeenCalledWith('/test/source.txt', '/test/dest.txt');
+      });
+
+      it('fails without overwriting when the destination already exists', async () => {
+        // lstat resolving means the destination is occupied.
+        mockFs.lstat.mockResolvedValueOnce({} as any);
+
+        await expect(moveFile('/test/source.txt', '/test/dest.txt')).rejects.toThrow(
+          'Destination already exists'
+        );
+
+        expect(mockFs.rename).not.toHaveBeenCalled();
       });
     });
 
--- a/src/filesystem/index.ts
+++ b/src/filesystem/index.ts
@@ -21,6 +21,7 @@ import {
   getFileStats,
   readFileContent,
   writeFileContent,
+  moveFile,
   searchFilesWithValidation,
   applyFileEdits,
   tailFile,
@@ -631,7 +632,7 @@ server.registerTool(
   async (args: z.infer<typeof MoveFileArgsSchema>) => {
     const validSourcePath = await validatePath(args.source);
     const validDestPath = await validatePath(args.destination);
-    await fs.rename(validSourcePath, validDestPath);
+    await moveFile(validSourcePath, validDestPath);
     const text = `Successfully moved ${args.source} to ${args.destination}`;
     const contentBlock = { type: "text" as const, text };
     return {
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -185,6 +185,25 @@ export async function writeFileContent(filePath: string, content: string): Promi
 }
 
 
+export async function moveFile(sourcePath: string, destinationPath: string): Promise<void> {
+  // The move_file tool contract (and README) state the operation fails if the
+  // destination already exists. fs.rename would silently overwrite it, which is
+  // a data-loss bug, so reject up front when anything - file, directory, or
+  // symlink - occupies the target. lstat is used so an existing symlink at the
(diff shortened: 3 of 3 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`fix(filesystem): make move_file fail instead of silently overwriting the destination`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining servers!

olaservo (approved):
Smallest correct fix of the five open PRs for #4628: `lstat` catches a file, directory (empty or not) or symlink at the destination, and `rename` stays atomic. Tests follow the existing mocked `lib.test.ts` style.

PR #3238, opened 2026-01-21
fix (filesystem): Add UTF-8 encoding and Unicode normalization for umlauts and symlinks

kernhuber:
## Description
I found that working with directories/filesystems on macOS did not work when path names contained umlauts (ä, ö, ü, ß). The same was true when using symbolic links which involved such characters. 

The root cause was threefold:
1. Missing UTF-8 encoding initialization for Windows STDIO communication
2. Lack of Unicode normalization (NFC vs NFD forms causing string comparison failures)
3. Inconsistent handling of symlinks pointing to or containing paths with umlauts

This PR fixes these issues by:
- Adding UTF-8 encoding initialization for Windows STDIO
- Implementing Unicode NFC normalization in path-utils
- Improving symlink handling with proper encoding support
- Adding better error messages for encoding-related issues

The fix benefits all languages with diacritical marks (German, French, Spanish, Nordic, Eastern European, etc.).

## Publishing Your Server
**Note: We are no longer accepting PRs to add servers to the README.** Instead, please publish your server to the [MCP Server Registry](https://github.com/modelcontextprotocol/registry) to make it discoverable to the MCP ecosystem.

To publish your server, follow the [quickstart guide](https://github.com/modelcontextprotocol/registry/blob/main/docs/modelcontextprotocol-io/quickstart.mdx). You can browse published servers at [https://registry.modelcontextprotocol.io/](https://registry.modelcontextprotocol.io/).

## Server Details

- Server: **filesystem**
- Changes to: **Core path handling utilities, validation logic, and initialization**
  - `index.ts`: UTF-8 encoding initialization
  - `path-utils.ts`: Unicode NFC normalization
  - `lib.ts`: Enhanced symlink resolution and error handling
  - `roots-utils.ts`: Root URI parsing improvements

### Problem
Users with non-ASCII characters in their file paths (common in German, French, Spanish, and many other languages) experienced complete filesystem access failures. Specifically:

1. **Windows users**: Umlauts displayed as `?` due to STDIO encoding mismatch
2. **macOS/Linux users**: Path validation failed because macOS stores filenames in NFD (decomposed Unicode: `ü` = `u` + combining diaeresis) while the server compared against NFC (composed: `ü` = single character)
3. **All platforms**: Symlinks pointing to or containing umlauts failed due to `fs.realpath()` returning inconsistent Unicode forms

### Real-world Impact
This made the filesystem server completely unusable for:
- Users with non-English names in their home directory path
- Projects with internationalized directory names
[...]

PR #4638, opened 2026-08-12
fix(filesystem): resolve Unicode-equivalent paths

### LINKED ISSUE #4633: move_file fails with non-ASCII characters in file path (e.g. French accents, œ, typographic apostrophe)
**Server:** @modelcontextprotocol/server-filesystem  
**OS:** Linux (Ubuntu)

**Description:**  
`move_file` fails with ENOENT when the source path contains non-ASCII characters 
such as French accented letters (é, è, à), ligatures (œ), or typographic 
apostrophes (').

Other tools like `read_text_file`, `list_directory`, and `search_files` handle 
these characters correctly.

**Steps to reproduce:**  
1. Have a file named e.g. `C1.1 - Mettre en œuvre une démarche.md`
2. Call `move_file` with that path as source
3. → ENOENT: no such file or directory

**Expected:** file is moved/renamed  
**Actual:** ENOENT error

KXHXK:
﻿## Description

Resolve filesystem paths whose requested spelling and on-disk entry differ only by Unicode normalization (for example NFC `?` versus NFD `e + combining acute`).

The normal lookup remains the fast path. After `fs.realpath()` returns `ENOENT`, validation now walks from the most specific allowed directory and:
- prefers exact component matches;
- falls back to canonically equivalent NFC matches;
- rejects ambiguous equivalent entries;
- resolves and revalidates every matched component against the existing allowed-directory boundary;
- preserves a missing final basename so create and move destinations continue to work.

## Server Details
- Server: filesystem
- Changes to: shared path validation used by filesystem tools

## Motivation and Context

Fixes #4633.

This differs from #3238: normalizing strings used by the security comparison does not resolve an OS lookup when the actual directory entry uses another normalization form. This change locates the real entry and then retains the existing realpath-based security checks.

## How Has This Been Tested?

- `npx vitest run src/filesystem/__tests__/unicode-paths.test.ts src/filesystem/__tests__/lib.test.ts --config src/filesystem/vitest.config.ts` (49 passed)
- `npm run build --workspace @modelcontextprotocol/server-filesystem`
- Full filesystem suite: 154 passed, 1 pre-existing Windows-only failure in `path-validation.test.ts > handles root directory as allowed` (`D:\other` with allowed `['/']`).

Added real-filesystem regression coverage for:
- composed request resolving a decomposed directory and filename;
- a new destination basename under a Unicode-equivalent parent;
- rejection of ambiguous canonically equivalent entries.

## Breaking Changes

None.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

The path containment implementation itself is unchanged. The fallback only runs after `ENOENT`, starts from an already allowed root, and re-checks each resolved component to prevent a normalization fallback from bypassing symlink boundaries.

### DIFF
--- a/src/filesystem/__tests__/unicode-paths.test.ts
+++ b/src/filesystem/__tests__/unicode-paths.test.ts
@@ -1,0 +1,49 @@
+import fs from 'fs/promises';
+import os from 'os';
+import path from 'path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { setAllowedDirectories, validatePath } from '../lib.js';
+
+describe('Unicode-equivalent filesystem paths', () => {
+  let testDirectory: string;
+
+  beforeEach(async () => {
+    testDirectory = await fs.mkdtemp(path.join(os.tmpdir(), 'mcp-unicode-paths-'));
+    setAllowedDirectories([testDirectory]);
+  });
+
+  afterEach(async () => {
+    setAllowedDirectories([]);
+    await fs.rm(testDirectory, { recursive: true, force: true });
+  });
+
+  it('resolves an existing decomposed path from a composed request', async () => {
+    const onDiskDirectory = 'de\u0301marche';
+    const onDiskFile = 're\u0301sume\u0301.txt';
+    await fs.mkdir(path.join(testDirectory, onDiskDirectory));
+    await fs.writeFile(path.join(testDirectory, onDiskDirectory, onDiskFile), 'content');
+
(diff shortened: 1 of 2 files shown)


olaservo (approved):
Reviewed the `validatePath` change carefully since it is the security boundary: every resolved component is realpath'd and re-checked against the allowed directories, ambiguous NFC/NFD matches are rejected, and a missing intermediate still fails. Suite passes locally on Windows.

One note for later: the ambiguity test creates both `café` and `cafe\u0301` in one directory. On macOS (HFS+ normalises, APFS is normalisation-insensitive) that second `mkdir` may fail with EEXIST, so the test could fail on a Mac dev machine even though CI (Linux) passes.

PR #4642, opened 2026-08-15
fix(memory): write the knowledge graph atomically

### LINKED ISSUE #4614: saveGraph() in src/memory uses non-atomic fs.writeFile — risk of corrupted memory file on interruption

teddiesloco:
## Description

`saveGraph()` in `src/memory/index.ts` wrote the knowledge graph directly to the memory file:

```ts
```

`fs.writeFile` opens the target with `'w'`, which truncates it **before** any new bytes are written. If the process is interrupted between truncation and completion — SIGKILL, container or host stop, OOM kill, power loss — the memory file is left empty or half-written.

Since that file is the only persistence layer for the graph, the window is small but the loss is total and unrecoverable.

This PR writes to a temporary file in the same directory and then renames it over the target. `rename(2)` is atomic on POSIX filesystems: a reader sees either the complete old file or the complete new one, never an intermediate state.

Two implementation details worth calling out:

- The temp file is created **in the same directory** as the target, so the rename stays on one filesystem. A temp file in `os.tmpdir()` would fail with `EXDEV` whenever the memory file lives on a different mount — a common setup when the memory file is on a mounted volume.
- On failure the temp file is unlinked, so repeated failures cannot accumulate strays next to the memory file.

Fixes #4614

## Server Details

- Server: `memory`
- Changes to: internal persistence only — no tool, resource, or prompt signatures change

## Motivation and Context

The memory server exists to accumulate knowledge across sessions, so durability is the one property users depend on most. The reporter of #4614 hit this with a backup script that stops a Distrobox/Podman container: if the server happens to be mid-write when the container stops, the accumulated graph can be lost.

I reproduced the underlying filesystem behaviour directly to confirm the mechanism rather than assuming it:

```
before : "ENTITY-A\nENTITY-B\nENTITY-C\n"
after   : "PARTIAL"
```

Opening the live file with `'w'` and dying before the write completes destroys the committed data.

## How Has This Been Tested?

Added `src/memory/__tests__/atomic-save.test.ts` with five tests:

1. the live memory file is never opened for truncating writes
2. a committed graph survives a write that fails midway
3. no temporary files remain after a successful write
4. the temporary file is cleaned up when the rename fails
5. graph contents still round-trip correctly across reloads

Tests 1 and 4 **fail against the previous implementation** and pass after the fix, so they pin the regression rather than merely describing it.

```
Test Files  4 passed (4)
     Tests  55 passed (55)      # 50 existing + 5 new
```

`npx tsc --noEmit` is clean.

I also verified the built output end-to-end: created entities, forced `rename` to throw to simulate a crash at commit time, and confirmed the previously committed entities were still intact with no temp files left behind.

## Breaking Changes

None. No configuration or client changes are required, and the on-disk format is unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have updated the server's README accordingly — not applicable, no user-facing behaviour or configuration changes
- [ ] I have documented all environment variables and configuration options — not applicable, no new options

## Additional context

I kept the change scoped to `saveGraph()`. Two adjacent hardening steps were deliberately left out to keep this reviewable and focused:

- **fsync before rename.** `rename` is atomic with respect to ordering, but without an `fsync` on the temp file the new contents may still be in the page cache during a power loss. Adding it costs a sync on every graph mutation, which is a real write-throughput tradeoff — worth a separate discussion.
- **Concurrent writer serialisation.** The temp filename includes the pid and a timestamp so two processes cannot collide on the same temp path, but concurrent `saveGraph()` calls still race on last-writer-wins. That is pre-existing behaviour and orthogonal to this fix.

Happy to fold either in if maintainers prefer.

### DIFF
--- a/src/memory/__tests__/atomic-save.test.ts
+++ b/src/memory/__tests__/atomic-save.test.ts
@@ -1,0 +1,133 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { promises as fs } from 'fs';
+import path from 'path';
+import os from 'os';
+import { KnowledgeGraphManager, Entity } from '../index.js';
+
+/**
+ * Regression tests for durable persistence of the knowledge graph.
+ *
+ * saveGraph() previously called fs.writeFile() directly on the memory file.
+ * fs.writeFile opens the target with 'w', which truncates it before any new
+ * bytes are written. If the process dies between truncation and completion
+ * (SIGKILL, container stop, OOM kill, power loss) the memory file — the sole
+ * persistence layer for the graph — is left empty or half-written, and the
+ * accumulated memory is unrecoverable.
+ *
+ * The fix writes to a temporary file in the same directory and then renames
+ * it over the target. rename(2) is atomic on POSIX filesystems: a reader
+ * either sees the complete old file or the complete new one, never a
+ * truncated intermediate state.
+ */
+describe('KnowledgeGraphManager persistence durability', () => {
+  let testDir: string;
+  let testFilePath: string;
+
+  beforeEach(async () => {
+    testDir = await fs.mkdtemp(path.join(os.tmpdir(), 'mcp-memory-atomic-'));
+    testFilePath = path.join(testDir, 'memory.jsonl');
+  });
+
+  afterEach(async () => {
+    vi.restoreAllMocks();
+    await fs.rm(testDir, { recursive: true, force: true });
+  });
+
+  it('never truncates the live memory file in place', async () => {
+    const manager = new KnowledgeGraphManager(testFilePath);
+    await manager.createEntities([
+      { name: 'Alice', entityType: 'person', observations: ['works at Acme Corp'] },
+    ]);
+
+    // Any write that targets the live file directly is destructive: it
+    // truncates committed data before the replacement is durable.
+    const writeFileSpy = vi.spyOn(fs, 'writeFile');
+    const openSpy = vi.spyOn(fs, 'open');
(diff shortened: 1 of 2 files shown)


Selven81:
Thanks for this — of the five PRs on #4614 it has the best test coverage, and
it's the only one that documents what it deliberately leaves out and why. I
took two things from it:

- `path.dirname` / `path.join` used explicitly, which makes the same-directory
  constraint visible in the code instead of implied by string concatenation;
- the test asserting the live file is **never opened for writing**, spying on
  both `writeFile` and `open`. That's stronger than checking the outcome,
  because it holds regardless of how the save is implemented — it fails against
  the current code, and would catch a future regression even if the resulting
  file happened to look right.

One thing I didn't carry over, in case it's useful: the temp filename derived
from `process.pid` + `Date.now()`. Since `saveGraph` is async and can run
concurrently within one process, two saves entering the same millisecond derive
the same path. I used `randomBytes` from #4618 instead.

Details in the summary on the issue:

teddiesloco:
Updated the temporary filename generation to use `randomBytes(16)` to eliminate any collision risk with concurrent async saves. Kept everything else intact alongside the test suite. Thanks @Selven81 for the thorough review and comparison!

olaservo (approved):
Temp file beside the target (no EXDEV), random name, cleanup on failure, and tests that fail on `main` for the right reasons. fsync is a separate durability concern and can be its own issue.

PR #4652, opened 2026-08-17
fix(sequentialthinking): restore nextThoughtNeeded in the advertised inputSchema required array

### LINKED ISSUE #4651: sequential-thinking: nextThoughtNeeded missing from inputSchema.required but required at runtime (regression from #3533)
### Summary

### Reproduction

```json
```

```json
```

```

```

### Root cause

```js
const schema = z.object({
  req: z.boolean(),
  pre: coercedBoolean,
  coerced: z.coerce.number(),
  opt: z.boolean().optional(),
});
```

Before #3533:

```diff
-      nextThoughtNeeded: z.boolean().describe("Whether another thought step is needed"),
+      nextThoughtNeeded: coercedBoolean.describe("Whether another thought step is needed"),
```

### Impact

### Suggested fix

```js
  z.boolean(),
]);
```

### How this was found

AmirF194:
## Root cause

Commit `1cdf806d` (#3533) wrapped `nextThoughtNeeded` in a `z.preprocess()`-based `coercedBoolean` to fix a real footgun (the string `"false"` coercing to `true`). `zod`'s `toJSONSchema(..., { io: "input" })` treats a `z.preprocess()`-wrapped field's input type as `unknown`, so it silently drops that field from the emitted `required` array, even though the field carries no `.optional()`. The advertised schema then says `nextThoughtNeeded` is optional while the runtime validator still requires it, so any client that builds its call arguments from the declared schema gets `-32602 Invalid params`.

## Invariant

The advertised `inputSchema.required` array and the runtime validator agree on which fields are mandatory for `sequentialthinking`.

## Fix

Rebuild `coercedBoolean` as a `.transform()` on an explicit `z.union([z.boolean(), z.string()])` instead of `z.preprocess()`. The union gives `toJSONSchema` a concrete input type to report, so `nextThoughtNeeded` stays in `required`, while parse behavior (including the case-insensitive string coercion #3533 added) is unchanged.

## Verification

- Docker (`node:22-slim`), clean checkout of `76d64c8` (current `main`): a live stdio MCP session (`initialize` → `tools/list` → `tools/call`) built its call arguments strictly from the advertised `required` array. **Fails on main** with `MCP error -32602: ... expected boolean, received undefined at nextThoughtNeeded`; **passes on this branch**, and `tools/list` now reports `required: ["thought","nextThoughtNeeded","thoughtNumber","totalThoughts"]`.
- Same session, sent `nextThoughtNeeded: "FALSE"` (the case #3533 fixed): still coerces to `false` on this branch, so the original footgun stays fixed.
- `npm run build` (tsc) and `npm test` (vitest, 14/14) both green on this branch.
- What I did **not** verify: this repo's test suite only covers `lib.ts` (`SequentialThinkingServer`), not `index.ts` (0% coverage before and after this change, per `npm test`'s own coverage report), because `index.ts` calls `runServer()` unconditionally at module load, so it isn't imported by the existing tests. I verified the fix with a live protocol session instead of adding a unit test, rather than invent a new import-safe seam (exporting the schema, guarding the `runServer()` call) that has no precedent anywhere else in this repo's ~30 other reference servers. Happy to add that seam if you'd like a permanent regression test here.

Fixes #4651


AmirF194:
modelcontextprotocol/servers#4655 (AbhiPra24, opened 71 minutes after this one) fixes the same nextThoughtNeeded/required drift with the same mechanism (drop z.preprocess for a union that keeps a real Zod type, so toJSONSchema stops dropping the field), plus a broader scope (also fixes #4575's version drift) and a full integration test suite this PR doesn't have. One behavioral difference worth flagging either way: this PR rejects a malformed nextThoughtNeeded string (anything besides "true"/"false") with a validation error, #4655's fallback silently coerces it to false. Happy to stand down in favor of #4655 if that's the one you'd rather take forward.

Update: modelcontextprotocol/servers#4669 (re2zero, opened two days after this one) landed the identical fix, same file, same mechanism, no test suite and no version-drift fix, so it does not change the comparison above.

Second update: modelcontextprotocol/servers#4695 (AbhiPra24, opened 2026-08-25) is that author's third attempt, now on the same z.union mechanism as this PR and with its own unit test suite, matching this PR's malformed-string rejection behavior rather than #4655's silent coercion. Still no version-drift fix. Whichever of #4655, #4695 or this PR you'd rather take forward is fine by us.


Third update: modelcontextprotocol/servers#4701 (Golchic, opened 2026-08-27) is a fourth independent fix, same z.union mechanism and the same malformed-string rejection as this PR, with its own unit and JSON-schema tests. Four PRs now converge on the same fix; still whichever you'd rather take forward is fine by us.

AmirF194:
Just checking in on this one, no rush. Happy to rebase or adjust scope if that would help it move.

olaservo:
Merged `main` and pushed a regression test to this branch: `tools/list` must advertise `nextThoughtNeeded` in `required`, `"True"`/`"FALSE"` strings are accepted, and `"yes"`/`""`/`"1"` are rejected. It runs against the built server so it checks what the SDK emits. Will merge once CI is green.

olaservo (approved):
Union+transform with an explicit issue for junk keeps the pre-#3533 semantics and puts `nextThoughtNeeded` back in `required`. Regression test added. CI green.

AmirF194:
Thanks for taking this the rest of the way and adding the regression test, appreciate it.

PR #4654, opened 2026-08-17
fix(filesystem): support recursive create_directory and prevent move_file overwrites

### LINKED ISSUE #4628: server-filesystem: move_file silently overwrites an existing destination (data loss)

### LINKED ISSUE #4629: server-filesystem: create_directory does not create parent directories

AbhiPra24:
## Summary

Fixes #4629 and #4628.

### 1. `create_directory` parent creation (Fixes #4629)
- **Problem**: When `create_directory` is called on a path with multiple non-existent parent directories (e.g. `<allowed_dir>/a/b/c`), `validatePath()` failed on `path.dirname()` throwing `"Parent directory does not exist: <allowed_dir>/a/b"`, preventing recursive directory creation even though `create_directory` passes `{ recursive: true }` to `fs.mkdir`.
- **Fix**: In `validatePath()`, when a path does not exist (`ENOENT`), walk up ancestor directories until an existing ancestor is found, resolving it via `fs.realpath` and verifying it stays within `allowedDirectories`.

### 2. `move_file` non-overwriting behavior (Fixes #4628)
- **Problem**: `move_file` called `fs.rename` directly, silently overwriting existing destination files and causing data loss, violating the documented contract ("If the destination exists, the operation will fail").
- **Fix**: Added pre-rename check with `await fs.lstat(validDestPath)` to throw `Destination already exists: <path>` when target is occupied.

## Verification
- Added regression tests in `src/filesystem/__tests__/lib.test.ts` and `src/filesystem/__tests__/structured-content.test.ts`.
- `npm run build` and `npm test` in `src/filesystem` passed (157/157 tests green).

---

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -188,18 +188,49 @@ describe('Lib Functions', () => {
         expect(result).toBe(path.resolve(newFilePath));
       });
 
-      it('rejects when parent directory does not exist', async () => {
+      it('handles non-existent paths with multiple non-existent parent directories by walking up to existing ancestor', async () => {
+        const nestedPath = process.platform === 'win32' ? 'C:\\Users\\test\\a\\b\\c' : '/home/user/a/b/c';
+        const ancestorPath = process.platform === 'win32' ? 'C:\\Users\\test' : '/home/user';
+        
+        const enoentError = new Error('ENOENT') as NodeJS.ErrnoException;
+        enoentError.code = 'ENOENT';
+        
+        // Mock ENOENT for the full path and intermediate directories, then resolve the existing ancestor
+        mockFs.realpath
+          .mockRejectedValueOnce(enoentError) // C:\Users\test\a\b\c
+          .mockRejectedValueOnce(enoentError) // C:\Users\test\a\b
+          .mockRejectedValueOnce(enoentError) // C:\Users\test\a
+          .mockResolvedValueOnce(ancestorPath); // C:\Users\test
+        
+        const result = await validatePath(nestedPath);
+        expect(result).toBe(path.resolve(nestedPath));
+      });
+
+      it('rejects when ancestor directory resolves outside allowed directories', async () => {
+        const nestedPath = process.platform === 'win32' ? 'C:\\Users\\test\\symlink_out\\a\\b' : '/home/user/symlink_out/a/b';
+        const outsidePath = process.platform === 'win32' ? 'C:\\Windows\\System32' : '/etc';
+        
+        const enoentError = new Error('ENOENT') as NodeJS.ErrnoException;
+        enoentError.code = 'ENOENT';
+        
+        mockFs.realpath
+          .mockRejectedValueOnce(enoentError) // symlink_out/a/b
+          .mockRejectedValueOnce(enoentError) // symlink_out/a
+          .mockResolvedValueOnce(outsidePath); // symlink_out resolves outside
+        
+        await expect(validatePath(nestedPath))
+          .rejects.toThrow('Access denied - parent directory outside allowed directories');
+      });
+
+      it('rejects when no ancestor directory exists', async () => {
         const newFilePath = process.platform === 'win32' ? 'C:\\Users\\test\\nonexistent\\newfile.txt' : '/home/user/nonexistent/newfile.txt';
         
         // Create errors with the ENOENT code
-        const enoentError1 = new Error('ENOENT') as NodeJS.ErrnoException;
-        enoentError1.code = 'ENOENT';
-        const enoentError2 = new Error('ENOENT') as NodeJS.ErrnoException;
-        enoentError2.code = 'ENOENT';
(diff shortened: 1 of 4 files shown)


olaservo:
The `move_file` half of this landed via #4630. The `create_directory` / `validatePath` half (#4629) is being reviewed separately against the other open PRs for that issue.

olaservo:
Closing: the `move_file` half landed via #4630 and the `create_directory` half is covered by #4631. Thanks for the PR.

PR #4655, opened 2026-08-17
fix(sequentialthinking): restore nextThoughtNeeded in required schema and dynamic version

### LINKED ISSUE #4575: sequentialthinking: serverInfo.version hardcoded as "0.2.0" (published package is 2026.7.4)

### LINKED ISSUE #4651: sequential-thinking: nextThoughtNeeded missing from inputSchema.required but required at runtime (regression from #3533)

AbhiPra24:
## Summary

Fixes #4651 and #4575.

### 1. Restore `nextThoughtNeeded` in `inputSchema.required` (Fixes #4651)
- **Problem**: `nextThoughtNeeded` was previously wrapped in a `z.preprocess()`-based boolean coercion. Zod's `toJSONSchema(..., { io: "input" })` treats `z.preprocess()` input as `unknown`, silently dropping the field from the emitted `required` array while runtime validation still requires it, causing `-32602 Invalid params` for clients constructing calls from the advertised schema.
- **Fix**: Replaced `z.preprocess()` with a union transformation:
  ```ts
    z.boolean(),
    z.string().transform((v) => v.toLowerCase() === "true"),
  ]);
  ```
  This preserves the input types in the JSON Schema and retains `nextThoughtNeeded` in `required: ["thought", "nextThoughtNeeded", "thoughtNumber", "totalThoughts"]`.

### 2. Dynamic Server Version (Fixes #4575)
- Resolves server version dynamically from `package.json` rather than a hardcoded `"0.2.0"`.

## Verification
- Added integration test suite in `src/sequentialthinking/__tests__/index.test.ts`.
- `npm run build` and `npm test` passed (21/21 tests green).

---

### DIFF
--- a/src/sequentialthinking/__tests__/index.test.ts
+++ b/src/sequentialthinking/__tests__/index.test.ts
@@ -1,0 +1,175 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as path from 'path';
+import * as fs from 'fs';
+import { fileURLToPath } from 'url';
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+describe('SequentialThinking MCP Server Integration', () => {
+  let client: Client;
+  let transport: StdioClientTransport;
+
+  beforeEach(async () => {
+    const serverPath = path.resolve(__dirname, '../dist/index.js');
+    transport = new StdioClientTransport({
+      command: 'node',
+      args: [serverPath],
+      env: {
+        ...process.env,
+        DISABLE_THOUGHT_LOGGING: 'true',
+      },
+    });
+
+    client = new Client(
+      {
+        name: 'test-client',
+        version: '1.0.0',
+      },
+      {
+        capabilities: {},
+      }
+    );
+
+    await client.connect(transport);
+  });
+
+  afterEach(async () => {
+    await client?.close();
+  });
+
+  it('should match server version with package.json', async () => {
+    const pkg = JSON.parse(
+      fs.readFileSync(path.resolve(__dirname, '../package.json'), 'utf-8')
+    );
+    const serverVersion = client.getServerVersion();
+    expect(serverVersion?.version).toBe(pkg.version);
+    expect(serverVersion?.name).toBe('sequential-thinking-server');
+  });
+
+  it('should include nextThoughtNeeded in inputSchema.required', async () => {
+    const toolsResult = await client.listTools();
+    const tool = toolsResult.tools.find((t) => t.name === 'sequentialthinking');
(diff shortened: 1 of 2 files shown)


olaservo:
Closing in favour of #4652 (#4651) and #4576 (#4575, already merged). For reference, the catch-all `z.string().transform(v => v.toLowerCase() === "true")` branch maps any unrecognised string such as `"yes"` or `""` to `false` instead of rejecting it, and `getPackageVersion()` falls back to a new hardcoded literal, which reintroduces the drift #4575 is about.

PR #4656, opened 2026-08-17
fix(memory): atomic graph persistence and validate entity existence in create_relations

### LINKED ISSUE #4457: memory: create_relations accepts relations to non-existent entities (dangling relations, silent graph corruption)

### LINKED ISSUE #4614: saveGraph() in src/memory uses non-atomic fs.writeFile — risk of corrupted memory file on interruption

AbhiPra24:
## Summary

Fixes #4614 and #4457.

### 1. Atomic Persistence in `saveGraph()` (Fixes #4614)
- **Problem**: `saveGraph()` directly called `fs.writeFile(this.memoryFilePath, ...)`, which truncates the file immediately upon opening. If the process is terminated mid-write, the memory file becomes corrupted or empty.
- **Fix**: Write serialized lines to a temporary file in the same directory (`${this.memoryFilePath}.${randomBytes(16).toString('hex')}.tmp`), followed by an atomic `fs.rename` over `this.memoryFilePath`. The temporary file is cleaned up if an error occurs.

### 2. Referential Integrity in `createRelations()` (Fixes #4457)
- **Problem**: `createRelations()` only checked for duplicate relations, accepting dangling relations between non-existent entities without validation.
- **Fix**: Validates that all referenced `from` and `to` entities exist in `graph.entities` prior to mutating the graph, throwing `Entity with name <name> not found` to match `addObservations()`.

## Verification
- Added regression tests in `src/memory/__tests__/knowledge-graph.test.ts`.
- `npm run build` and `npm test` passed (56/56 tests green).

---

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { promises as fs } from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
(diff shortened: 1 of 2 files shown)


Selven81:
Thanks for this. I took the explicit `"utf-8"` on the write — it doesn't change
behaviour, but it makes the assumption visible, and none of the other PRs spell
it out.

On the relation validation: it's good work, but it addresses #4457, which is a
separate concern from persistence. Bundling it here makes both harder to review
in isolation — which I think is exactly why you opened #4696. I said as much in
the summary.

olaservo:
The atomic-write half of this landed via #4642. The `createRelations` entity validation is a separate behaviour change (#4457); if you rebase that part alone onto `main` as its own PR it can be reviewed on its own.

olaservo:
Closing per the comment above: the atomic-write half landed via #4642, so this branch no longer applies cleanly. The `createRelations` entity validation for #4457 is still welcome as its own PR against `main` — the tests here for missing `from`/`to` entities and the fail-fast batch case are the right shape to bring along.

PR #4282, opened 2026-06-05
Add schema constraints to time server strings

bmtriet:
## Summary
- add maxLength/pattern schema hints for time server timezone inputs
- add a bounded 24-hour time pattern for convert_time
- cover the advertised input schemas and representative pattern examples in tests

## Context
This is a small defensive follow-up for #3537. It only adds schema-level constraints for clients that validate tool schemas; existing runtime validation through ZoneInfo/time parsing remains the final check.

## Tests
- cd src/time && uv run pyright
- cd src/time && uv run ruff check src/mcp_server_time/server.py test/time_server_test.py
- cd src/time && uv run pytest

Issue #3537, opened 2026-03-12
Security Audit: Unconstrained string parameters across all official servers


manja316:
## Summary

I ran automated security audits on 7 official MCP servers using [mcp-security-audit](https://pypi.org/project/mcp-security-audit/). All servers scored Grade A or B (85-100/100), which is great. But there's one consistent finding across every server except `mcp-server-fetch`:

**No string parameters use `maxLength`, `pattern`, or `enum` constraints.**

## Findings

| Server | Score | Unconstrained Strings |
|--------|-------|-----------------------|
| Fetch | 100/100 | 0 |
| SQLite | 95/100 | 5 (SQL queries, table names) |
| Time | 95/100 | 4 (timezone identifiers) |
| GitHub | 94/100 | Multiple |
| Memory | 92/100 | 1 (search query) |
| Git | 90/100 | 18 (paths, messages, branches) |
| Filesystem | 85/100 | Multiple (paths) |

## Why this matters

1. **DoS vector** — Unconstrained strings accept arbitrarily long input. A 10MB commit message or 50K-char file path must be processed by the server.
2. **No boundary validation** — If an LLM is compromised via prompt injection, the tool schema provides no defense against malformed input.
3. **Network transports** — With MCP supporting Streamable HTTP, servers may be network-exposed where these constraints become critical.

## Suggested fix

One line per parameter:

```json
{
  "name": "path",
  "type": "string",
  "maxLength": 4096
}
```

For known-format params:
```json
{
  "name": "timezone",
  "type": "string",
  "pattern": "^[A-Za-z_/]+$",
  "maxLength": 50
}
```

## Full reports

Individual JSON/text reports for each server: [LuciferForge/mcp-audit-reports](https://github.com/LuciferForge/mcp-audit-reports)

The audit tool: `pip install mcp-security-audit`

Happy to discuss findings or help with fixes.

razashariff:
MCPS (MCP Secure) solves this at the message layer — every JSON-RPC body is SHA-256 hashed and ECDSA P-256 signed before transmission. A compromised proxy can't inject malformed parameters because the signature covers the full message body. Nonce + timestamp binding prevents replay. Designed as a drop-in security layer for MCP servers. [IETF Internet-Draft](https://datatracker.ietf.org/doc/draft-sharif-mcps-secure-mcp/) | [github.com/razashariff/mcps](https://github.com/razashariff/mcps)

joergmichno:
[...]

PR #4662, opened 2026-08-18
fix(memory): constrain search_nodes query length

feiiiiii5:
## Summary

Add a length constraint to the memory server's `search_nodes` `query` parameter, per #3537 (official servers' string parameters lack `maxLength`/`pattern` constraints, creating DoS and malformed-input risk).

## Root Cause

`src/memory/index.ts` registers `search_nodes` with `query: z.string()` — completely unbounded. Each call with an oversized query still triggers a full O(graph) scan over every entity name, type, and observation, for no retrieval value.

## Fix

Follows the pattern in #4282 (time server string constraints):

- `SEARCH_QUERY_MAX_LENGTH = 2048` named constant — generous for real retrieval queries, hard cap against abuse.
- Exported `SearchNodesQuerySchema` (`z.string().max(...)`) used by the tool's `inputSchema`, so the constraint is testable directly.

Scope kept to `query` as named in #3537's memory-server finding; other parameters can follow in separate PRs.

## Test

New `src/memory/__tests__/search-nodes-schema.test.ts` (vitest, per CONTRIBUTING):

- normal query accepted; query at exactly the limit accepted; over-limit rejected; non-string input still rejected.

`npm test` in `src/memory`: 4 test files, 54 tests passed (including the 4 new cases).

## Diff scope

`src/memory/index.ts` (+8/-1), new test file (+25).

### DIFF
--- a/src/memory/__tests__/search-nodes-schema.test.ts
+++ b/src/memory/__tests__/search-nodes-schema.test.ts
@@ -1,0 +1,25 @@
+import { describe, it, expect } from 'vitest';
+import { SearchNodesQuerySchema, SEARCH_QUERY_MAX_LENGTH } from '../index.js';
+
+describe('search_nodes input schema', () => {
+  it('should accept a normal query', () => {
+    expect(SearchNodesQuerySchema.safeParse('Alice').success).toBe(true);
+    expect(SearchNodesQuerySchema.safeParse('works at Acme Corp').success).toBe(true);
+  });
+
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Checking in on this one — happy to address any review feedback or adjust scope if helpful.

olaservo (approved):
Matches the #4282 pattern and the limit is generous enough not to affect real queries. Approving.

PR #4667, opened 2026-08-19
fix(filesystem): preserve UTF-8 across head and tail chunks

### LINKED ISSUE #4666: headFile/tailFile in filesystem server corrupts multi-byte UTF-8 characters at 1024-byte chunk boundaries
### Description
independently:

    ...

that position.

### Reproduction
   1024 bytes.
   that 1024-byte boundary.
   path.

### Suggested fix

### Environment
  build dated 2026-07-22)
- macOS, Claude Desktop

Excelius-Wang:
## Description

Fixes #4666.

Preserve multi-byte UTF-8 characters when `headFile()` and `tailFile()` read across 1024-byte chunk boundaries.

- Use `StringDecoder` for forward chunk decoding in `headFile()`.
- Keep reverse-read chunks as bytes in `tailFile()` and decode only after enough line separators have been collected.
- Add regression tests with a CJK character split across the chunk boundary in both directions.

## Server Details

- Server: filesystem
- Changes to: `head_file` and `tail_file`

## Motivation and Context

Decoding every chunk independently replaces split UTF-8 sequences with U+FFFD. This corrupts CJK and other multi-byte text returned by the filesystem server.

## How Has This Been Tested?

- `npm test` — 154 tests passed
- `npm run build`
- Added regression tests for `headFile()` and `tailFile()` with `界` split across a 1024-byte boundary

The change was tested at the filesystem operation boundary rather than through an LLM client.

## Breaking Changes

None.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] The code follows the repository style
- [x] No README update is required because the public interface and configuration are unchanged
- [x] No environment variables or configuration options were added

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -26,6 +26,23 @@ import {
 vi.mock('fs/promises');
 const mockFs = fs as any;
 
+function createMockFileHandle(content: Buffer) {
+  return {
+    read: vi.fn(
+      async (buffer: Buffer, offset: number, length: number, position: number) => {
+        const bytesRead = content.copy(
+          buffer,
+          offset,
+          position,
+          Math.min(position + length, content.length),
+        );
+        return { bytesRead, buffer };
+      },
+    ),
+    close: vi.fn().mockResolvedValue(undefined),
+  };
+}
+
 describe('Lib Functions', () => {
   beforeEach(() => {
     vi.clearAllMocks();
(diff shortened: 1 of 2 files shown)


olaservo (approved):
Of the four open PRs for #4666 this is the only one whose `tailFile` path is correct: raw bytes are accumulated and decoded once, and newline counting happens on bytes. `headFile` uses a `StringDecoder` with a final `end()`. Both tests place a multibyte character across the 1024-byte boundary.

PR #4669, opened 2026-08-19
fix(sequential-thinking): keep nextThoughtNeeded in inputSchema.required

### LINKED ISSUE #4651: sequential-thinking: nextThoughtNeeded missing from inputSchema.required but required at runtime (regression from #3533)

re2zero:
Closes #4651

## Problem
`nextThoughtNeeded` is declared as a required field at runtime but is missing from the `inputSchema.required` list. Client code that builds arguments from the advertised schema omits the field and gets `-32602`.

Root cause: `z.preprocess()` produces a schema whose input type accepts `unknown`, so zod-to-JSON-Schema treats it as optional and drops it from `required`.

## Fix
Replace `z.preprocess` with `z.union([z.boolean(), z.enum(["true", "false"]).transform(...)])`. This preserves boolean coercion while keeping the field correctly listed in `required`.

## Testing
- `server-sequential-thinking` test suite: **14 passed** (1 test file).

### DIFF
--- a/src/sequentialthinking/index.ts
+++ b/src/sequentialthinking/index.ts
@@ -6,14 +6,14 @@ import { z } from "zod";
 import { SequentialThinkingServer } from './lib.js';
 
 /** Safe boolean coercion that correctly handles string "false" */
-const coercedBoolean = z.preprocess((val) => {
-  if (typeof val === "boolean") return val;
-  if (typeof val === "string") {
-    if (val.toLowerCase() === "true") return true;
-    if (val.toLowerCase() === "false") return false;
-  }
-  return val;
-}, z.boolean());
+// Use a union (rather than z.preprocess) so that zod-to-JSON-Schema conversion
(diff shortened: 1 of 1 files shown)


olaservo:
Closing in favour of #4652 for #4651. For reference, `z.enum(["true", "false"])` rejects `"True"` and `"FALSE"`, which the pre-#3533 code accepted.

PR #4670, opened 2026-08-19
fix(filesystem): use StringDecoder for UTF-8 safe headFile/tailFile

### LINKED ISSUE #4666: headFile/tailFile in filesystem server corrupts multi-byte UTF-8 characters at 1024-byte chunk boundaries

re2zero:
Closes #4666

## Problem
`headFile()` and `tailFile()` read files in 1024-byte chunks and call `.toString('utf-8')` on each chunk independently. When a multi-byte UTF-8 character (e.g. CJK characters, which are 3 bytes) straddles a chunk boundary, the split decoding produces mojibike (U+FFFD replacement characters).

## Fix
Replace per-chunk `.toString('utf-8')` with Node.js `StringDecoder`:

- **headFile**: reads forward, the decoder naturally buffers incomplete trailing byte sequences across `write()` calls. A final `decoder.end()` flushes any remaining bytes.

- **tailFile**: reads backwards from file end. The function now collects raw `Buffer` slices in reverse order, reverses them for correct chronology, then decodes the concatenated buffer as a single UTF-8 stream. Newline counting is done on raw bytes (0x0A is single-byte even in UTF-8), so we stop reading at the right point.

## Testing
- `server-filesystem` test suite: **152 passed** (7 test files).

### DIFF
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -336,7 +336,6 @@ export async function tailFile(filePath: string, numLines: number): Promise<stri
   } finally {
     await fileHandle.close();
   }
-}  }
 }
 
 // Read the first N lines of a file (UTF-8 safe)
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -4,6 +4,7 @@ import os from 'os';
 import { randomBytes } from 'crypto';
 import { diffLines, createTwoFilesPatch } from 'diff';
 import { minimatch } from 'minimatch';
+import { StringDecoder } from 'string_decoder';
 import { normalizePath, expandHome } from './path-utils.js';
 import { isPathWithinAllowedDirectories } from './path-validation.js';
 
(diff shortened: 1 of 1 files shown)


tarunag10 (commented):
I compared this with #4667 and ran the filesystem suite plus two real-file edge cases. This PR passes the existing 46 tests, but both added edge cases fail at `fc760fe3`:

1. `tailFile(file, 2)` on `discard\\n` + `界` + 1017 ASCII bytes + `\\nlast` returns U+FFFD at the 1024-byte boundary. `rawBuffers.push(chunk.slice(...))` stores views into the single reused `chunk`; later reads overwrite earlier entries. Copying each read (`Buffer.from(chunk.subarray(0, bytesRead))`) avoids that aliasing.
2. `tailFile(file, 0)` returns the whole file because `newlineCount <= numLines` enters the loop and `slice(-0)` is `slice(0)`.

The same two tests pass on #4667, which also has focused head/tail boundary tests (48/48 filesystem tests versus 46/46 here). I recommend either adopting the copying and zero-line behavior here with regression tests, or closing this in favor of #4667. There is also an unused `relevantLines` local in the current diff.

olaservo:
Closing; #4667 has been merged for #4666. For reference, this version had a bug: `rawBuffers.push(chunk.slice(0, bytesRead))` stores a view into the reused `chunk`, so each read overwrites the buffers pushed before it and any file longer than one chunk comes back corrupted.

PR #4674, opened 2026-08-20
fix(filesystem): ignore trailing newline in tail

jstar0:
## Description

Fix `tailFile` so a final newline is treated as a line terminator rather than an additional empty line. This makes `tail=1` return the final content line for files that end with `\n` or `\r\n`.

## Server Details

- Server: filesystem
- Changes to: file utility implementation and regression tests

## Motivation and Context

When a file ends with a newline, splitting the collected tail text produces a synthetic empty segment after the final terminator. Callers requesting the last line therefore receive an empty string instead of the final content line. The fix detects a trailing LF at the byte level, collects one additional newline delimiter so chunked reads still return complete content lines, and removes only the synthetic segment after the chunks have been decoded and line endings normalized. Existing UTF-8 chunk-boundary behavior is preserved.

## How Has This Been Tested?

- `npm exec --workspace @modelcontextprotocol/server-filesystem -- vitest run __tests__/lib.test.ts --reporter=dot` - 58 passed
- `npm run build --workspace @modelcontextprotocol/server-filesystem` - passed
- `npm exec --workspace @modelcontextprotocol/server-filesystem -- vitest run --coverage --exclude __tests__/unicode-paths.test.ts` - 167 passed
- Runnable cases from `unicode-paths.test.ts` - 2 passed

The regressions cover LF and CRLF endings and a trailing newline whose requested content crosses the 1 KiB read boundary. The remaining Unicode fixture cannot be constructed on macOS/APFS because creating the second canonically equivalent directory fails with `EEXIST` during test setup, before the filesystem server code under test runs.

## Breaking Changes

No. This corrects the returned value for files ending in a newline; files without a trailing newline retain their existing behavior.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have updated the server README accordingly
- [x] My code follows the repository style guidelines

## Additional context

The change is limited to the existing tail operation in the filesystem server and adds no protocol, configuration, or dependency changes.

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -643,6 +643,33 @@ describe('Lib Functions', () => {
         expect(mockFileHandle.close).toHaveBeenCalled();
       });
 
+      it('ignores trailing newline when returning last lines', async () => {
+        const content = Buffer.from('line1\nline2\n');
+        mockFs.stat.mockResolvedValue({ size: content.length } as any);
+
+        const mockFileHandle = {
+          read: vi.fn(
+            (buffer: Buffer, offset: number, length: number, position: number) => {
+              const bytesRead = content.copy(
+                buffer,
+                offset,
+                position,
+                Math.min(position + length, content.length)
+              );
+              return Promise.resolve({ bytesRead });
+            }
+          ),
+          close: vi.fn().mockResolvedValue(undefined)
+        } as any;
+
+        mockFs.open.mockResolvedValue(mockFileHandle);
+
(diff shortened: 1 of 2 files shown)


jstar0:
Merged current `main` into head `473df2e2` to resolve the branch conflicts while preserving the two-file PR scope. GitHub Actions are green on the current head (16/16 checks), and the focused filesystem tests pass 58/58 with the filesystem build also passing. The regressions now cover LF, CRLF, and a trailing newline across the 1 KiB read boundary while retaining the upstream UTF-8 boundary coverage. A fresh review of the current head would be appreciated.

PR #4679, opened 2026-08-22
Fix create_directory for nested paths with non-existent parents

### LINKED ISSUE #4629: server-filesystem: create_directory does not create parent directories

sachinsharma3191:
## Summary

Fixes #4629

- Added `ValidatePathOptions` interface with `allowNonExistentAncestors` flag to `validatePath()`
- When the flag is set, walks up via `path.dirname()` to find the first existing ancestor and validates it against allowed directories
- `create_directory` handler now passes `{ allowNonExistentAncestors: true }` so `mkdir -p` style recursive creation works correctly
- 4 new tests covering: allowed deep paths, paths outside allowed dirs, default behavior unchanged, and root walkup exhaustion

## Test plan

- [x] `create_directory` with nested non-existent parents succeeds when ancestor is in allowed dirs
- [x] `create_directory` still rejects paths outside allowed directories
- [x] Default `validatePath()` behavior (without flag) unchanged — existing tests pass
- [x] All 54 tests pass

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -190,19 +190,80 @@ describe('Lib Functions', () => {
 
       it('rejects when parent directory does not exist', async () => {
         const newFilePath = process.platform === 'win32' ? 'C:\\Users\\test\\nonexistent\\newfile.txt' : '/home/user/nonexistent/newfile.txt';
-        
+
         // Create errors with the ENOENT code
         const enoentError1 = new Error('ENOENT') as NodeJS.ErrnoException;
         enoentError1.code = 'ENOENT';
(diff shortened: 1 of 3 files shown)


olaservo:
Closing in favour of #4631 for #4629. For reference, this version has a security gap: the ancestor walk sits inside the existing bare `catch {}`, so when the immediate parent is a symlink out of the allowed tree (`allowed/link -> /etc`, request `allowed/link/a`) the `Access denied` thrown by the parent check is swallowed, the walk restarts from the allowed directory, and `mkdir -p` would create `/etc/a`.

Issue #3628, opened 2026-03-18
Docker Image 1.0.2 Breaks Linux Users with Letter at Start of Path


jvanvolken:
**Describe the bug**
This 1.0.2 breaks linux users who's paths start with a letter. i.e. `/h/path/to/data`. We were very confused when our linux paths were converted to windows paths: 

This error does not happen with `mcp/filesystem:sha256-35fcf0217ca0d5bf7b0a5bd68fb3b89e08174676c0e0b4f431604512cf7b3f67`; it only happens with `mcp/filesystem:1.0.2`.

**To Reproduce**
1. Docker pull `mcp/filesystem:1.0.2`
2. Configure with a linux path that starts with a /h/ or other letter
3. Start up mcp server

**Expected behavior**
Should mount the correct path without changing it.

**Logs**
This is the error when the gateway tries to configure the mcp server:
`Running mcp/filesystem:1.0.2 with [run --rm -i --init --security-opt no-new-privileges --cpus 1 --memory 2Gb --pull never -l docker-mcp=true -l docker-mcp-tool-type=mcp -l docker-mcp-name=filesystem -l docker-mcp-transport=stdio --network none -v /h/<username>/MCP_Development/data:/h/<username>/MCP_Development/data] and command [/h/<username>/MCP_Development/data]`

```
mcp-gateway  | - filesystem: Error accessing directory H:\<username>\MCP_Development\data: Error: ENOENT: no such file or directory, stat 'H:\<username>\MCP_Development\data'
mcp-gateway  | - filesystem:     at async Object.stat (node:internal/fs/promises:1039:18)
mcp-gateway  | - filesystem:     at async file:///app/dist/index.js:44:23
mcp-gateway  | - filesystem:     at async Promise.all (index 0)
mcp-gateway  | - filesystem:     at async file:///app/dist/index.js:42:1 {
mcp-gateway  | - filesystem:   errno: -2,
mcp-gateway  | - filesystem:   code: 'ENOENT',
mcp-gateway  | - filesystem:   syscall: 'stat',
mcp-gateway  | - filesystem:   path: 'H:\\<username>\\MCP_Development\\data'
mcp-gateway  | - filesystem: }
mcp-gateway  |   > Can't start filesystem: failed to connect: calling "initialize": EOF
```

**Additional context**
N/A

owendevereaux:
I investigated this issue and found something interesting:

[...]

Issue #3527, opened 2026-03-11
UNC/network share paths (\\server\share\subdir) fail access check despite being under allowed directory


Drakhanas:
Environment:

Claude Desktop App (Windows)
MCP Filesystem Server
Windows 11, accessing SMB shares on a QNAP NAS
@modelcontextprotocol/server-filesystem version: 2026.1.14
node.js version: v24.13.1

Configured allowed directories (from list_allowed_directories):
\\192.168.4.96\Plex\ 
\\192.168.4.96\Mega\
C:\
D:\
E:\

Issue:
Filesystem server can list the root of a UNC allowed directory but rejects access to any subdirectory within it, even though subdirectory access should be permitted by the path prefix check.

Reproduction steps:

Configure \192.168.4.96\Mega\ as an allowed directory
Call list_directory with path \192.168.4.96\Mega — succeeds, returns full directory listing
Call list_directory with path \192.168.4.96\Mega\Drops — fails with: Access denied - path outside allowed directories: \192.168.4.96\Mega\Drops not in [allowed list]

Observations:

NOT a spaces-in-path issue. Tested with Drops (no spaces) and 00R RPGs (has space) — both fail identically.
Local drive paths (C:, D:, E:) appear to handle subdirectory traversal correctly.
Bug is specific to UNC paths — the prefix matching logic doesn't recognize \server\share\subfolder as a child of \server\share.
Exact root match works (listing the share itself), but any path one level deeper fails the access check.

Expected behavior:

If \192.168.4.96\Mega\ is an allowed directory, all paths under it (e.g., \192.168.4.96\Mega\Drops, \192.168.4.96\Mega\00R RPGs) should pass the access check, consistent with how local drive paths behave.

Impact:

All UNC/network share allowed directories are effectively limited to root-level listing only. No subdirectory navigation is possible, making NAS-hosted file access non-functional for any real use.

Additional note:

Windows mapped drive letters (ie B: mapped to \192.168.4.96\Mega) are not visible to the MCP server process, presumably because mapped drives are per user/per session and the server process doesn't inherit them. This means UNC paths are the only way to access network shares making this bug a complete blocker for NAS-based workflows.

gdimas38-eng:
Confirmed, with the root cause and a reproduction.

`path.resolve()` returns a UNC **share root** with a trailing separator — the same way it does for a drive root — whether or not one was supplied:

```js
[...]

PR #4689, opened 2026-08-23
fix(filesystem): reject Windows-style paths on POSIX hosts instead of writing literal filenames

### LINKED ISSUE #4686: filesystem: a Windows-style path is silently accepted on POSIX and becomes a literal filename inside the allowed root
**Describe the bug**

On a POSIX host, passing a Windows-style path such as `C:\Users\me\notes\file.md` to `write_file` does not fail. It succeeds, and the server answers `Successfully wrote to C:\Users\me\notes\file.md`.

What actually happens is that a single file is created directly inside the allowed root, whose *filename* is the entire string, backslashes included:

```
/Users/me/C:\Users\me\notes\file.md
```

This is **not** a containment failure — nothing escapes the sandbox, and `isPathWithinAllowedDirectories` is doing its job. It is the *absence of a rejection* for a string that can only mean something on another OS. The caller believes it wrote to a Windows location; what exists is an unpronounceable file at the root of the allowed directory, and the success message gives no hint. In our case the file sat unnoticed for about three weeks.

**Why it happens**

In `src/filesystem/lib.ts`:

```ts
export async function validatePath(requestedPath: string): Promise<string> {
    ? path.resolve(expandedPath)
    : resolveRelativePathAgainstAllowedDirectories(expandedPath);
```

Under POSIX, `path.isAbsolute("C:\\Users\\me\\notes\\file.md")` is `false`. The string therefore takes the *relative* branch, and `resolveRelativePathAgainstAllowedDirectories` does `path.resolve(allowedDir, thatString)`. The result lands **inside** the allowed root, so `isPathWithinAllowedDirectories` legitimately approves it, and the write proceeds.

The same reasoning applies to `C:/Users/me/...` — there the directory tree `C:/Users/...` is created inside the allowed root instead.

**To Reproduce**

1. Run the server on macOS or Linux with a single allowed directory, e.g. `/tmp/sandbox`.
2. Call `write_file` with `path: "C:\\Users\\me\\notes\\file.md"` and any content.
3. Observe the response: `Successfully wrote to C:\Users\me\notes\file.md`.
4. `ls /tmp/sandbox` shows one entry named `C:\Users\me\notes\file.md`.

**Expected behavior**

`validatePath` should refuse a string that matches a Windows drive-letter prefix (`^[A-Za-z]:[\\/]`, and the bare `^[A-Za-z]:$`) when `process.platform !== 'win32'`, instead of silently treating it as a relative path. An explicit error — something like `Access denied - Windows-style path received on a POSIX host` — turns a silent, late-discovered mistake into an immediate one.

A narrower alternative, if outright rejection is considered too strict: keep accepting it but never let a drive-letter string reach the relative branch, and say so in the response.

**Please be careful not to reintroduce #3628**

That report is the mirror image of this one: a Linux path like `/h/user/data` was converted into `H:\user\data`. The fix proposed here is deliberately one-directional and platform-guarded — reject the drive-letter *form*, only when not running on Windows, and never rewrite POSIX paths.

**Environment**

- `@modelcontextprotocol/server-filesystem` — reproduced against the code currently on `main` (`src/filesystem/lib.ts`, the `validatePath` / `resolveRelativePathAgainstAllowedDirectories` pair quoted above).
- Also reproduced against the shipped Claude Desktop Filesystem extension, version `2026.7.10`, on macOS with Node v24.19.0.

Parker-Fawcett:
## Description

On a POSIX host, `validatePath()` accepts Windows drive-letter paths like `C:\Users\me\notes\file.md` as if they were relative paths. `path.resolve(allowedDir, requested)` then places them inside the allowed root as a single entry literally named `C:\Users\me\notes\file.md` — or, for `C:/Users/me/...`, creates an actual `C:/Users/...` directory tree. The tool answers "Successfully wrote to ...", so the mistake is only discovered when inspecting disk.

`validatePath()` now rejects strings matching a Windows drive-letter form (`^[A-Za-z]:(?:[\\/]|$)`) up front whenever `process.platform !== 'win32'`, with an explicit error:

```
Access denied - Windows-style path received on a POSIX host: C:\Users\me\notes\file.md
```

Fixes #4686.

## Server Details
- Server: filesystem
- Changes to: path validation (`validatePath` in `src/filesystem/lib.ts`) + tests

## Motivation and Context

Reported in #4686 with a full root-cause analysis: on POSIX `path.isAbsolute("C:\\...")` is false, so the string falls into `resolveRelativePathAgainstAllowedDirectories()` and resolves *inside* the sandbox. Nothing escapes containment — `isPathWithinAllowedDirectories()` is working correctly — but the caller's intent is silently misinterpreted, and the resulting file is easy to overlook (the reporter's sat unnoticed for three weeks).

The guard is deliberately one-directional and platform-guarded:
- Only active when `process.platform !== 'win32'`, so Windows behavior is untouched.
- POSIX paths are never rewritten or normalized into Windows form, avoiding the mirror-image regression described in #3628.
- The rejection happens before any filesystem access and before `expandHome`, failing fast with an actionable message.

## How Has This Been Tested?

- New tests in `src/filesystem/__tests__/lib.test.ts` (POSIX-only via `describe.skipIf(process.platform === 'win32')`, following the file's existing platform-conditional style):
  - Rejects `C:\Users\me\notes\file.md`, `C:/Users/me/file.md`, `Z:\`, and bare `C:` with the explicit error.
  - Rejection occurs before any filesystem access (`realpath` never called).
  - Regression guard: relative paths that merely contain a colon after the first character (`notes/file:C.md`) still resolve normally on all platforms.
- Full suite green: filesystem **158/158** (152 pre-existing + 6 new).
- Live verification against the real filesystem on macOS: all four drive-letter forms now throw the new error; ordinary absolute and relative paths are unaffected.

## Breaking Changes

Behavioral change only for inputs that were already broken in intent: passing a Windows path to a server running on macOS/Linux previously produced a misleading success plus a junk file; it now produces an immediate, descriptive error. No valid workflow is affected, and no client configuration changes are needed.

## Types of changes
- [ ] Documentation update

## Checklist
- [ ] I have updated the server's README accordingly (N/A — error-path hardening, no tool surface change)
- [ ] I have tested this with an LLM client (verified via direct `validatePath` invocation over the built server module plus unit suites; N/A granularity for a validation-layer fix)
- [ ] I have documented all environment variables and configuration options (N/A)

## Additional context

The regex intentionally follows the form proposed in #4686 (`^[A-Za-z]:[\\/]` plus bare `^[A-Za-z]:$`, combined here as `^[A-Za-z]:(?:[\\/]|$)`). Known trade-off, accepted per that issue's spec: a POSIX filename that genuinely begins with a single letter followed by a colon (e.g. `a:b/c`) will now be rejected on POSIX hosts; such names are vanishingly rare in practice and were previously indistinguishable from this bug class anyway.

Happy to follow up with the same treatment for UNC forms (`\\server\share`) on POSIX if maintainers want them rejected too (#3527 covers the Windows side of UNC handling).

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -228,6 +228,30 @@ describe('Lib Functions', () => {
           process.cwd = originalCwd;
         }
       });
+
+      describe.skipIf(process.platform === 'win32')('Windows-style paths on POSIX hosts', () => {
+        it.each([
+          'C:\\Users\\me\\notes\\file.md',
+          'C:/Users/me/file.md',
+          'Z:\\',
+          'C:'
+        ])('rejects %s instead of treating it as a relative path', async (windowsPath) => {
+          await expect(validatePath(windowsPath))
+            .rejects.toThrow('Access denied - Windows-style path received on a POSIX host');
+        });
+
+        it('rejects before touching the filesystem', async () => {
+          await expect(validatePath('C:\\Users\\me\\notes\\file.md')).rejects.toThrow();
+          expect(mockFs.realpath).not.toHaveBeenCalled();
+        });
+      });
+
+      it('still resolves relative paths that merely contain a colon after the first character', async () => {
+        const colonPath = process.platform === 'win32' ? 'C:\\Users\\test\\notes\\file:C.md' : 'notes/file:C.md';
+        const result = await validatePath(colonPath);
+        const expectedBase = process.platform === 'win32' ? 'C:\\Users\\test' : '/home/user';
+        expect(result).toBe(path.resolve(expectedBase, colonPath));
+      });
     });
   });
 
--- a/src/filesystem/lib.ts
+++ b/src/filesystem/lib.ts
@@ -97,6 +97,11 @@ function resolveRelativePathAgainstAllowedDirectories(relativePath: string): str
 
 // Security & Validation Functions
 export async function validatePath(requestedPath: string): Promise<string> {
+  // Security: reject Windows drive-letter paths on POSIX hosts; treating them as
+  // relative paths would silently create literal files like "C:\Users\me\file" in the allowed root.
+  if (process.platform !== 'win32' && /^[A-Za-z]:(?:[\\/]|$)/.test(requestedPath)) {
+    throw new Error(`Access denied - Windows-style path received on a POSIX host: ${requestedPath}`);
+  }
   const expandedPath = expandHome(requestedPath);
   const absolute = path.isAbsolute(expandedPath)
     ? path.resolve(expandedPath)


Alpha-W0lf (commented):
Independently reproduced #4686 on `main` before testing this PR: `write_file` with `C:\Users\me\notes\file.md` succeeds and creates a literal backslash filename inside the allowed root.

Verified this PR against that repro — the drive-letter guard rejects with `isError` before any filesystem touch, and the full filesystem suite passes (52/52). Fix is minimal, matches the reject-don't-normalize expectation in the issue, and the tests cover the drive-letter variants plus the colon-after-first-character case.

One scoped note: plain `dir\file.txt` (backslash, no drive letter) is still accepted on POSIX and still becomes a literal filename — consistent with the issue's stated scope, but worth tracking alongside the UNC follow-up.

PR #4697, opened 2026-08-25
fix(filesystem): allow create_directory to create nested parent directories

### LINKED ISSUE #4629: server-filesystem: create_directory does not create parent directories

AbhiPra24:
## Description

Fixes #4629.

The tool description and documentation for `create_directory` explicitly state that it can create multiple nested directories in one operation ("Creates parent directories if needed", "Can create multiple nested directories in one operation"). However, `validatePath()` rejected paths when their immediate parent did not yet exist with `"Parent directory does not exist: <path>"`.

This PR adds an `allowMissingParents` parameter to `validatePath()` (defaulting to `false` for other file operations to maintain strict behavior). When enabled for `create_directory`, `validatePath()` securely verifies all ancestors up to the closest existing directory against the allowed directories list, ensuring path security is maintained while allowing `fs.mkdir(validPath, { recursive: true })` to create intermediate directories as documented.

## Changes
- Update `validatePath` in `src/filesystem/lib.ts` to accept `allowMissingParents: boolean = false` and securely walk up to the nearest existing ancestor.
- Update `create_directory` in `src/filesystem/index.ts` to pass `allowMissingParents = true`.
- Add unit test in `src/filesystem/__tests__/lib.test.ts` verifying `allowMissingParents: true` succeeds for nested non-existent directory hierarchies within allowed roots.

## Verification
- Ran `npm test` and `npm run build` in `src/filesystem`. All 153 tests pass.

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -205,6 +205,23 @@ describe('Lib Functions', () => {
           .rejects.toThrow('Parent directory does not exist');
       });
 
+      it('allows missing parent directory when allowMissingParents is true', async () => {
+        const allowedDir = process.platform === 'win32' ? 'C:\\Users\\test' : '/home/user';
+        const deeplyNestedPath = process.platform === 'win32' ? 'C:\\Users\\test\\deep\\nested\\dir' : '/home/user/deep/nested/dir';
+
+        const enoentError = new Error('ENOENT') as NodeJS.ErrnoException;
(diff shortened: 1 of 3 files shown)


olaservo:
Closing in favour of #4631 for #4629. Thanks for the PR.

PR #4701, opened 2026-08-27
fix(sequentialthinking): keep nextThoughtNeeded in inputSchema `required`

### LINKED ISSUE #4651: sequential-thinking: nextThoughtNeeded missing from inputSchema.required but required at runtime (regression from #3533)

Golchic:
## Description

`nextThoughtNeeded` is declared as a non-optional field on the `sequentialthinking`
tool, but it is missing from the generated JSON Schema's `required` array. Clients
that build call arguments from the advertised schema omit it and get
`-32602 Input validation error`, which contradicts the schema.

Root cause: the `coercedBoolean` helper was `z.preprocess(fn, z.boolean())`. Under
Zod v4, `z.preprocess` widens the *input* type to `unknown`, so the SDK's JSON
Schema generation (`io: "input"`) treats the field as optional and drops it from
`required` — even though it is not `.optional()`. (Regression from #3533.)

Fix: reimplement `coercedBoolean` as `z.union([z.boolean(), z.string().transform(...)])`.
The union keeps a concrete input type, so non-optional fields stay in `required`,
while still coercing `"true"`/`"false"` (case-insensitively) and now rejecting other
strings with a clearer message. `isRevision` / `needsMoreThoughts` remain optional.

The only schema shape change: these boolean fields now serialize as
`{ "anyOf": [{ "type": "boolean" }, { "type": "string" }] }`, which accurately
reflects the accepted input.

## Server Details
- Server: sequentialthinking
- Changes to: tools (input schema for `sequentialthinking`)

## Motivation and Context

Fixes #4651. Without this, schema-driven MCP clients cannot call the tool without
manually knowing `nextThoughtNeeded` is required despite the schema saying otherwise.

## How Has This Been Tested?

- New `src/sequentialthinking/__tests__/schema.test.ts`:
  - Round-trips a probe tool through the MCP SDK (`McpServer` + `Client` over
    `InMemoryTransport`) and asserts a required `coercedBoolean` field stays in
    `inputSchema.required`, while an `.optional()` one does not.
  - Value coercion: `true`/`false`, `"true"`/`"false"`/`"TRUE"`/`"False"`, and
    rejection of `"yes"`, `""`, `1`, `null`.
- `npm run build` and `npm test` pass in `src/sequentialthinking` (19 tests).
- Verified against the SDK's actual conversion path (Zod v4 → `zod/v4-mini`
  `toJSONSchema`, `io: "input"`), not a hand-rolled converter.

## Breaking Changes

None for tool callers — the field was always meant to be required and the runtime
validation already required it. MCP client configs need no changes.

## Types of changes

## Checklist
- [ ] I have updated the server's README accordingly — not needed; the README
      already lists `nextThoughtNeeded` as a non-optional `(boolean)` input
- [ ] I have tested this with an LLM client — tested with the MCP SDK client and
      automated tests
- [x] I have documented all environment variables and configuration options — n/a

## Additional context

`coercedBoolean` moved from `index.ts` to `lib.ts` (exported) so it can be unit
tested; `index.ts` now imports it. `zod` is added as an explicit dependency of the
package since it is imported directly (matching the `everything` server).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3906,7 +3906,8 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.30.0",
         "chalk": "^5.3.0",
-        "yargs": "^17.7.2"
+        "yargs": "^17.7.2",
+        "zod": "^4.0.0"
       },
       "bin": {
         "mcp-server-sequential-thinking": "dist/index.js"
--- a/src/sequentialthinking/__tests__/schema.test.ts
+++ b/src/sequentialthinking/__tests__/schema.test.ts
@@ -1,0 +1,57 @@
+import { describe, it, expect } from 'vitest';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js';
+import { coercedBoolean } from '../lib.js';
+
+describe('coercedBoolean - value coercion', () => {
+  it('accepts real booleans unchanged', () => {
+    expect(coercedBoolean.parse(true)).toBe(true);
+    expect(coercedBoolean.parse(false)).toBe(false);
+  });
+
+  it('coerces the strings "true"/"false" (case-insensitive)', () => {
+    expect(coercedBoolean.parse('true')).toBe(true);
+    expect(coercedBoolean.parse('false')).toBe(false);
+    expect(coercedBoolean.parse('TRUE')).toBe(true);
+    expect(coercedBoolean.parse('False')).toBe(false);
+  });
+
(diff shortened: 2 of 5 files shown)


Golchic:
Superseded by #4652, which landed the same fix. Closing — thanks!

PR #4703, opened 2026-08-28
fix(filesystem): preserve UTF-8 across head/tail chunks

### LINKED ISSUE #4666: headFile/tailFile in filesystem server corrupts multi-byte UTF-8 characters at 1024-byte chunk boundaries

CryoThrust:
## What does this PR do?

Fixes #4666.

`headFile` and `tailFile` previously decoded each 1024-byte read independently. A UTF-8 character split across a chunk boundary could therefore become replacement characters.

This change keeps tail reads as raw bytes until the selected range is assembled, then decodes once, and uses Node's `StringDecoder` for forward head reads so incomplete trailing sequences are buffered across reads.

## Verification

- `git diff --check` passes.
- Local test execution was attempted, but the installed Node binary cannot start because Homebrew's `simdjson` dylib is missing (`dyld: Library not loaded`).

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -718,6 +718,26 @@ describe('Lib Functions', () => {
     });
 
     describe('headFile', () => {
+      it('preserves UTF-8 characters split across read chunks', async () => {
+        const content = Buffer.concat([
+          Buffer.from('a'.repeat(1023)),
+          Buffer.from('中\nsecond line', 'utf8'),
+        ]);
+        const mockFileHandle = {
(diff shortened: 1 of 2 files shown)


CryoThrust:
The updated branch now includes a regression test for a UTF-8 character split at the 1024-byte read boundary. Build and filesystem tests are green on the latest commit. Ready for review.

olaservo:
Closing in favour of #4667, which has been merged for #4666. Thanks for the PR.

PR #4705, opened 2026-08-28
fix(filesystem): allow recursive directory creation

### LINKED ISSUE #4629: server-filesystem: create_directory does not create parent directories

CryoThrust:
## Summary

- resolve the nearest existing ancestor when validating a new path
- preserve realpath checks for symlink escapes while allowing nested paths
- make the existing recursive `create_directory` contract work for arbitrarily deep paths
- update validation coverage for missing intermediate directories

Closes #4629

## Validation

- `git diff --check`
- TypeScript/Vitest execution is currently blocked on this host because the Homebrew Node 22 binary cannot load `libsimdjson.29.dylib`.

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -190,21 +190,22 @@ describe('Lib Functions', () => {
         expect(result).toBe(path.resolve(newFilePath));
       });
 
-      it('rejects when parent directory does not exist', async () => {
+      it('walks up to the nearest existing ancestor for nested paths', async () => {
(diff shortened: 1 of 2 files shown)


CryoThrust:
Closing this PR because #4654 already covers the same recursive create_directory path-validation fix. Avoiding duplicate implementation.

PR #4707, opened 2026-08-28
fix(memory): persist graph snapshots atomically

### LINKED ISSUE #4614: saveGraph() in src/memory uses non-atomic fs.writeFile — risk of corrupted memory file on interruption

CryoThrust:
## Summary

- write memory graph snapshots to a unique temporary file
- atomically replace the live JSONL file with `rename`
- clean up temporary files on both success and failure
- add regression coverage ensuring persistence leaves no temporary artifact

Closes #4614

## Validation

- `git diff --check`
- Local Vitest execution is blocked because the Homebrew Node 22 binary cannot load `libsimdjson.29.dylib`; the change is covered by the repository test suite in CI.

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -59,6 +59,16 @@ describe('KnowledgeGraphManager', () => {
       const newEntities = await manager.createEntities([]);
       expect(newEntities).toHaveLength(0);
     });
+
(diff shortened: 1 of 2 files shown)


CryoThrust:
Closing because earlier PRs #4656 and #4696 already cover atomic memory graph persistence. Avoiding duplicate implementation.

PR #4708, opened 2026-08-28
fix(filesystem): prevent UTF-8 corruption across 1KB chunk boundaries in tailFile and headFile (#4666)

### LINKED ISSUE #4666: headFile/tailFile in filesystem server corrupts multi-byte UTF-8 characters at 1024-byte chunk boundaries

teddiesloco:
## Summary

Fixes #4666.

In `src/filesystem/lib.ts`, `tailFile` and `headFile` were converting each 1KB raw buffer chunk directly to a string using `chunk.slice(0, bytesRead).toString('utf-8')` before string concatenation. When multi-byte UTF-8 sequences (such as emojis, accented Latin, Vietnamese, or Asian scripts) fall across the 1024-byte buffer boundary, this produced invalid UTF-8 replacement characters (``) and data corruption.

This PR:
- Buffers raw binary chunks using `Buffer.concat` before UTF-8 decoding.
- Keeps remaining partial line bytes as raw `Buffer` across iterations.
- Adds regression unit test in `src/filesystem/__tests__/lib.test.ts` verifying multi-byte UTF-8 integrity across 1024-byte boundaries.

## Test Plan
- Run test suite: `npm test` in `src/filesystem` (158/158 passed).

### DIFF
--- a/src/filesystem/__tests__/lib.test.ts
+++ b/src/filesystem/__tests__/lib.test.ts
@@ -715,6 +715,34 @@ describe('Lib Functions', () => {
         
         expect(mockFileHandle.close).toHaveBeenCalled();
       });
+
+      it('handles multi-byte UTF-8 characters across chunk boundaries without corruption', async () => {
+        // Construct a string with emoji or multi-byte unicode characters placed across 1024-byte boundary
+        const pad = 'a'.repeat(1020);
+        const multibyteChar = '🌟'; // 4 bytes in UTF-8
+        const line1 = 'first line';
+        const line2 = `${pad}${multibyteChar}`;
+        const line3 = 'third line';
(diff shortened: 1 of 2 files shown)


olaservo:
Closing; #4667 has been merged for #4666. For reference, this version did not fix the boundary case: the partial first line is stored as `Buffer.from(chunkLines[0], 'utf-8')`, which re-encodes text that was already decoded with the replacement character baked in. The test does not straddle the 1024-byte boundary (the emoji sits inside one chunk), so it passes on `main` unchanged.

PR #4712, opened 2026-08-29
fix(everything): clean up subscriptions on session disconnect

### LINKED ISSUE #4710: everything: subscriptions map never drops a disconnected session

arimu1:
## Summary

- Add `removeSubscriber(sessionId)` in `src/everything/resources/subscriptions.ts` to remove a disconnected session from every URI subscriber set and drop empty map entries.
- Call `removeSubscriber(sessionId)` from the server factory `cleanup()` alongside the existing logging, resource-update interval, and task-store teardown.
- Add unit tests verifying disconnected sessions stop receiving simulated resource updates and that other sessions remain subscribed.

Fixes #4710

## Test plan

- [x] `npx vitest run __tests__/resources.test.ts __tests__/server.test.ts` — 31/31 passing
- [x] `npm run build` — clean TypeScript compile

### DIFF
--- a/src/everything/__tests__/resources.test.ts
+++ b/src/everything/__tests__/resources.test.ts
@@ -19,9 +19,13 @@ import {
 } from '../resources/session.js';
 import { registerFileResources } from '../resources/files.js';
 import {
+  SubscribeRequestSchema,
+} from '@modelcontextprotocol/sdk/types.js';
+import {
   setSubscriptionHandlers,
   beginSimulatedResourceUpdates,
   stopSimulatedResourceUpdates,
+  removeSubscriber,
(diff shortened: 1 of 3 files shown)


JosephDoUrden (commented):
Pulled 3d68bcb, 109 passing vs 107 on base, build ok, merge-tree onto main clean. Stubbed removeSubscriber to a no-op and the drop test fails. Same result through a real Client over InMemoryTransport with two uris and a second session.

One gap, nothing calls removeSubscriber through cleanup, I deleted that call from server/index.ts and all 109 still pass. @arimu1 worth one case through cleanup.

Fix is right, looks merge-ready to me.

olaservo (approved):

PR #4715, opened 2026-08-29
fix(everything): cleanup subscriptions on session disconnect

### LINKED ISSUE #4710: everything: subscriptions map never drops a disconnected session

dlowzzxx:
Fixes #4710 — resources/subscribeadded tosubscriptionsmap butcleanup()never removed session on disconnect, leaking sessionIds. AddremoveSubscriber(sessionId)iterating map and deleting empty entries, called fromserver/index.ts:cleanup(). Verified with npm test (107 passed).


olaservo:

PR #4716, opened 2026-08-29
everything: remove disconnected sessions from resource subscriptions

### LINKED ISSUE #4710: everything: subscriptions map never drops a disconnected session

vishalhabib99:
`subscriptions` (`Map<uri, Set<sessionId>>`) only ever had entries removed via an explicit `resources/unsubscribe` call. A session that disconnects without unsubscribing stayed in every `Set` it had joined for the life of the process, since `cleanup(sessionId)` stopped logging/update intervals and task-store timers but never touched `subscriptions`.

Adds `removeSubscriber(sessionId)`, which drops the session from every URI's subscriber set and removes any URI entry left with no subscribers, and calls it from `cleanup()`.

Fixes #4710

(Re-opened as a fresh PR — the original #4711 was closed automatically when my fork was deleted; same fix, no changes.)

### DIFF
--- a/src/everything/resources/subscriptions.ts
+++ b/src/everything/resources/subscriptions.ts
@@ -166,3 +166,22 @@ export const stopSimulatedResourceUpdates = (sessionId?: string) => {
     subsUpdateIntervals.delete(sessionId);
   }
 };
+
+/**
+ * Removes a session from every URI's subscriber set, dropping any URI entry
+ * that ends up with no remaining subscribers.
+ *
+ * A session that disconnects without explicitly unsubscribing otherwise stays
+ * in `subscriptions` for the life of the process. Call this from the
+ * transport's `cleanup(sessionId)` when a session ends.
+ *
+ * @param {string} [sessionId]
+ */
+export const removeSubscriber = (sessionId?: string) => {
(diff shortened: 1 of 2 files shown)


vishalhabib99:
Pulled 8e2f0fe1, 109 passing vs 107 on base (`d73f99e`), build and `tsc --noEmit` on `src/everything` both clean.

Added two regression tests for `removeSubscriber`: one confirming a removed session stops receiving simulated resource-update notifications, one confirming removing that session doesn't affect other sessions still subscribed to the same URI. Verified both actually catch the regression — stubbing `removeSubscriber` to a no-op locally fails the first test, restoring the fix passes it again.

vishalhabib99:
Rebased onto latest main and CI is green (all Python/TypeScript build+test jobs passing). Added two regression tests for the `removeSubscriber` fix (a removed session stops receiving resource-update notifications; removing one session doesn't affect others still subscribed to the same URI) — verified both actually catch the regression by stubbing the fix back out locally.

Ready for review whenever you have a chance.

olaservo:

vishalhabib99:
Thanks for the explanation, @olaservo — makes sense, glad the fix landed in #4712. I'll check for open PRs before filing next time.

PR #2054, opened 2025-06-09
fix: MCP error -32603: Cannot read properties of undefined (reading 'toLowerCase')

caizhiqiangzeyuan:
fix the bug [2044](https://github.com/modelcontextprotocol/servers/issues/2044)

## Description
1.Safer Query Handling:
    Since query is typed as string, we skip unnecessary safety checks and directly process it with lowerQuery =  query.trim().toLowerCase()

2.Extracted Matching Logic:
    Created reusable checkMatch helper that encapsulates:
        Type checking (typeof text === 'string')
        Normalization (trim()+toLowerCase())
        Search logic (includes())

3.Performance Optimizations:
    Pre-computes lowerQuery once instead of per-filter iteration
    Eliminates repeated arrow function creation in array operations

4.Improved Readability:
    Flattened nested conditions
    Clear separation of concerns between filtering steps
   More descriptive variable names

## Server Details
- Server: The search_nodes function in the MCP memory server fails with a JavaScript error when attempting to search the knowledge graph. The error indicates that the search function is trying to call .toLowerCase() on an undefined value without proper input validation.
- Changes to: Map to safe lower case

## Motivation and Context
[bug address](https://github.com/modelcontextprotocol/servers/issues/2044)

## How Has This Been Tested?
Yes, Cursor and Claude.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Protocol Documentation](https://modelcontextprotocol.io)
- [x] My changes follows MCP security best practices
- [ ] I have updated the server's README accordingly
- [x] I have tested this with an LLM client
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have documented all environment variables and configuration options

PR #4717, opened 2026-08-29
fix(memory): validate knowledge graph entries when loading from disk

### LINKED ISSUE #2044: "Cannot read properties of undefined (reading 'toLowerCase')" Error
## Description
The `search_nodes` function in the MCP memory server fails with a JavaScript error when attempting to search the knowledge graph. The error indicates that the search function is trying to call `.toLowerCase()` on an undefined value without proper input validation.

## Error Message
```
MCP error -32603: Cannot read properties of undefined (reading 'toLowerCase')
```

## Environment
- **OS**: macOS (using Homebrew)
- **Node.js version**: v23.11.0
- **MCP Memory Server version**: Latest (installed via npx)
- **Client**: Claude Desktop and Claude Code

## Configuration
```json
{
  "memory": {
    "command": "/opt/homebrew/bin/npx",
    "args": ["-y", "@modelcontextprotocol/server-memory"]
  }
}
```

The MCP server was also added to Claude Code using:
```bash
claude mcp add-from-claude-desktop
```

## Steps to Reproduce
1. Set up MCP memory server with the above configuration
2. Create some entities in the knowledge graph (this works fine)
3. Attempt to use the `search_nodes` function with any query string
4. Error occurs consistently in both Claude Desktop and Claude Code environments

## Expected Behavior
The `search_nodes` function should successfully search through entity names, types, and observation content, returning relevant matches.

## Actual Behavior
The function throws a JavaScript error and fails to return any results.

## Workarounds
- `read_graph` function works correctly
- `open_nodes` function works correctly when entity names are known
- The issue is isolated to the search functionality

## Additional Context
This appears to be related to insufficient input validation in the search function. The error suggests that the function attempts to call `.toLowerCase()` on a parameter that is undefined or null.

A similar error pattern has been reported with `.includes()` method: "Cannot read properties of undefined (reading 'includes')" indicating a broader input validation issue in the memory server's search implementation.

## Impact
This bug prevents users from effectively searching their knowledge graphs, limiting the utility of the memory server for knowledge retrieval operations.

---
*This issue was reported with assistance from Claude (Anthropic).*

Ethanz11-creat:
## Description

Fixes #2044 by addressing the root cause rather than adding defensive checks to `searchNodes`.

The crash happens because `loadGraph()` trusts the persisted `memory.jsonl` file and pushes every parsed line into the in-memory graph without validating it. A corrupted or legacy entry — an entity missing `entityType`, an observation that is `null`, a relation missing `relationType` — reaches `searchNodes` and throws `Cannot read properties of undefined (reading 'toLowerCase')`.

This change validates each line against the existing `EntitySchema` / `RelationSchema` before adding it to the graph, and skips malformed entries (including malformed JSON lines) with a warning on stderr. Tools already validate their inputs with these schemas, so this closes the last unvalidated path (the on-disk file) and guarantees the in-memory graph only ever contains well-formed entities and relations.

This follows the direction from the earlier attempt (#2054): fix the source of non-strings rather than papering over the crash in `searchNodes`.

## Server Details
- Server: memory
- Changes to: knowledge graph persistence (loading)

## Motivation and Context

`search_nodes` crashes with `MCP error -32603: Cannot read properties of undefined (reading 'toLowerCase')` for users whose memory file contains a legacy or hand-edited entity. See #2044.

## How Has This Been Tested?

- `npm run build` (tsc) passes.
- `npx vitest run` passes: 58 tests across 4 files (unit level; not yet exercised through a live LLM client).
- Added 3 tests in `knowledge-graph.test.ts` under `loadGraph validation`:
  - corrupt entities (missing `entityType`, `null` observation) are skipped and `searchNodes` no longer throws;
  - corrupt relations (missing `relationType`) are skipped;
  - malformed JSON lines are skipped.
- Existing tests (including "strip type field when loading") still pass.

## Breaking Changes

None. Valid memory files load exactly as before; only malformed entries are now skipped instead of crashing the server.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

I kept the skip-with-warning behavior minimal and scoped to loading, so no existing data is mutated and nothing is silently dropped without a stderr trace. The warning text mirrors the existing `console.error` style used for file migration notices.

### DIFF
--- a/src/memory/__tests__/knowledge-graph.test.ts
+++ b/src/memory/__tests__/knowledge-graph.test.ts
@@ -515,4 +515,51 @@ describe('KnowledgeGraphManager', () => {
       expect(result.relations[0]).not.toHaveProperty('type');
     });
   });
+
+  describe('loadGraph validation', () => {
+    it('skips corrupt entities instead of crashing search', async () => {
+      const lines = [
+        JSON.stringify({ type: 'entity', name: 'Alice', entityType: 'person', observations: ['works at Acme Corp'] }),
+        JSON.stringify({ type: 'entity', name: 'Broken', observations: ['missing entityType'] }),
+        JSON.stringify({ type: 'entity', name: 'BadObs', entityType: 'person', observations: ['ok', null] }),
+      ];
+      await fs.writeFile(testFilePath, lines.join('\n') + '\n');
+
+      const graph = await manager.readGraph();
+      expect(graph.entities).toHaveLength(1);
+      expect(graph.entities[0].name).toBe('Alice');
+
+      // searchNodes must not throw even though the file contains corrupt entries
+      const result = await manager.searchNodes('Acme');
+      expect(result.entities).toHaveLength(1);
+      expect(result.entities[0].name).toBe('Alice');
+    });
+
+    it('skips corrupt relations', async () => {
+      const lines = [
+        JSON.stringify({ type: 'entity', name: 'Alice', entityType: 'person', observations: [] }),
+        JSON.stringify({ type: 'relation', from: 'Alice', to: 'Bob' }), // missing relationType
+        JSON.stringify({ type: 'relation', from: 'Alice', to: 'Bob', relationType: 'knows' }),
+      ];
+      await fs.writeFile(testFilePath, lines.join('\n') + '\n');
+
+      const graph = await manager.readGraph();
+      expect(graph.entities).toHaveLength(1);
+      expect(graph.relations).toHaveLength(1);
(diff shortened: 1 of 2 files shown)


olaservo (approved):
Something to consider in a follow-up: a skipped entity is dropped from the file on the next write, because `saveGraph` rewrites from the loaded graph. For a line that is not JSON there is nothing else to do. For an entity that only has one non-string observation it would be gentler to filter `observations` down to the strings and keep the entity, so a single bad element does not delete the whole record. But it makes sense to keep this PR focused so I am fine merging as is.

PR #4731, opened 2026-09-01
fix(memory): guard against undefined observations in searchNodes

### LINKED ISSUE #1818: search_nodes fails with non-empty query
- Empty query ("") works and returns all entities/relations

- Non-empty query fails with the exact error described in the bug report: "Cannot read properties of undefined (reading 'some')"

Environment:

Windows 10, Visual Studio Code 1.100.2 with Cline extension 3.16, Python 3.12.

CryoThrust:
Closes #1818

## Problem

`search_nodes` throws `TypeError: Cannot read properties of undefined (reading 'some')` when an entity in the JSONL file has a missing `observations` field. This can happen with older data files or malformed entries.

## Fix

Use nullish coalescing (`??`) to default `observations` to an empty array before calling `.some()`:

```typescript
(e.observations ?? []).some(o => o.toLowerCase().includes(query.toLowerCase()))
```

## Tests

All 50 existing tests pass. The fix is backward-compatible — entities with properly defined `observations` arrays are unaffected.

### DIFF
--- a/src/memory/index.ts
+++ b/src/memory/index.ts
@@ -192,7 +192,7 @@ export class KnowledgeGraphManager {
     const filteredEntities = graph.entities.filter(e => 
       e.name.toLowerCase().includes(query.toLowerCase()) ||
       e.entityType.toLowerCase().includes(query.toLowerCase()) ||
-      e.observations.some(o => o.toLowerCase().includes(query.toLowerCase()))
+      (e.observations ?? []).some(o => o.toLowerCase().includes(query.toLowerCase()))
     );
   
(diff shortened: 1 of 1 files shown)


olaservo:
Thanks for this! #4717 addresses the same crash one level down: it validates each line of the memory file against the existing `EntitySchema` / `RelationSchema` in `loadGraph`, so a missing `observations` field, a null observation, or a malformed line never reaches `searchNodes` in the first place. That also covers #2044, which is the same failure on `toLowerCase`. Closing this one in favour of #4717.

PR #4734, opened 2026-09-01
feat(git): Add optional tamper-evident execution audit logging via GE…

viruswami5511:
### Summary
This PR adds optional tamper-evident execution audit logging to `mcp-server-git` via the `GIT_MCP_AUDIT_DIR` environment variable, implementing the open [GEF-SPEC-1.0](https://github.com/viruswami5511/guardclaw) standard (RFC 8785 canonicalization + Ed25519 hash chaining).

### Why This is Useful
When autonomous AI agents (Claude Desktop, Cursor, Cline) execute Git operations (`git_commit`, `git_reset`, `git_checkout`), users need forensic proof of what actions were taken without relying on mutable logs.

- **100% Optional & Zero Overhead**: If `GIT_MCP_AUDIT_DIR` is not set or `guardclaw` is not installed, the server operates with zero overhead and zero forced dependencies.
- **Offline Verifiable**: Ledgers can be verified anytime with `guardclaw verify <DIR>`.

### Verification
- [x] Verified standard execution works with 0 changes when unconfigured.
- [x] Added `test_audit_logging_when_enabled` in `tests/test_server.py`.

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from pathlib import Path
 from typing import Sequence, Optional
 from mcp.server import Server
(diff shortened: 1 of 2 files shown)


blev8824-ai:
This is the right shape: offline-verifiable, append-only, keyed on Ed25519. The property that actually needs pinning down is the trust anchor, not the signature scheme. A hash chain only proves ordering and tamper-evidence if the root is anchored somewhere the host cannot edit afterward; if the auditor and the agent share the same filesystem, the tamper-evidence collapses to the same trust boundary as the log itself. Storing the chain root out-of-band, a second process or a remote verifier that saw the first block, is what makes it forensic against a compromised host. Also worth deciding what a history-rewriting operation (like a reset) does to chain continuity. This is the same evidence model AgentKey uses, a hash-chained root plus an Ed25519 key only the holder can sign with, verified per call: https://agentkey.us

PR #4738, opened 2026-09-02
fix(memory): stop reporting deletions that did not happen

### LINKED ISSUE #4740: memory: delete tools report success when nothing matched
`delete_entities`, `delete_observations` and `delete_relations` return `success: true` with a hardcoded message regardless of what matched.

`src/memory/index.ts`:

```ts
async deleteEntities(entityNames: string[]): Promise<void> {
  graph.entities = graph.entities.filter(e => !entityNames.includes(e.name));
```

```ts
structuredContent: { success: true, message: "Entities deleted successfully" }
```

The method returns `void`, so the handler never learns whether a name existed. A call naming an entity that is not in the graph reports the same success as one that deleted something.

Reproduction, against the real `KnowledgeGraphManager`:

```ts
await manager.createEntities([{ name: "Alice", entityType: "person", observations: [] }]);
await callTool("delete_entities", { entityNames: ["Alise"] });
// → { success: true, message: "Entities deleted successfully" }
// Alice is still in the graph.
```

Same for `delete_observations` with an entity or observation that is absent, and `delete_relations` with a relation that matches nothing.

`addObservations` at line 167 throws `Entity with name ${o.entityName} not found` for the same condition, so the two paths in this file answer differently.

`README.md` documents the tools as silent when a target is missing, so not throwing looks intentional. The response claiming a deletion that did not occur is the part that appears unintended.

ConnorMoss02:
Closes #4740

## Problem

`delete_entities`, `delete_observations` and `delete_relations` report success no matter what matched:

```ts
structuredContent: { success: true, message: "Entities deleted successfully" }
```

The message is hardcoded, and the three manager methods return `void`, so nothing ever looks at whether a name existed. An agent that deletes `"Alise"` instead of `"Alice"` is told its memory is clean while the entity is still on disk, and nothing in the response contradicts that.

Ten lines above, `addObservations` throws `Entity with name ${o.entityName} not found` for exactly this condition, so the file already disagreed with itself about whether a missing entity is worth mentioning.

## Approach

`README.md` documented "Silent operation if entity doesn't exist", so the tools still do not throw. This changes the response only:

- Nothing new throws.
- `outputSchema` is unchanged: still `{ success: boolean, message: string }`.
- When every target is found, the message is byte-identical to before, so a caller matching on `"Entities deleted successfully"` is unaffected.

Only the case that was previously untrue changes:

```
delete_entities(["Alice", "Alise"])
  → "Deleted 1 of 2 entities. Not found: Alise"

delete_observations([{ entityName: "Carol", observations: ["x"] }])
  → "Deleted 0 of 1 observations. Entities not found: Carol"

delete_relations([...one match, one miss...])
  → "Deleted 1 of 2 relations. The rest matched nothing."
```

## Changes

- `src/memory/index.ts`: the three manager methods return what they matched; the handlers report it.
- `src/memory/__tests__/delete-reporting.test.ts`: 7 tests over the real `KnowledgeGraphManager`, no mocks.
- `src/memory/README.md`: the three "Silent operation" bullets described the absence of an error, which is still true, but read as if the response said nothing either.

## Validation

```bash
cd src/memory && npx tsc --noEmit && npx vitest run
```

62 passed across 5 files. Typecheck clean.

Confirmed the tests are load-bearing: with `index.ts` reverted, 6 of the 7 new tests fail.

## Note

I put the tests in a new file rather than `__tests__/index.test.ts`, since #4692 is adding that file. Its delete tests assert on the persisted graph rather than the response text, so the two do not overlap.

### DIFF
--- a/src/memory/README.md
+++ b/src/memory/README.md
@@ -86,7 +86,7 @@ Example:
   - Remove entities and their relations
   - Input: `entityNames` (string[])
   - Cascading deletion of associated relations
-  - Silent operation if entity doesn't exist
+  - No error if an entity doesn't exist; the response reports which names were not found
 
 - **delete_observations**
   - Remove specific observations from entities
@@ -94,7 +94,7 @@ Example:
     - Each object contains:
       - `entityName` (string): Target entity
       - `observations` (string[]): Observations to remove
-  - Silent operation if observation doesn't exist
+  - No error if an observation doesn't exist; the response reports how many were deleted
 
 - **delete_relations**
   - Remove specific relations from the graph
@@ -103,7 +103,7 @@ Example:
       - `from` (string): Source entity name
       - `to` (string): Target entity name
       - `relationType` (string): Relationship type
-  - Silent operation if relation doesn't exist
+  - No error if a relation doesn't exist; the response reports how many were deleted
 
 - **read_graph**
   - Read the entire knowledge graph
--- a/src/memory/__tests__/delete-reporting.test.ts
+++ b/src/memory/__tests__/delete-reporting.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { promises as fs } from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { KnowledgeGraphManager, Entity, Relation } from '../index.js';
+
+/**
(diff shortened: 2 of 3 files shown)


olaservo (approved):
Looks good to me. The success path is byte-identical, the output schema is untouched, and the counts make the mistyped-name case visible to the agent without turning it into an error. Approving.

PR #4744, opened 2026-09-03
fix(memory): add a cross-process file lock so separate server processes do not overwrite each other

daichiyasunami-vottia:
Refs #1819, #3286. Builds on #4555.

#4555 serialises mutations within one process, and it fully fixes that case. The server is stdio-only, so every client is its own process with its own queue: two editor windows, or two git worktrees pointed at one `MEMORY_FILE_PATH`, still overwrite each other's load→mutate→save.

**Measured on `d73f99e`** (this branch's base), 20 `create_entities` calls:

```
serial   kept=20 lost=0
single   kept=20 lost=0     one process, 20 in flight   <- #4555
multi    kept=10 lost=10    two processes, one file      <- 5 runs, exactly 10 each time
```

Each process serialises its own half correctly and the last one to write the file wins. Every call returns success.

**Change** (one method, no new dependency): wrap each mutation in a sidecar lock `MEMORY_FILE_PATH.lock` created with `fs.open(..., 'wx')` — O_EXCL is atomic on POSIX and Windows. Retry with a short backoff; a lock older than 30 s is treated as left by a crashed process and reclaimed. The in-process queue is kept, so a process never contends with itself. With this branch the `multi` row is 20/20.

**Relation to #3286:** same goal. That PR is now `CONFLICTING` against main after #4555 and adds `proper-lockfile` (+1007 lines). This is the minimal version on top of the current queue; happy to close in favour of a rebased #3286 if that is preferred.

Tests: `__tests__/cross-process-lock.test.ts` — two instances on one file (a faithful stand-in for two processes, since the queue is per instance) keep 20/20; the lock file is removed after each mutation; a stale lock is reclaimed. Full suite: 90 passed.

### DIFF
--- a/src/memory/__tests__/cross-process-lock.test.ts
+++ b/src/memory/__tests__/cross-process-lock.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { promises as fs } from 'fs';
+import path from 'path';
+import os from 'os';
+import { KnowledgeGraphManager } from '../index.js';
+
+/**
+ * Regression tests for cross-process exclusion (#1819, #3286).
+ *
+ * The in-process mutation queue (#4555) serialises one instance's mutations,
+ * but the server is stdio-only, so every client is its own process with its
+ * own queue. Two instances on one file are a faithful stand-in: nothing but
+ * the sidecar lock file separates them, exactly as with two processes.
+ */
+describe('KnowledgeGraphManager cross-process exclusion', () => {
+  let testDir: string;
+  let testFilePath: string;
+
+  beforeEach(async () => {
+    testDir = await fs.mkdtemp(path.join(os.tmpdir(), 'mcp-memory-xproc-'));
+    testFilePath = path.join(testDir, 'memory.jsonl');
+  });
+
+  afterEach(async () => {
+    await fs.rm(testDir, { recursive: true, force: true });
+  });
+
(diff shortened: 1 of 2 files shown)


daichiyasunami-vottia:
Addressed the review in bcc54ec — both races were real, and both are the kind that reintroduce exactly the lost update this PR is for.

**What was wrong.** With a bare `stat → unlink → open('wx')`, N waiters that all observed the same stale lock could each unlink it, so a slow waiter deleted the fast waiter's fresh lock and both proceeded. Separately, any operation longer than the stale threshold had its live lock taken, and its `finally` then unlinked the successor's lock.

**What changed** (still one method, still no dependency):

- the lock file holds the holder's token (`pid:random`)
- a live holder **heartbeats** the lock's mtime every 5 s, so a long operation is never mistaken for a crashed one
- stale takeover goes through **`rename(2)`**, not unlink: of N reclaimers exactly one rename succeeds, the rest get `ENOENT` and go back to waiting on the winner's lock
- release unlinks **only if the file still holds our token**, so a holder can never remove a replacement's lock

**Tests added** for the two review cases: eight simultaneous reclaimers of one pre-aged stale lock keep 8/8 entities and leave no lock behind; a slow, heartbeating holder is not stolen from, and its release does not remove the successor's lock. Full suite 92/92. The two-process harness still keeps 20/20 (3 runs).

Happy to take another pass if the protocol should look different — the alternative is a proven lockfile dependency, which #3286 proposes; this keeps the reference server dependency-free.

PR #4749, opened 2026-09-04
fix(sequentialthinking): accurate readOnlyHint and idempotentHint annotations (#4721)

### LINKED ISSUE #4721: sequential-thinking: readOnlyHint and idempotentHint annotations are inaccurate (server is stateful, non-idempotent)
## Summary

The `sequentialthinking` tool annotations `readOnlyHint: true` and `idempotentHint: true` are inaccurate. The server is **stateful** and **non-idempotent** — it maintains per-session `thoughtHistory` and `branches` instance state that persists across calls and changes on every invocation.

These annotations were added per #3403's suggestion, but the suggested values were not verified against the server's actual stateful behavior.

## Evidence (BETS cycle, 2026-08-30)

Ran a governed Benchmarking/Evaluation/Testing/Standards (BETS) cycle against `@modelcontextprotocol/server-sequential-thinking` (published npm version, via `npx -y`). 11 test cases across 4 lanes.

### `readOnlyHint: true` is misleading

The `SequentialThinkingServer` class (confirmed in `src/sequentialthinking/lib.ts` on `main`) holds instance state:
- `private thoughtHistory: ThoughtData[] = []` (line 23-24)
- `private branches: Record<string, ThoughtData[]> = {}` (line 25-26)

Each `processThought` call **mutates** this state:
- `this.thoughtHistory.push(input)` (line 101) — appends to history on every call
- `this.branches[input.branchId].push(input)` (line 111-112) — appends to branch when `branchFromThought` + `branchId` provided

Observed across 11 calls in one session: `thoughtHistoryLength` incremented linearly (1, 2, 3, ... 11). The server is not read-only — it accumulates session state with each invocation.

### `idempotentHint: true` is false

A tool is idempotent if repeated calls with the same arguments produce the same result. Calling `sequentialthinking` twice with identical arguments produces **different** `thoughtHistoryLength` values (N, then N+1) and potentially different `branches` contents. The tool is non-idempotent by definition.

## Suggested fix

```typescript
annotations: {
  readOnlyHint: false,    // server maintains thoughtHistory and branches state
  destructiveHint: false, // accurate — no external side effects
  idempotentHint: false,  // repeated calls produce different history lengths
  openWorldHint: false,   // accurate — operates on internal reasoning state
},
```

## Impact

Clients relying on `readOnlyHint: true` for safety assumptions (e.g., "safe to call in parallel", "no state side effects to roll back") will miss the session-state accumulation. Clients relying on `idempotentHint: true` for caching/retry (e.g., "safe to retry without checking result") will misbehave — a retried call produces a different history length and may create a duplicate branch entry.

## Source confirmed on `main`

`src/sequentialthinking/index.ts` lines 201-209 (current `main` as of 2026-08-30):
```typescript
annotations: {
  readOnlyHint: true,     // ← should be false
  destructiveHint: false,
  idempotentHint: true,   // ← should be false
  openWorldHint: false,
},
```

`src/sequentialthinking/lib.ts` lines 23-26, 101, 111-112 confirm the stateful behavior.

## Context

- BETS receipt: `agents/_state/bets/20260830T092705Z_sequential-thinking.json` (HUMMBL internal, available on request)
- Related: #3403 (added these annotations without verification — the suggestion was accepted as-is)
- Discovered during a governed BETS cycle on the MCP fleet, not via automated scanner

Happy to open a PR if that's preferred.

teddiesloco:
## Summary
- Fixes #4721
- Corrects `readOnlyHint` from `true` to `false` because `SequentialThinkingServer` actively mutates session state (`thoughtHistory` and `branches` array) on each invocation.
- Corrects `idempotentHint` from `true` to `false` because repeated invocations with the same inputs produce different `thoughtHistoryLength` values and branch mutations.
- Both hints now accurately reflect the stateful runtime nature of the sequentialthinking tool.

### DIFF
--- a/src/sequentialthinking/index.ts
+++ b/src/sequentialthinking/index.ts
@@ -93,9 +93,9 @@ You should:
       needsMoreThoughts: coercedBoolean.optional().describe("If more thoughts are needed")
     },
     annotations: {
-      readOnlyHint: true,
+      readOnlyHint: false,
       destructiveHint: false,
-      idempotentHint: true,
+      idempotentHint: false,
       openWorldHint: false,
     },
     outputSchema: {


asd59478126-commits:
I think there may be a simpler way to approach this kind of issue.

The important part is not only determining whether `readOnlyHint` or `idempotentHint` is correct, but reliably locating the state changes that justify the classification in the first place.

For `sequentialthinking`, the relevant chain is essentially:

```text
tool call
→ processThought()
→ thoughtHistory / branches mutation
→ persistent state change
→ next tool call
```

If the repository structure preserves that sequence and its boundaries explicitly, the annotation can be evaluated from the recorded lineage instead of requiring the caller to reconstruct the behavior manually from multiple source locations.

In other words, the difficult part is often finding and connecting the relevant evidence, not the final boolean decision.

This is the kind of problem where a lineage-oriented intermediate representation could be useful.

PR #4761, opened 2026-09-05
fix(git): refuse to commit when nothing is staged

### LINKED ISSUE #4762: git_commit reports success and creates an empty commit when nothing is staged
### Describe the bug

`git_commit` reports success when nothing is staged. It creates an empty commit and returns a hash, so a caller cannot tell it apart from a real commit.

`src/git/src/mcp_server_git/server.py:128-130`:

```python
def git_commit(repo: git.Repo, message: str) -> str:
    commit = repo.index.commit(message)
    return f"Changes committed successfully with hash {commit.hexsha}"
```

`repo.index.commit()` writes a tree from the index unconditionally; GitPython has no `--allow-empty` gate. The return string cannot be false.

The practical failure: an agent edits files, calls `git_commit` without `git_add` — or after a `git_add` whose paths matched nothing — and gets a hash back. It tells the user the work is committed. The working tree is still dirty, `HEAD` still holds the old content, and the repository now carries an empty commit.

`git commit` itself refuses this with "no changes added to commit".

### To Reproduce

```python
import git
repo = git.Repo(".")            # one tracked file, edited but not staged
commit = repo.index.commit("agent says: fix applied")
print(commit.parents[0].tree.hexsha == commit.tree.hexsha)   # True — empty
```

```
status before:  M a.txt
git_commit ->   Changes committed successfully with hash efd05335b0ab
status after:   M a.txt
a.txt in HEAD:  v1        # the edit was never committed
```

### Expected behavior

Refuse, as `git commit` does without `--allow-empty`. Two cases git does permit and that should keep working: the first commit on an unborn branch, and an empty merge commit while `MERGE_HEAD` is present.

### Additional context

No test covers this; `test_git_commit` stages a file first.

#4761 implements it.

ConnorMoss02:
Closes #4762

`git_commit` returns `Changes committed successfully with hash <sha>` even when nothing is staged.

`repo.index.commit()` writes a tree from the index unconditionally — GitPython has no `--allow-empty` gate — and the return string cannot be false:

```python
def git_commit(repo: git.Repo, message: str) -> str:
    commit = repo.index.commit(message)
    return f"Changes committed successfully with hash {commit.hexsha}"
```

An agent edits files, calls `git_commit` without `git_add` (or after a `git_add` that matched nothing), and gets a hash. It reports the work as committed. The working tree is still dirty, `HEAD` still holds the old content, and the repo now carries an empty commit. Nothing in the response distinguishes that from a real commit.

```
status before:  M a.txt
returns:        Changes committed successfully with hash efd05335b0ab
empty commit:   True        # parent tree == new tree
status after:   M a.txt
a.txt in HEAD:  v1          # the edit was never committed
```

`git commit` itself refuses this: *"no changes added to commit"*.

### Fix

Raise when the index matches HEAD, matching git's default. Two cases git does allow are preserved:

- the first commit on an unborn branch, where there is no HEAD to compare against
- an empty merge commit, which git permits while `MERGE_HEAD` exists — a conflict resolved back to HEAD's content leaves the index matching HEAD, and that commit is legitimate

No new parameter: git requires `--allow-empty` explicitly, so refusing by default matches it.

### Tests

Six, covering both directions. Three assert the refusal (unstaged edit, clean tree, untracked-only) and fail with the source change reverted. Three assert it does not over-refuse (staged deletion, unborn branch, empty merge commit) and pass either way.

53 passed in `src/git`.

### DIFF
--- a/src/git/src/mcp_server_git/server.py
+++ b/src/git/src/mcp_server_git/server.py
@@ -125,7 +125,28 @@ def git_diff(repo: git.Repo, target: str, context_lines: int = DEFAULT_CONTEXT_L
     repo.rev_parse(target)  # Validates target is a real git ref, throws BadName if not
     return repo.git.diff(f"--unified={context_lines}", target)
 
+def _has_staged_changes(repo: git.Repo) -> bool:
+    """Whether the index holds anything git would record as a commit.
+
+    Mirrors `git commit`, which refuses to create an empty commit unless
+    --allow-empty is given, but permits one while a merge is in progress.
+    """
+    if (Path(repo.git_dir) / "MERGE_HEAD").exists():
+        return True
+    if not repo.head.is_valid():
+        # Unborn branch: the first commit, so anything in the index counts.
+        return bool(repo.index.entries)
+    return bool(repo.index.diff(repo.head.commit))
+
 def git_commit(repo: git.Repo, message: str) -> str:
+    # repo.index.commit() writes a tree unconditionally, so without this check
+    # a caller that forgot to stage gets a hash back for an empty commit and no
+    # way to tell it apart from a real one.
+    if not _has_staged_changes(repo):
+        raise ValueError(
+            "No changes staged for commit. Use git_add to stage changes first; "
+            "git_status shows what is currently staged."
+        )
     commit = repo.index.commit(message)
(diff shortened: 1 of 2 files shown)


ConnorMoss02:
Closes #4762

PR #2568, opened 2025-08-17
Add host allowlisting and IP blocking to Fetch server

olaservo:
Implements security feature to prevent access to local/internal IP addresses by default. Addresses issue #2317.

## Changes
- Block access to private IP ranges by default
- Add host allowlisting with wildcard support
- Add configuration options for security customization
- Implement socket-level validation to prevent TOCTOU attacks
- Update documentation with security examples

Generated with [Claude Code](https://claude.ai/code)

PR #4770, opened 2026-09-06
feat(fetch): add opt-in host allowlisting (--allowed-hosts)

### LINKED ISSUE #2317: Add support for host allowlisting to Fetch server
**Is your feature request related to a problem? Please describe.**
The `fetch` MCP server currently allows connections to local/internal IP addresses without any restrictions, which poses a security risk. While the README has been updated with a caution notice, this reference implementation could be misused if deployed without proper safeguards.

**Describe the solution you'd like**
Implement a host allowlisting mechanism for the fetch server that would:
- Allow administrators to specify which hosts/domains the server can connect to
- By default, block connections to local/internal IP ranges (127.0.0.0/8, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, etc.)
- Provide configuration options to explicitly allow internal IPs if needed for specific use cases
- Implement checks at the socket level to avoid TOCTOU (Time-of-Check-Time-of-Use) vulnerabilities

**Describe alternatives you've considered**
- **Leave as-is with documentation only**: Continue with just the warning in the README, but this leaves the reference implementation vulnerable and sets a poor security example

**Additional context**
- Consider adding example configurations to the Readme showing secure deployment patterns.

g-yixuan:
## Description

Adds an opt-in host allowlist to the `fetch` server, addressing the SSRF surface tracked in #2317 **without changing default behavior**.

Fixes #2317.

- New `--allowed-hosts HOST [HOST ...]` CLI flag. When unset, behavior is exactly as before (all hosts allowed), so deployments that intentionally fetch localhost/internal endpoints are unaffected (per the discussion on the issue).
- Entry forms: exact host, case-insensitive (`example.com`); wildcard (`*.example.com`) matching the bare domain and any subdomain; literal IPs.
- Enforcement covers the initial request, the `robots.txt` pre-check, and **every redirect hop**: redirects are now followed manually (bounded at 20, same as httpx's default) and each hop is re-validated before connecting, so a 302 from an allowed host can no longer bounce a fetch to a disallowed host.
- Denials fail closed with a clear error that does not echo the configured allowlist.

## Publishing Your Server

N/A — change to an existing server.

## Server Details

- Server: fetch (Python, `mcp-server-fetch`)
- Changes to: tools/prompts request handling (URL validation + redirect handling), CLI args, README

## Motivation and Context

#2317 asks for host allowlisting. The issue discussion raised valid concerns about breaking users who intentionally fetch internal endpoints, so this PR deliberately does **not** change the default posture — the default-deny/IP-range part of the issue is separately covered by #4497, and this PR composes cleanly with it. A previous allowlisting attempt (#2568) went stale and was closed with an invitation to resubmit on a current base; this is that fresh take, scoped to avoid the default-behavior controversy.

On the TOCTOU/socket-level point in the issue: with a hostname allowlist, the validated hostname is exactly what gets connected (SNI/Host), so DNS check-time/connect-time skew does not weaken the control; the realistic bypass was redirect hops, which this closes.

## How Has This Been Tested?

- `uv run pytest` — 64 passed (existing suite untouched and green, plus new coverage: exact/wildcard/IP matching, case/trailing-dot normalization, userinfo bypass attempt, per-hop redirect enforcement parametrized over 301/302/303/307/308, redirect-loop limit, empty and malformed `Location` headers (matching httpx's prior behavior), relative `Location` resolution, robots.txt path enforcement, per-request timeout semantics, and unset-flag backwards compatibility).
- `uv run pyright` — 0 errors; `uv run ruff check .` — clean.
- Exercised end-to-end over the MCP protocol: stdio client against the real server binary started with `--allowed-hosts 127.0.0.1`, plus a real local HTTP server — allowlisted host fetched, `http://169.254.169.254/latest/meta-data` blocked, redirect to a non-allowlisted host blocked, malformed `Location` handled with a clean error.
- URL validation uses `httpx.URL` — the same parser that performs the connection — so the validated host is always the host that gets connected.

## Breaking Changes

None — default behavior is unchanged when `--allowed-hosts` is not set.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

- Wildcard semantics are documented in the README: `*.example.com` matches `example.com` itself and any subdomain.
- The README also states an intentional limitation: the allowlist matches hostnames, not resolved IPs, which is what makes deliberately allowlisting internal hosts possible.


blev8824-ai:
Nice that redirect hops are re-validated instead of trusting the first DNS result; reverifying a 302 against the allowlist before connecting is the right way to close the classic SSRF bounce. Two edges worth pinning down. First, the robots.txt precheck also opens a connection, so it should go through the same gate (and count against the 20-hop bound) or it becomes a side channel for probing hosts you intend to deny. Second, decide explicitly whether `*.example.com` also allows the bare apex `example.com`; that single wildcard rule is what decides whether a plain root domain resolves under an allowlist entry. Nice that denials fail closed without echoing the allowlist. Same boundary we enforce in AgentKey, which checks each tool call at dispatch and keeps the decision record per request: https://agentkey.us

g-yixuan:
Thanks for the look — both points are already covered:

- The robots.txt pre-check goes through the same gate: it uses the same `_get_following_redirects` helper, so the robots fetch and every one of its redirect hops are validated against the allowlist and count against the same 20-redirect bound (see `test_robots_redirect_to_denied_host_blocked`).
- Wildcard semantics are explicit: `*.example.com` matches the bare apex `example.com` as well as any subdomain — documented in the README and pinned by `test_wildcard_matches_bare_domain`.

blev8824-ai:
Both covered, agreed. The seam left here is TOCTOU: the gate revalidates the redirect URL string against the allowlist, then the HTTP client re-resolves DNS at connect time, so an attacker can race a name that answers with an allowlisted IP at the moment the gate checks and a private or link-local address at connect. Closing it means resolving each hop once, pinning the validated address set, and connecting to the pinned IPs, plus normalizing representations (IPv6-mapped [::ffff:7f00:1], decimal-octet 2130706433, trailing-dot FQDN, punycode) so the host you match equals the address you connect to. That validate-then-connect coupling is the same one AgentKey pins per tool call at dispatch, with the decision record kept per request: https://agentkey.us. Free pre-SOC2.

tiagovilasboas:
### AppSec review — fetch `--allowed-hosts` (#4770)

Solid allowlist slice of #2317: httpx.URL parsing, per-hop redirect revalidation (incl. robots), wildcard/exact/IP tests, and fail-closed denials that do not echo the list. Hostname-only matching (DNS→internal for an allowlisted name) is documented and intentional.

**Verdict: Request changes** (comment form — no formal review permission on this repo).

### Blocking

1. **Empty/whitespace `Location`:** `Location: ""` / spaces → `urljoin` keeps current URL and burns up to 20 hops; missing `Location` on 3xx returns the redirect body. Please fail closed on empty/missing Location when status is redirect (same ask as on #4773) — update the tests that currently pin httpx’s loop/`as-is` behavior.

2. **Compose with #4773 via one redirect helper:** This PR’s `_get_following_redirects` (allowlist, cap 20) and #4773’s `_get_with_validation` (scheme + private-IP denylist, cap 5) both replace `follow_redirects=True`. Land a shared per-hop helper (allowlist ∪ IP/scheme checks, one hop budget, one Location policy) or the merge will drop a control.

3. **Don’t `Fixes #2317` wholesale:** Issue also wanted default private-IP blocking and socket-level TOCTOU. This correctly scopes to opt-in allowlisting only — use Partial/#2317 and leave default-deny to #4497/#4773.

### Non-blocking

- IDN: httpx returns Unicode hosts; punycode allowlist entries silently never match — normalize or document.
- Error code `INTERNAL_ERROR` vs `INVALID_PARAMS` for denials — minor.

Happy to re-review once the three blockers are addressed.

g-yixuan:
Thanks for the detailed pass — responses to the three blocking points:

**1. Empty/missing `Location`:** this is deliberate. The manual loop preserves httpx's prior semantics exactly — an empty `Location` self-redirects until the shared 20-hop bound raises a clean error, and a redirect status without `Location` is returned as the final response. Neither case opens a connection to a new host (the allowlist check fires before every hop's connection), so there's no SSRF implication; failing closed here would change default behavior for users who never set the flag, which this PR intentionally avoids. Happy to switch if maintainers prefer fail-closed as a blanket policy.

**2. Composing with #4773:** agreed that both PRs replacing `follow_redirects=True` need one seam. That looks like a merge-order decision for maintainers — whichever lands second should fold into a shared per-hop helper. This PR's `_get_following_redirects` is intentionally that minimal seam; I'm happy to rebase onto a common helper once there's direction.

**3. `Fixes #2317` linkage:** the PR body already scopes itself to opt-in allowlisting and points to #4497 for the default-deny portion. I kept `Fixes` since this delivers the issue's title feature, but no strong attachment — can switch to `Part of` if maintainers prefer.

Non-blocking items: IDN mismatch is fixed — hosts and entries are now IDNA-normalized before comparison, so Unicode and punycode forms compare equal in both directions, wildcards included (see `test_idn_matches_punycode_form`). Latest push: `3e5bf661`, CI green, 65 tests passing locally.

PR #4773, opened 2026-09-07
fix(fetch): harden URL fetching against SSRF

### LINKED ISSUE #3741: Security Hardening Recommendations for Fetch Server (SSRF Prevention)
## Security Review: Fetch Server

I have been building MCP servers for production use (12 servers, 280+ tools deployed) and noticed some security considerations for the fetch server that could benefit the reference implementation:

### 1. URL Validation

**Current:**
```typescript
const url = new URL(args.url);
```

**Issue:** This allows fetching from internal network addresses (169.254.x.x, 10.x.x.x, etc.) which could expose metadata services or internal APIs.

**Recommendation:**
```typescript
const BLOCKED_HOSTS = [
  /^169\.254\./,  // Link-local
  /^10\./,        // Private A
  /^172\.(1[6-9]|2[0-9]|3[0-1])\./,  // Private B
  /^192\.168\./,  // Private C
  /^127\./,       // Loopback
  /^0\./,         // Current network
  /^::1$/,        // IPv6 loopback
  /^fc00:/i,      // IPv6 private
  /^fe80:/i,      // IPv6 link-local
];

function isUrlAllowed(url: URL): boolean {
  const hostname = url.hostname;
  return !BLOCKED_HOSTS.some(pattern => pattern.test(hostname));
}
```

### 2. Content-Type Validation

**Current:** No validation on returned content types.

**Risk:** Could fetch binary executables or malicious content.

**Recommendation:**
```typescript
const ALLOWED_CONTENT_TYPES = [
  'text/html',
  'text/plain', 
  'text/markdown',
  'application/json',
  'application/xml',
];

// Validate before processing
const contentType = response.headers.get('content-type') || '';
if (!ALLOWED_CONTENT_TYPES.some(t => contentType.includes(t))) {
  throw new Error(`Content type not allowed: ${contentType}`);
}
```

### 3. Response Size Limits

**Current:** No limit on response size.

**Risk:** Memory exhaustion from huge responses.

**Recommendation:**
```typescript
const MAX_RESPONSE_SIZE = 10 * 1024 * 1024; // 10MB

const contentLength = response.headers.get('content-length');
if (contentLength && parseInt(contentLength) > MAX_RESPONSE_SIZE) {
  throw new Error(`Response too large: ${contentLength} bytes`);
}
```

### 4. Timeout Configuration

**Current:** Uses default fetch timeout.

**Risk:** Hanging connections consuming resources.

**Recommendation:**
```typescript
const FETCH_TIMEOUT = 30000; // 30 seconds

const controller = new AbortController();
const timeout = setTimeout(() => controller.abort(), FETCH_TIMEOUT);

try {
  const response = await fetch(url, { 
    signal: controller.signal,
    // ... other options
  });
} finally {
  clearTimeout(timeout);
}
```

### 5. Redirect Handling

**Current:** Default redirect behavior.

**Risk:** Open redirect vulnerabilities.

**Recommendation:**
```typescript
const response = await fetch(url, {
  redirect: 'manual', // Handle redirects explicitly
  // ...
});

if (response.status >= 300 && response.status < 400) {
  const location = response.headers.get('location');
  // Validate redirect target before following
  const redirectUrl = new URL(location!, url);
  if (!isUrlAllowed(redirectUrl)) {
    throw new Error('Redirect to blocked host');
  }
}
```

---

## Context

I run mcp-kali-orchestration and mcp-proxmox-admin in production environments where these security patterns are essential. Happy to discuss implementation details or contribute a PR.

**Author:** Eric Grill (https://ericgrill.com)
**Related work:** https://github.com/EricGrill/mcp-kali-orchestration

Ethanz11-creat:
## Summary

The fetch server currently accepts an arbitrary URL and follows HTTP redirects with no address/scheme checks, so an agent could read loopback, RFC1918/ULA, or cloud **metadata endpoints** (e.g. `169.254.169.254`, IMDS) when the MCP server runs on a cloud VM. This implements the SSRF hardening described in #3741.

**Changed (`src/fetch/src/mcp_server_fetch/server.py`):**
- Restrict allowed schemes to `http`/`https` (blocks `file://`, `data://`, `ftp://`, ... protocol-abuse vectors).
- Block non-public address ranges for both IPv4 and IPv6: loopback, RFC1918, IPv6 unique-local, link-local (incl. cloud metadata `169.254.0.0/16`), CGNAT, multicast, reserved, and IPv4-mapped IPv6 (`::ffff:127.0.0.1`).
- Hostnames are resolved (all A/AAAA records) and rejected **fail-closed** if *any* address lands on a blocked network.
- Redirects are now followed **manually, with a hop cap, re-validating every intermediate target** — previously `follow_redirects=True` let an open-redirect re-point the request at an internal address after the first URL passed.

Validation is a standalone pure function split so it is unit-testable without network.

## Tests

Added `TestIsBlockedIp`, `TestValidateUrl`, and `TestRedirectValidation` covering private/loopback/metadata IPv4+IPv6 literals, IPv4-mapped IPv6, scheme ablation, hostnames resolving to a blocked address, and redirects into private networks. Existing tests pin hostname resolution to a public IP so the suite stays hermetic.

Green on this branch:
```
pytest -q            55 passed  (21 pre-existing + 34 new)
ruff check           0 errors
ruff format --check  0 diffs
pyright              0 errors
```

Fixes #3741


blev8824-ai:
Solid hardening, and restricting schemes while blocking non-public ranges is the right baseline. Two edges worth tightening.

First, DNS rebinding: if validation resolves the hostname and blocks when any address lands in a blocked network, but the actual fetch is allowed to resolve independently (or uses a different resolver), an attacker who can influence resolution after the check can still reach loopback or the metadata endpoint. The fail-closed check is only sound if the validation resolves once and that exact result is what the fetch uses, or the fetch re-validates the address it actually connects to.

Second, redirects: re-validating every hop is good, but it should also re-check the scheme after each redirect. An https-to-http downgrade passes a pre-redirect scheme check and can land on an internal-only http target that the scheme filter thought it had excluded.

Does the validator force a single resolution that the subsequent fetch reuses, or can the fetch resolve separately after validation? And is the scheme re-validated on every redirect hop, or only on the initial URL?

A URL that classifies as allowed but targets an internal address is a per-call authorization decision, and the enforcement belongs at the tool gate with fail-closed behavior, which is the boundary AgentKey puts around the fetch (https://agentkey.us).

tiagovilasboas:
### AppSec review — fetch SSRF harden (#4773)

Strong direction on #3741: scheme allowlist, private/link-local denylist, fail-closed on any blocked A/AAAA at check time, and manual redirect revalidation (incl. robots.txt) are the right threat model. IPv4-mapped coverage looks good.

**Verdict: Request changes** (comment form — no formal review permission on this repo).

### Blocking

1. **DNS TOCTOU / rebinding:** `_validate_url` uses `getaddrinfo`, then `httpx` resolves again on connect. A name that flips public→IMDS between check and dial still bypasses. Prefer connect-by-validated-IP (keep Host/SNI/TLS for the original name), or document as accepted residual risk — I would not call the harden “done” while rebinding still hits metadata.

2. **Azure special IP:** `168.63.129.16` is not in `_BLOCKED_NETWORKS` and `ipaddress` treats it as global — literal fetch would pass today. Also consider NAT64 `64:ff9b::/96` and non-mapped `::127.0.0.1` forms (or move toward `not ip.is_global` with explicit exceptions).

3. **Empty `Location`:** `Location: ""` / whitespace via `urljoin` keeps `current` and burns the hop budget; missing `Location` on 3xx returns the redirect body. Please fail closed on empty/missing Location when `is_redirect` (#4770 already tests this).

### Non-blocking

- Decimal/octal/short IPv4: blocked via glibc `getaddrinfo` here but untested / libc-dependent — add cases.
- Hop cap 5 vs #4770’s 20: prefer one shared redirect helper that runs IP denylist + optional `--allowed-hosts` per hop (mechanical merge conflict otherwise).
- Error code `INTERNAL_ERROR` vs `INVALID_PARAMS` for blocked targets — minor.

Happy to re-review quickly once the three blockers are addressed or explicitly deferred with docs.

PR #4776, opened 2026-09-08
fix(fetch): add timeout to robots.txt request in autonomous fetch check

DawnofGenX:
## Summary

The content-fetch request in `fetch_url` sets `timeout=30`, but the robots.txt request issued by `check_may_autonomously_fetch_url` (which runs on **every** autonomous fetch) had no timeout. A slow or unresponsive `robots.txt` could therefore hang the autonomous-fetch check indefinitely, with no way for the caller to recover.

This adds the same 30s timeout to the robots.txt GET so it fails fast (surfacing as the existing connection-error path) instead of hanging, and adds a regression test asserting the timeout is present.

## Changes
- `src/fetch/src/mcp_server_fetch/server.py`: add `timeout=30` to the robots.txt `client.get(...)` call in `check_may_autonomously_fetch_url`.
- `src/fetch/tests/test_server.py`: new `test_robots_txt_request_uses_timeout` asserting the robots.txt GET carries `timeout=30`.

## Testing
- `uv run pytest -q` → 21 passed (was 20; +1 new regression test)
- `uv run ruff check .` → All checks passed
- `uv run pyright` → 0 errors, 0 warnings


DawnofGenX:
Gentle ping — all CI checks have been green since this was opened. The change is a one-line timeout on the robots.txt fetch in the autonomous fetch path, so it can't hang the server on a slow/unresponsive robots endpoint. Happy to make any adjustments if needed.

PR #4787, opened 2026-09-09
fix(docker): work around npm 10.x's arborist crash in the TS builders

AmirF194:
Bisected this a bit further than the issue's own finding: it isn't `@vitest/coverage-v8` specifically, plain `vitest` alone triggers the crash, even in an otherwise empty package.json. So it's an npm 10.x arborist bug in peer-dependency resolution, not anything about this repo's dependency graph. Node 24's bundled npm 11.19.0 resolves the exact same package.json with no errors and no extra flags; npm 10.9.8 (what the unpinned `node:22-alpine` tag ships today) still crashes.

`release.yml` already upgrades past this for a different reason (OIDC trusted publishing needs npm >=11.5.1, and its own comment says node 22's bundled npm is 10.x), so I did the same thing in each builder stage rather than adding a `--legacy-peer-deps` workaround or bumping the base image's node major. Confirmed npm 10.9.x in the release stage still reads a lockfile that npm 11 wrote via `npm ci`, so nothing downstream needs to change.

One thing that surprised me: none of the "Build"/"Test" CI jobs actually build these Dockerfiles. They run `npm ci` from a full monorepo checkout against the checked-in root workspace lockfile, a different install path that never hits this, which is probably why it shipped unnoticed.

Ran `docker buildx build --no-cache` for all four servers against a fresh main checkout first and got the issue's exact error on every one, then again on this branch where all four build clean. Also ran each built image through a real MCP `initialize` handshake over stdio to confirm it still starts and responds, and drove `filesystem` the same way with a mounted directory.


AmirF194:
This is a one-line npm upgrade fixing an arborist crash on Docker builds for everything/filesystem/memory/sequentialthinking, disjoint from the four open non-root-user PRs touching the same Dockerfiles (checked each for line overlap, there is none).

PR #4805, opened 2026-09-15
fix(git): report a detached HEAD from git_checkout

### LINKED ISSUE #4804: git_checkout reports "Switched to branch 'X'" when the checkout actually detached HEAD
### What happens

`git_checkout` validates `branch_name` with `repo.rev_parse(...)`, which resolves any revision — a commit sha, a tag, `HEAD~1`, a remote-tracking ref — not only branch names. It then checks out and returns a fixed string:

```python
repo.rev_parse(branch_name)  # Validates branch_name is a real git ref, throws BadName if not
repo.git.checkout(branch_name)
return f"Switched to branch '{branch_name}'"
```

`src/git/src/mcp_server_git/server.py` lines 202-204 on `main` (the flag-injection guard above them is elided). The tool description is `Switches branches`, the parameter is `branch_name`, and the README says the same: "`git_checkout` … Returns: Confirmation of branch switch". For every revision that is not a branch, the checkout detaches HEAD and the reply still claims a branch switch.

### Reproduction

`mcp-server-git` 2026.8.18, `mcp` 1.30.0, over stdio, on a fresh repo with three commits, a tag `v1` and a branch `feature`:

```
branch_name "feature"            -> Switched to branch 'feature'            | On branch feature   (correct)
branch_name "nonexistent"        -> isError=true, repo unchanged                                  (correct)
branch_name "49cea63" (a sha)    -> Switched to branch '49cea63'            | HEAD detached at 49cea63
branch_name "v1" (a tag)         -> Switched to branch 'v1'                 | HEAD detached at v1
branch_name "HEAD~1"             -> Switched to branch 'HEAD~1'             | HEAD detached at f1de938
branch_name "origin/main"        -> Switched to branch 'origin/main'        | HEAD detached at origin/main
branch_name "refs/heads/feature" -> Switched to branch 'refs/heads/feature' | HEAD detached at refs/heads/feature
```

The last two are the realistic ones: an agent writes `origin/main` or a full ref name far more often than a raw sha.

### Why this matters for the calling model

The reply is the only thing the model sees: `content` is a single TextContent, `structuredContent` is null, `isError` is false, the tool declares no `outputSchema`, and its annotations are static. Nothing in the response distinguishes an attached checkout from a detached one.

Git itself does distinguish them. `git checkout <sha>` writes the `You are in 'detached HEAD' state` advisory to stderr, and leaving a detached HEAD prints `Warning: you are leaving 1 commit behind` together with the `git branch <new-branch-name> <sha>` recipe. `repo.git.checkout()` returns `''` and that advisory is discarded, so the server replaces git's own warning with a success sentence. Work committed from there is on no branch and reachable only through the reflog until gc expires it; the tool never says so, and nothing in its reply suggests calling `git_status`.

### Suggested fix

Report the state rather than assume it:

```python
repo.git.checkout(branch_name)
if repo.head.is_detached:
    return f"HEAD is now detached at {repo.head.commit.hexsha[:7]}"
return f"Switched to branch '{repo.active_branch.name}'"
```

Verified against `main`: `src/git/tests/test_server.py` stays at 47 passed, and `git_checkout` on a branch name still returns `Switched to branch '<name>'`. Rejecting non-branch revisions would also close it, but that removes a capability some callers use deliberately.

Same root cause as #4762 and #4763 — no verification of the outcome after the git call — so this could fold into the same fix if that is easier to review.

CryoThrust:
## Summary
`git_checkout` validates its argument with `rev_parse`, which resolves *any* revision — a sha, a tag, `HEAD~1`, a remote-tracking ref, a full `refs/...` name. For everything that is not a branch, `git checkout` detaches HEAD, but the tool still returned `Switched to branch '<name>'`.

Report the resulting state instead:

```python
repo.git.checkout(branch_name)
if repo.head.is_detached:
    return f"HEAD is now detached at {repo.head.commit.hexsha[:7]}"
return f"Switched to branch '{repo.active_branch.name}'"
```

The reply is the only thing the calling model sees — `content` is a single text block, `structuredContent` is null and `isError` is false — so a wrong success sentence is not corrected downstream. The branch wording is now taken from `active_branch` too, so it cannot disagree with the ref that was actually checked out.

Rejecting non-branch revisions would also close the issue, but it removes a capability some callers use deliberately (checking out a tag or a remote ref), so I kept it and made the message truthful instead.

## Tests
Added three cases to `src/git/tests/test_server.py`:
- a sha detaches HEAD and the reply says so, without "Switched to branch"
- a tag behaves the same
- a real branch still returns `Switched to branch '<name>'` and leaves HEAD attached

On `main` the first two fail (`assert 'detached' in "Switched to branch 'v1'"`).

Validation: `uv run pytest` in `src/git` — 46 passed; `pyright` — 0 errors.

Closes #4804.


yangshuaiyufalv (commented):
[MCP demo] 测试评审（只读组织，预期权限错误）。

buyan201430-code:
Reviewed as the reporter of #4804. I ran the head (`d37bd1e`) locally: `pytest` in `src/git` gives 46 passed, matching your note.

The fix also covers the three rows of the #4804 table that the new tests don't exercise. Checked each against this branch:

```
HEAD~1              -> HEAD is now detached at <sha>   (head.is_detached: True)
refs/heads/feature  -> HEAD is now detached at <sha>   (head.is_detached: True)
origin/main         -> HEAD is now detached at <sha>   (head.is_detached: True)
```

A name that is both a branch and a tag (`dup`) still attaches to the branch and returns `Switched to branch 'dup'`, so taking the wording from `active_branch` holds there too.

One non-blocking suggestion: `origin/main` and `refs/heads/<branch>` are the two an agent is most likely to send, and they are the only ones not pinned by a test. This case passes on this branch and fails on `main` (3 failed with the old return line); `ruff format --check` is clean:

```python
@pytest.mark.parametrize("revision", ["HEAD~1", "refs/heads/feature", "origin/main"])
def test_git_checkout_non_branch_revisions_report_detached_head(
    test_repository, tmp_path, revision
):
    test_repository.git.branch("feature")
    test_repository.index.commit("second commit")
    git.Repo.init(tmp_path / "remote.git", bare=True)
    test_repository.create_remote("origin", str(tmp_path / "remote.git"))
    test_repository.git.push("origin", "HEAD:refs/heads/main")

    result = git_checkout(test_repository, revision)

    assert test_repository.head.is_detached
    assert result.startswith("HEAD is now detached at ")
```

Otherwise this looks right to me — keeping non-branch checkouts and making the reply truthful is a better trade than rejecting them.

CryoThrust:
Thanks for running it against the head and for checking the rows the tests didn't cover — that's exactly the gap I left, so I took the suggestion verbatim in `723b128`.

`test_git_checkout_non_branch_revisions_report_detached_head` is parametrized over `HEAD~1`, `refs/heads/feature` and `origin/main`, with the bare-remote scaffolding from your snippet. Confirmed the signal both ways: with the old unconditional return all three fail, and with the fix they pass and assert `result.startswith("HEAD is now detached at ")` rather than just the `is_detached` flag.

```
src/git: 49 passed   (46 before, +3 from the parametrization)
```

Your point about the branch-and-tag name resolving to the branch is a good one to have on record too — that's the case where taking the wording from `active_branch` matters most, since echoing the argument would have been ambiguous there.

buyan201430-code:
Re-checked at `723b128`. `pytest` in `src/git` gives 49 passed, which is the 46 from `d37bd1e` plus the three parametrized rows.

Confirmed the guard bites: with `src/mcp_server_git/server.py` reverted to `562feeb` and the new test file kept, `test_git_checkout_non_branch_revisions_report_detached_head` fails on all three revisions (`HEAD~1`, `refs/heads/feature`, `origin/main`), alongside the two pre-existing detached-HEAD tests, and everything passes again once the fix is back. So the rows from #4804 that had no coverage are pinned now.

Nothing further from me on this one.

